PR #1, opened 2020-02-05
Bulk inserts

stephenh:


integration/Author.ts
```diff
@@ -1,6 +1,22 @@
...
 export class Author {
-  id!: string;
+  readonly __meta = { type: "Author", data: {} as Record<any, any> };
 
-  firstName!: string;
+  constructor(private em: EntityManager) {
+    em.register(this);
+  }
+
+  get id(): string {
```
stephenh: These will be in the codegen'd base classes.

src/EntityManager.ts
```diff
@@ -45,27 +71,44 @@ export class EntityManager {
...
+}
+
 interface EntityMetadata {
   cstr: EntityConstructor<any>;
   tableName: string;
-  columns: Array<{ fieldName: string; columnName: string }>;
+  columns: Array<{ fieldName: string; columnName: string; serde: ColumnSerde }>;
 }
 
 const authorMeta: EntityMetadata = {
```
stephenh: These would also get codegen'd.

src/utils.ts
```diff
@@ -0,0 +1,10 @@
+export function getOrSet<T extends Record<keyof unknown, unknown>>(
```
stephenh: Needs more scala.

src/EntityManager.ts
```diff
@@ -2,41 +2,67 @@ import DataLoader from "dataloader";
...
+    const ps = Object.values(this.entities).map(async list => {
+      const meta = entityMeta[list[0].__meta.type];
+
+      const rows = list.map(entity => {
+        const row = {};
+        meta.columns.forEach(c => c.serde.setOnRow(entity.__meta.data, row));
+        return row;
+      });
+
+      const ids = await this.knex.batchInsert(meta.tableName, rows).returning("id");
```
stephenh: Only doing `INSERT`s, `UDPATE`s are wip.

PR #3, opened 2020-04-02
Reproduce flush-in-loop + write up.

stephenh:


flush-in-loop-behavior.markdown
```diff
@@ -0,0 +1,48 @@
...
+We've `N+1`'d.
+
+Ways to fix this:
+
+1. We can get fancy and combine Q2 and Q3 into a single query
+
+    I.e. `findOneOrFail` combines any `find`s within the same event loop into something like:
+
+    ```sql
+    SELECT "Q1" as query_tag, ... FROM publishers WHERE id = 3`
```
stephenh, line 36: In this case, the queries are so simple that only 0 or 1 rows are returned, but potentially it could be a `id > ...` or whatever, that returns N rows.

flush-in-loop-behavior.markdown
```diff
@@ -0,0 +1,48 @@
...
+* `em.find` issues Q1 `SELECT * FROM authors WHERE id IN (1, 2);`
+* Q1 returns two authors, `.map` creates lambdas for both
+* Lambda 1 issues Q2 `SELECT * FROM publishers WHERE id = 3`, `await`s
+* Lambda 2 issues Q3 `SELECT * FROM publishers WHERE id = 4`, `await`s
+* Q2 returns, Lambda 1 modifies it's `a`, calls `em.flush`
+* Lambda 1's `flush` issues Q4 `UPDATE authors SET first_name = ...`
+* Q3 returns, Lambda 2 modifies it's `a`, calls `em.flush`
+* Lambda 2's `flush` issues Q4 `UPDATE authors SET first_name = ...`
+
+We've `N+1`'d.
```
stephenh, line 27: I.e. Joist's general ethos/goal of "we'll bulk update as many operations as possible" didn't happen, where if this was a loop with, say, 50 authors, we'd probably end up with 50 `UPDATE` calls instead of a single `UPDATE`.

Granted, some/most ORMs don't do bulk `UPDATE`s at all, but it's part of Joist's mental model, so I think it's surprising to user's when this doesn't happen. I.e. a causal read of the `.map` it _seems_ like that `em.flush()` should happen "at the same time" / be N+1 safe between both lambdas.

flush-in-loop-behavior.markdown
```diff
@@ -0,0 +1,48 @@
...
+
+Take this example:
+
+```typescript
+const authors = await em.find(Author, { id: ["1", "2"] } );
+await Promise.all(
+  authors.map(async (a) => {
+    const p = await em.findOneOrFail(Publisher, { id: a.publisher.id });
+    a.firstName = a.firstName + p.name;
+    await em.flush();
```
stephenh, line 11: Granted, this example is somewhat contrived b/c the `em.flush()` could just be moved outside the loop, but often times you'll have business logic that is still inside the loop, but after the `em.flush`, i.e. you needed to flush to get some ids assigned or something, and then want both lambdas to continue doing work.

stephenh:
Implemented on master.

PR #10, opened 2020-06-03
hook rewrite

zgavin:
WIP
- refactored hooks to run in a loop so that entities added/deleted/changed in a hook are hooked in turn then flushed
- added `beforeDelete` hook
- added 'addCascadedDelete` rule that employs `beforeDelete`
- disallowed simultaneous `em.flush` calls
- disallowed modifying entities during a flush 
- added entity/reference/collection proxies that allow hooks to still edit objects during a flush


packages/orm/src/EntityProxy.ts
```diff
@@ -0,0 +1,123 @@
...
+      const getter = prototype.__lookupGetter__(property);
+      if (getter !== undefined) {
+        value = getter.apply(receiver);
+      } else {
+        value = (entity as any)[property];
+      }
+
+      if (isPlainFunction(value)) {
+        return function (...args: any[]) {
+          return value.apply(receiver, ...args);
```
stephenh, line 94: I was refreshing my memory on proxies and was reading this:

https://exploringjs.com/es6/ch_proxies.html#sec_proxies-explained ... see the section "28.3.6.1 Wrapping an object affects this". They have an example that I flushed out to be:

```typescript
class Target {
  foo() {
    return this.bar();
  }
  bar() {
    return 2;
  }
}

describe("foo", () => {
  it("foo", () => {
    const target = new Target();
    const seenProps: string[] = [];
    const handler: ProxyHandler<any> = {
      get(target, prop) {
        seenProps.push(String(prop));
        return Reflect.get(target, prop);
      },
    };

    target.foo();
    expect(seenProps).toEqual([]);

    const proxy = new Proxy(target, handler);
    proxy.foo();
    expect(seenProps).toEqual(["foo", "bar"]);
  });
});
```

I.e. it looks like the `this` within proxied function calls is already the proxy... 

Does that simplify some of what you're trying to do?

stephenh: It seems crazy that this works:

```
class Target {
  ref = new Ref();
  foo() {
    return this.bar();
  }
  bar() {
    return this.ref;
  }
}
class Ref {
  zaz() {
    return (this as any).__secret;
  }
}

describe("foo", () => {
  it("foo", () => {
    const target = new Target();
    const seenProps: string[] = [];
    const handler: ProxyHandler<any> = {
      get(target, prop) {
        console.log(`GETGET ${String(prop)}`);
        seenProps.push(String(prop));
        if (prop === "__secret") {
          return (this as any).knownSecret;
        }
        const r = Reflect.get(target, prop);
        return typeof r === "object" && r ? new Proxy(r, handler) : r;
      },
      apply(target, prop) {
        console.log(`APPLY ${prop}`);
      }
    };

    target.foo();
    expect(seenProps).toEqual([]);

    (handler as any).knownSecret = "secret";
    const proxy = new Proxy(target, handler);
    const foundSecret = proxy.foo().zaz();
    expect(seenProps).toEqual(["foo", "bar", "ref", "zaz", "__secret"]);
    expect(foundSecret).toEqual("secret");
  });
});
```

It's like they built this aspect of proxies with our use case specifically in mind...

packages/orm/src/BaseEntity.ts
```diff
@@ -86,6 +86,18 @@ export abstract class BaseEntity implements Entity {
     return this.__orm.deleted !== undefined;
   }
 
+  get isDirtyEntity(): boolean {
```
stephenh, line 89: Nice cleanup.

packages/orm/src/BaseEntity.ts
```diff
@@ -86,6 +86,18 @@ export abstract class BaseEntity implements Entity {
...
+    return Object.keys(this.__orm.originalData).length > 0;
+  }
+
+  get isPendingFlush(): boolean {
```
stephenh, line 93: I wouldn't mind keeping this a little more internal vs. obviously accessible in the entities public API, but not a huge deal.

packages/orm/src/EntityManager.ts
```diff
@@ -482,30 +492,48 @@ export class EntityManager {
...
-    // We defer doing this cascade logic until flush() so that delete() can remain synchronous.
-    await this.cascadeDeletesIntoUnloadedCollections();
-    recalcDerivedFields(this.entities);
-    const entityTodos = sortEntities(this.entities);
-    await addReactiveAsyncDerivedValues(entityTodos);
-    await recalcAsyncDerivedFields(this, entityTodos);
+    const entitiesToFlush: Entity[] = [];
+    let pendingEntities = this.entities.filter((e) => e.isPendingFlush);
+
+    while (pendingEntities.length > 0) {
```
stephenh, line 511: Eventually could put in a sanity check counter like `numberOfLoops` and if it gets over 5, console.log a warn, and if it gets over say 10 or 100, blow up.

packages/orm/src/EntityManager.ts
```diff
@@ -862,7 +891,7 @@ async function addReactiveAsyncDerivedValues(todos: Record<string, Todo>): Promi
       (await followReverseHint([...todo.inserts, ...todo.updates], reverseHint)).forEach((entity) => {
         const todo = getTodo(todos, entity);
         if (!todo.inserts.includes(entity) && !todo.updates.includes(entity) && !entity.isDeletedEntity) {
-          todo.updates.push(entity);
+          todo.updates.push(util.types.isProxy(entity) ? entity : proxyEntity(entity, getEm(entity)["flushSecret"]));
```
stephenh, line 894: Do we need to de-proxy this for the `.includes` checked right before this line?

Er, wait, you're actually creating a proxy...can we keep these as not-proxies and then just create the proxy in `runHook`?

packages/orm/src/EntityManager.ts
```diff
@@ -883,24 +912,32 @@ async function validate(todos: Record<string, Todo>): Promise<void> {
...
   });
   await Promise.all(p);
 }
 
+async function beforeDelete(todos: Record<string, Todo>): Promise<void> {
+  await runHook("beforeDelete", todos, ["deletes"]);
+}
+
+async function beforeFlush(todos: Record<string, Todo>): Promise<void> {
+  await runHook("beforeFlush", todos, ["inserts", "updates"]);
```
stephenh, line 936: Slick `runHook` method.

packages/orm/src/EntityPersister.ts
```diff
@@ -140,14 +141,11 @@ function cleanSql(sql: string): string {
 export function sortEntities(entities: Entity[]): Record<string, Todo> {
   const todos: Record<string, Todo> = {};
   for (const entity of entities) {
-    const isNew = entity.id === undefined;
-    const isDirty = !isNew && Object.keys(entity.__orm.originalData).length > 0;
-    const isDelete = !isNew && entity.__orm.deleted === "pending";
```
stephenh, line 145: Nice clean up.

packages/orm/src/EntityPersister.ts
```diff
@@ -163,6 +161,15 @@ export function getTodo(todos: Record<string, Todo>, entity: Entity): Todo {
   let todo = todos[meta.type];
   if (!todo) {
     todo = { metadata: entity.__orm.metadata, inserts: [], updates: [], deletes: [], validates: [] };
+    [todo.inserts, todo.updates, todo.deletes, todo.validates].forEach((array) => {
+      array.includes = function (entityOrProxy: any) {
```
stephenh, line 165: Oh I see...

packages/orm/src/EntityManager.ts
zgavin: somehow non-proxied entities were sneaking in here.  This was causing the code to blow up later when trying to fetch some meta data that proxies are bad at.  This was a quick and dirty fix, but should probably figure out a better one.

packages/orm/src/index.ts
```diff
@@ -324,5 +371,5 @@ export function configureMetadata(metas: EntityMetadata<any>[]): void {
 }
 
 export function getEm(entity: Entity): EntityManager {
-  return entity.__orm.em;
+  return isFlushProxy(entity) ? proxyEntityManager(entity.__orm.em, entity.__flushSecret) : entity.__orm.em;
```
stephenh, line 374: Ah sure, needing to hook this makes sense...

packages/orm/src/FlushProxy.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    if (property === "entity") {
+      return proxyEntity(entity, flushSecret);
+    }
+
+    return Reflect.get(loader, property, receiver);
+  });
+}
+
+export function proxyEntityManager(em: EntityManager, flushSecret: number): FlushProxy<EntityManager> {
```
stephenh, line 74: Hrm. Could we remove all of the `proxyLoaderResult` / `proxyEntity` / `proxyEntityManager` specific logic? I'm thinking ...probably? Given that demo of "if you keep returning a proxy with your same handler, and you can see anything they access through `this` N-levels deep"...

I could see needing to inspect the `Reflect.get` return value to handle a few cases like arrays and promises...but other than that, do we care about not proxying a few extra methods? (I.e. AFAICT the main goal of having these per-type proxies is to a) know about promises (but could check that generically via `then` detection and b) not proxying things that we don't think we need to).

I suppose if you wanted to early return from a proxy, what about proxying every method, but then stopping if the return value is not an array/promise/Entity/EntityManager/Reference?

packages/orm/src/FlushProxy.ts
```diff
@@ -0,0 +1,120 @@
...
+  return createFlushProxy(em, flushSecret, (em, property, receiver) => {
+    const value = Reflect.get(em, property, receiver);
+
+    // async functions that return entities
+    if (
+      typeof property === "string" &&
+      ["load", "loadAll", "loadFromQuery", "find", "findOne", "findOneOrFail", "findOrCreate"].includes(property)
+    ) {
+      return function (...args: any[]) {
+        return value.apply(receiver, args).then((result: any) => proxyLoaderResult(result, flushSecret));
```
stephenh, line 84: I.e. here instead of checking method names, go ahead and let `Reflect.get` do it's thing, it value is a function, return a new function that calls that 1st function, and if the return value is a promise, `then`s it, and then just returns a new proxy...

packages/orm/src/FlushProxy.ts
stephenh: (re-articulating for my own benefit... ideally the `block` param of `createFlushProxy` would go away, and it could generically handle any result that it traverses across, by defaulting to always re-proxying it's own handler again/again at each step, and ideally special casing as little as possible, i.e. maybe only promises? and that is it?)

stephenh: The idea being both in terms of simplification/less LOC but also robustness against future changes to the entity/EM/loader APIs.

packages/orm/src/FlushProxy.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  return createFlushProxy(entity, flushSecret, (entity, property, receiver) => {
+    const value = Reflect.get(entity, property, receiver);
+
+    if (
+      value instanceof OneToManyCollection ||
+      value instanceof ManyToManyCollection ||
+      value instanceof ManyToOneReference
+    ) {
+      if (loaderCache[property] === undefined) {
```
stephenh, line 111: In theory these proxies are cheap, so maybe punt on the `loaderCache` for now?

packages/orm/src/FlushProxy.ts
```diff
@@ -0,0 +1,120 @@
...
+    return target;
+  }
+
+  return new Proxy(target, {
+    get(target, property, receiver) {
+      if (property === "__flushSecret") {
+        return flushSecret;
+      }
+
+      if (property === "__proxyTarget") {
```
stephenh, line 45: I'm cool with this trap, and I've done it several times before, just copy/pasting that I also saw this way:

```
const proxies = new WeakSet();

export function createProxy(obj) {
    const handler = {};
    const proxy = new Proxy(obj, handler);
    proxies.add(proxy);
    return proxy;
}

export function isProxy(obj) {
    return proxies.has(obj);
}
```

Or doing proxy detection and de-wrapping that seems fine too.

packages/orm/src/FlushProxy.ts
zgavin: The problem here isn't the promise, it's the eventually result.  Those are separate objects that aren't going to get proxied naturally, so we need to do it ourselves.

zgavin: Disregard the previous, i think i understand what you're saying better now.

stephenh: >  The entire entity chain needs to be composed of proxies or things won't work

Right, I didn't mean stop doing proxies, if anything I meant "proxy all the things" with less checks for "is the method we're going x or y or z"...

Like, if you could/as a starting pointer, have a handler like:

```
handler = get(target, prop) {
  // trap __sessionKey and __proxyTarget 
  value = Reflect.get(target, prop)
  return new Proxy(value, handler);
}
``` 

Naively, wouldn't that work? Or at least as a starting point?

zgavin: Maybe if this weren't javascript it would work 🤣  Proxy behaves weirdly with some data types (eg Date) and won't work at all with primitives.  I think get what you're after though.  

I've stripped out the `proxyEntity`/`proxyEntityManager`/`proxyLoader` methods locally and replaced them with the much more generic:

```typescript
function maybeProxy<T extends any>(object: T, flushSecret: number): any {
  if (typeof object === "function" && !/^class\s/.test(object.toString())) {
    return function (this: any, ...args: any[]) {
      return maybeProxy(object.apply(this, args), flushSecret);
    };
  } else if (typeof object === "object" && object !== null) {
    if (object instanceof Promise) {
      return object.then((result) => maybeProxy(result, flushSecret));
    } else if (Array.isArray(object)) {
      return object.map((element) => maybeProxy(element, flushSecret));
    } else if (
      object instanceof EntityManager ||
      object instanceof BaseEntity ||
      object instanceof OneToManyCollection ||
      object instanceof ManyToManyCollection ||
      object instanceof ManyToOneReference
    ) {
      return createFlushProxy(object, flushSecret);
    }
  }

  return object;
}

export function createFlushProxy<T extends object>(target: T, flushSecret: number): FlushProxy<T> {
  if (isFlushProxy(target)) {
    return target;
  }

  return new Proxy(target, {
    get(target, property, receiver) {
      if (property === "__flushSecret") {
        return flushSecret;
      }

      if (property === "__proxyTarget") {
        return target;
      }

      // knex seems to really dislike being put through maybeProxy
      if (property === "knex") {
        return (target as any).knex;
      }

      return maybeProxy(Reflect.get(target, property, receiver), flushSecret);
    },
  }) as FlushProxy<T>;
}
```

This is failing on 4 tests, all related to hooks.  It's tricky to debug.

stephenh: Ah yeah, sorry about the overlapping comment, I had it drafted and didn't see your reply. 

Yep, that's exactly what I'm wondering if it would work...

Why does your `maybeProxy` not have a `new Proxy` call it in? Oh I see, it's only for that last "else if"... this is going to sound super-naive, but what if `maybeProxy` was not `maybeProxy` and instead `alwaysProxy`, i.e change `return maybeProxy(Reflect.get...` to `return createFlushProxy(Reflect.get)`...

stephenh: How about a `FlushProxy.test.ts` that creates an EM/entity, manually wraps the proxy, and then starts various call chains of `entity.__flushSecret`, `author.books.load()...[0].__flushSecret`, `author.publisher.get.__flushSecret`, etc. that you want to all expect to work.

stephenh: ...the test might have to live in the integration-tests to be able to use the domain objects. But that's probably fine.

zgavin: Many things in javascript cannot be proxied (eg primitives, it's an immediate error), behave unpredictably (eg, Date), or break in other ways (eg, class constructors).

stephenh: >  Proxy behaves weirdly with some data types (eg Date) and won't work at all with primitives

Yeah, it's annoying that it's not easier to differentiate primitives/value-ish-objects from classes / object literals...

stephenh: I poked at this just enough locally to see that, right, you pretty quickly hit things like ` Proxy behaves weirdly with some data types (eg Date) and won't work at all with primitives`...

stephenh: Huh. Wonder if this is applicable:

https://github.com/tc39/ecma262/issues/1114
https://tvcutsem.github.io/js-membranes
https://github.com/ajvincent/es-membrane

stephenh: After thinking on it, yeah, since proxies seem to have more boundary cases than expected, and the bespoke-proxy-per-loader/entity/EM was working/passing tests, I'm good with going with that approach. I might ask for that `FlushProxy.test.ts` to make future iteration on this easier, but otherwise this looks great.

(I spent ~15-20m looking at the membrane stuff and it seems non-trivial? Was hoping it'd be like ~10 lines of code to drop it in and do what we want, but doesn't seem like it...)

stephenh: https://github.com/salesforce/observable-membrane looks like a much nicer membrane API...

https://github.com/salesforce/observable-membrane/pull/48 is interesting, there is a comment from jdalton, who created lodash, and the github popup says hes as salesforce as a UI security engineer. Afaiu a usecase of membranes is to provide secure/protected views of an object graph to unprotected/untrusted code, so I assume that is what their library is mainly for.

stephenh: Huh, I scanned the source code of the `observable-membrane` project and it's surprisingly short for what it asserts to do...

stephenh: "This membrane implementation is tailored for observable objects, which are objects with the prototype chain set to Object.prototype or null. Any other object is not wrapped in a Proxy by design." ah shoot from `observable-membrane`...

stephenh: https://github.com/salesforce/observable-membrane/issues/50

packages/orm/src/EntityManager.ts
```diff
@@ -139,24 +147,15 @@ export type LoaderCache = Record<string, DataLoader<any, any>>;
 type FilterAndOrder<T> = [FilterOf<T>, OrderOf<T> | undefined];
 
 export class EntityManager {
-  constructor(public knex: Knex) {}
+  constructor(public knex: Knex) {
+    this.contexty.create();
```
stephenh, line 151: Hm, my understanding is that we'd call this `.create()` just once as a module-level constant, b/c it installs a "static" hook into the async_hooks runtime...

packages/orm/src/EntityManager.ts
```diff
@@ -139,24 +147,15 @@ export type LoaderCache = Record<string, DataLoader<any, any>>;
...
-      await this.currentFlushPromise;
-      this.currentFlushPromise = undefined;
-      return keys;
-    },
-    { cache: false },
-  );
-  private currentFlushPromise?: Promise<void>;
+  private flushSecret: number = 0;
+  private _isFlushing: boolean = false;
+  private contexty = new Contexty();
```
stephenh, line 158: Er, this is the that that is a module-level const. And looking at the source, they only install the hook once, which is good, but I think this `new Contexty` is the equivalent of `private static final ThreadLocal foo = new ThreadLocal()` that only needs instantiated once and then within handles multiplexing whoever called it back their right result.

packages/orm/src/EntityManager.ts
```diff
@@ -482,30 +481,46 @@ export class EntityManager {
...
-    await addReactiveAsyncDerivedValues(entityTodos);
-    await recalcAsyncDerivedFields(this, entityTodos);
+    this._isFlushing = true;
+
+    // We need to split the thread so that Node's executionAsyncId changes and in turn
+    // Contexty give us a context that's distinct from whatever called flush
+    await new Promise((res) => setTimeout(res, 0));
+    const context = this.contexty.create();
+
+    const entitiesToFlush: Entity[] = [];
```
stephenh, line 495: Would we should make this a `Set` just for good measure.

packages/orm/src/EntityManager.ts
stephenh: Separate from moving the `new Contexty()` up to be a module-level const, do we need this `.create()` call in the constructor? Maybe but just want to understand.

packages/orm/src/collections/OneToManyCollection.ts
```diff
@@ -44,13 +42,15 @@ export class OneToManyCollection<T extends Entity, U extends Entity> extends Abs
 
   add(other: U): void {
     ensureNotDeleted(this.entity);
-    if (this.loaded === undefined) {
-      if (!this.addedBeforeLoaded.includes(other)) {
-        this.addedBeforeLoaded.push(other);
+    const loaded = this.loaded;
```
stephenh, line 45: Hm. Maybe if you want to get rid of the `this.` prefixes, do this as:

```javascript
const { loaded, addedBeforeLoaded } = this;
```

packages/orm/src/collections/OneToManyCollection.ts
```diff
@@ -167,22 +167,27 @@ export class OneToManyCollection<T extends Entity, U extends Entity> extends Abs
   }
 
   private maybeAppendAddedBeforeLoaded(): void {
-    if (this.loaded) {
-      const newEntities = this.addedBeforeLoaded.filter((e) => !this.loaded?.includes(e));
-      this.loaded.unshift(...newEntities);
+    const loaded = this.loaded;
```
stephenh, line 170: Same for the destructure on `this`.

packages/orm/src/index.ts
```diff
@@ -107,7 +112,21 @@ interface Flavoring<FlavorT> {
 export type Flavor<T, FlavorT> = T & Flavoring<FlavorT>;
 
 export function setField(entity: Entity, fieldName: string, newValue: any): void {
-  ensureNotDeleted(entity);
+  ensureNotDeleted(entity, { ignore: "pending" });
+  const em = getEm(entity);
+
+  if (em.isFlushing) {
+    const { context } = em["contexty"];
```
stephenh, line 119: This is fine but I think the idea with contexty is that this could also be `const { context } = the-module-const`.

packages/orm/src/collections/OneToManyCollection.ts
zgavin: This is actually just leftover changes from my attempts to proxify these functions.  I can revert to the original.

stephenh: Cool, any of reverting or leaving or destructuring is fine.

packages/orm/src/EntityManager.ts
zgavin: We don't need it necessarily, but it makes it so that code checking the context doesn't need to worry about there not being one yet.

packages/orm/src/EntityManager.ts
zgavin: Actually on second thought, I don't think this is accurate.   A module level contexty would not be able to hand back the right result in the case of multiple entity managers.

packages/orm/src/EntityManager.ts
zgavin: we could I suppose. The underlying `em.entities` is not a `Set`, though, so it would seem a little weird to make it one at this level instead.

packages/orm/src/EntityManager.ts
```diff
@@ -482,30 +481,46 @@ export class EntityManager {
...
+      await addReactiveAsyncDerivedValues(entityTodos);
+      await recalcAsyncDerivedFields(this, entityTodos);
+
+      // TODO Run beforeFlush first, so it can fill in derived values for validate?
+      await addReactiveValidations(entityTodos);
+      await validate(entityTodos);
+      await beforeDelete(entityTodos);
+      await beforeFlush(entityTodos);
+
+      entitiesToFlush.push(...pendingEntities);
```
stephenh, line 513: Per the `Set` comment, couldn't something in `pendingEntities` already be in `entitiesToFlush`? ...oh, I see, ont he next line, we explicitly don't add entities that were already in it. Cool.

PR #13, opened 2020-06-09
Run em.delete on loaded-entity cascaded deletes.

stephenh:


packages/integration-tests/src/EntityManager.test.ts
```diff
@@ -757,6 +756,18 @@ describe("EntityManager", () => {
     expect(rows.length).toEqual(0);
```
stephenh, line 756: Should this test assert that the beforeFlush hook was run? I.e. right now the cascade delete could be happening via a db fk cascade and the test would still pass / not realize that it's not our beforeFlush that is doing the cascade.

packages/orm/src/collections/ManyToManyCollection.ts
```diff
@@ -150,7 +150,7 @@ export class ManyToManyCollection<T extends Entity, U extends Entity> extends Ab
 
   /** Some random entity got deleted, if it was in our collection, remove it. */
   onDeleteOfMaybeOtherEntity(maybeOther: Entity): void {
-    ensureNotDeleted(this.entity);
+    ensureNotDeleted(this.entity, { ignore: "pending" });
```
stephenh, line 153: Another data point for `ensureNotDeleted` should either always ignore pending or just go away...

packages/integration-tests/src/entities/Author.test.ts
```diff
@@ -80,10 +80,10 @@ describe("Author", () => {
     const b1 = await em.load(Book, "1");
     // When we change the book
     b1.title = "b2";
-    // And the book that it would reactively trigger is deleted
+    // And the author that it would reactively trigger is deleted
     em.delete(a1);
-    // Then we can't b/c the book.author is required
-    await expect(em.flush()).rejects.toThrow("Validation error: author is required");
+    // Then it works b/c the author cascade delete takes precedence
+    await em.flush();
```
stephenh, line 86: Make sense that the `Author name != book name` validation rule isn't run anymore...makes me wonder if we need a boundary case that uses different entities where a "parent"/reactive rule in one deleted entity doesn't run (and specifically doesn't stack trace while realizing it should not run) when the non-cascade-deleted entity that triggers it changes.

PR #14, opened 2020-06-12
Added CustomReference

zgavin:
CustomReference encapsulates references that need some more complex logic than ManyToOneReference

### DIFF
--- a/packages/integration-tests/src/collections/CustomReference.test.ts
+++ b/packages/integration-tests/src/collections/CustomReference.test.ts
@@ -1,0 +1,80 @@
+import { EntityManager } from "joist-orm";
+import { knex, numberOfQueries, resetQueryCount } from "../setupDbTests";
+import { Author, Book, BookReview } from "../entities";
+import { insertAuthor, insertBook, insertBookReview, insertPublisher } from "../entities/factories";
+
+describe("CustomReference", () => {
+  it("can load a reference", async () => {
+    await insertAuthor({ first_name: "f" });
+    await insertBook({ title: "t", author_id: 1 });
+    await insertBookReview({ rating: 5, book_id: 1 });
+
+    const em = new EntityManager(knex);
+    const review = await em.load(BookReview, "1");
+    const author = await review.author.load();
+    expect(author.firstName).toEqual("f");
+  });
+
+  it("can populate a reference", async () => {
+    await insertAuthor({ first_name: "f" });
+    await insertBook({ title: "t", author_id: 1 });
+    await insertBookReview({ rating: 5, book_id: 1 });
+
+    const em = new EntityManager(knex);
+    const review = await em.load(BookReview, "1", "author");
+    expect(review.author.get.firstName).toEqual("f");
+  });
+
+  it("can set a reference", async () => {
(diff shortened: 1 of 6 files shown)


packages/integration-tests/src/collections/CustomReference.test.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    const rows = await knex.select("*").from("books");
+    expect(rows[0].author_id).toEqual(1);
+  });
+
+  it("can set a reference through setFromOpts", async () => {
+    const em = new EntityManager(knex);
+    const author = em.create(Author, { firstName: "a1" });
+    const book = em.createUnsafe(Book, { title: "t1" });
+    const review = em.createUnsafe(BookReview, { book, rating: 5, author } as any);
```
stephenh, line 44: Huh. That's fancy. It's not technically in the Opts type, but the `as any` makes it work?

packages/integration-tests/src/entities/BookReview.ts
```diff
@@ -1,7 +1,30 @@
...
+    never
+  >(this, "author", {
+    load: async (review) => {
+      await review.load((r) => r.book.author);
+    },
+    get: (review) => review.book.get.author.get,
+    set: (review, author) => {
+      review.book.get.author.set(author);
+    },
+    setFromOpts: (review, author) => {
```
stephenh, line 19: Can we skip `setFromOpts` being a hook and have CustomReference just implement that internally and call `.set`?

I think `setFromOpts` means special things for other collections, but maybe/probably/hopefully not for our current CustomReference usage?

packages/orm/src/collections/CustomReference.ts
```diff
@@ -0,0 +1,126 @@
...
+  constructor(public entity: T, public fieldName: keyof T, private opts: CustomReferenceOpts<T, U, H, N>) {
+    super();
+  }
+
+  private filterDeleted(entity: U | N, opts?: { withDeleted?: boolean }): U | N {
+    return opts?.withDeleted === true || entity === undefined || !entity.isDeletedEntity ? entity : (undefined as N);
+  }
+
+  private doGet(opts?: { withDeleted?: boolean }): U | N {
+    ensureNotDeleted(this.entity, { ignore: "pending" });
```
stephenh, line 32: Separately, do all of our `ensureNotDeleted` pass ignore pending at this point? If they're not, should they/should the param go away and it just always ignores pending?

packages/integration-tests/src/entities/BookReview.ts
zgavin: will remove it for now

packages/orm/src/collections/CustomReference.ts
zgavin: Some of our calls do not pass `ignore: "pending"`.  For example, o2m's `add` and `set` both still just have `ensureNotDeleted(this.entity)`, which is, imo, correct.

packages/integration-tests/src/collections/CustomReference.test.ts
zgavin: yup

PR #16, opened 2020-06-19
Set defaults from database on create

zgavin:


packages/codegen/src/generateEntityCodegenFile.ts
```diff
@@ -238,6 +244,47 @@ export function generateEntityCodegenFile(config: Config, meta: EntityDbMetadata
   `;
 }
 
+function fieldHasDefaultValue(
+  config: Config,
+  meta: EntityDbMetadata,
+  field: EntityDbMetadata["primitives"] extends Array<infer U> ? U : never,
```
stephenh, line 250: Huh, what's this extends + infer doing? Are we trying to constrain that the runtime `field` value of comes the `meta.primitives` column? I'm not sure we have the type information to enforce that.

packages/codegen/src/generateEntityCodegenFile.ts
```diff
@@ -238,6 +244,47 @@ export function generateEntityCodegenFile(config: Config, meta: EntityDbMetadata
...
+  if (
+    ormMaintainedFields.includes(fieldName) ||
+    isDerived(config, meta.entity, fieldName) ||
+    isAsyncDerived(config, meta.entity, fieldName)
+  ) {
+    return false;
+  }
+
+  // try to validate that we actually got a primitive value and not arbitrary SQL
+  return (
```
stephenh, line 273: Wouldn't mind a unit test ... in the codegen directory itself? That threw some input at this and see what happens. Not a big deal given Joist's historical reliance on integration-tests for most...everything test related.

packages/codegen/src/generateEntityCodegenFile.ts
```diff
@@ -205,7 +211,7 @@ export function generateEntityCodegenFile(config: Config, meta: EntityDbMetadata
       ${[o2m, m2o, m2m]}
 
       constructor(em: ${EntityManager}, opts: ${entityName}Opts) {
-        super(em, ${metadata});
+        ${hasDefaultValues ? code`super(em, ${metadata}, {...${defaultValuesName}})` : code`super(em, ${metadata})`};
```
stephenh, line 214: Cool approach with the `...`

packages/integration-tests/src/entities/BookCodegen.ts
```diff
@@ -67,6 +67,8 @@ export interface BookOrder {
   author?: AuthorOrder;
 }
 
+export const BookDefaultValues = { order: 0 };
```
stephenh, line 70: Nit picky but lower case `bookDefaultValues`?

packages/codegen/src/generateEntityCodegenFile.ts
zgavin: primitives is defined as `PrimitiveField[]`  but we don't export `PrimitiveField` from in `EntityDbMetadata.ts`, so this allows us to use that type without needing to export it.  I could just add the export if you'd rather i do that.

packages/codegen/src/generateEntityCodegenFile.ts
zgavin: I was initially just passing the DefaultValues object directly, which broke everything because every entity of a class ended up having the same object for `__orm.data` 😂

packages/integration-tests/src/entities/BookCodegen.ts
zgavin: Will do, I wasn't sure what the convention here might be.

packages/codegen/src/generateEntityCodegenFile.ts
zgavin: since we don't have a `.test.ts` for this file yet it seems like making one would probably be enough work to be a PR all on it's own.  Since this PR is relatively simple i'd prefer to just rely on integration-tests.

packages/codegen/src/generateEntityCodegenFile.ts
stephenh: Ah cool. Does just `EntityDbMetadata["primitives"]` work? I was mostly curious about the `Array<infer U>` seeming unnecessary/or else I'm missing something.

packages/codegen/src/generateEntityCodegenFile.ts
stephenh: Sgtm.

packages/codegen/src/generateEntityCodegenFile.ts
zgavin: No, because EntityDBMetaData is an array.  So we need to extract out the generic argument of the array with infer.  This was the only syntax i could get to work, although i've just tried `EntityDbMetadata["primitives"] extends (infer U)[] ? U : never` and it also works.  Do you have a preference?  Just exporting the type is also still an option.

stephenh: I think `EntityDBMetadata["primitives][]` or `EntityDBMetadata["primitives"][0]` is supposed to work? I forget the exact incantation.

zgavin: Neat, `EntityDBMetadata["primitives"][0]` does actually work as well.  I'm not sure which option is the most clear/least magical, though:

`EntityDBMetadata["primitives"][0]`
`EntityDBMetadata["primitives"] extends (infer U)[] ? U : never`
`EntityDBMetadata["primitives"] extends Array<infer U> ? U : never`

or, in 

`export type PrimitiveField  = ...` in `EntityDbMetadata.ts`

stephenh: `EntityDBMetadata["primitives"][0]` is pretty okay-ish/idiomatic in TS code, once you learn/see it b/c its somewhat unintuitive, so I'm +1 with that, but also +1 to the exported type.

zgavin: Gonna go with just exporting all those types.  I'm not sure there's any benefit to them being private from the rest of joist.

PR #17, opened 2020-06-21
Implement tagged ids

stephenh:
Somewhat proof-of-concept but it seems to be working pretty well/somewhat tempted to go ahead with it. At least don't see a reason to rule it out so far.

Basically our entity `.id`s already "look like strings" (still ints in the db), i.e. "1" / "2" / "3", and this changes the strings to be `"author:1"`, `"author:2"`, etc., so that as we pass around the id to the client / back to the server, we "really really really" know what entity it's for, and can verify that we didn't pass around the wrong id.

Historically when discussing tagged ids with other/pre-Homebound teams (i.e. at Remind, where a few of the "fancy tech shops" like FB/LinkedIn use tagged ids), the con of "well, that'd be a lot of boilerplate to go back/forth from tagged string / int-in-the-db" was one of the largest cons/blockers.

With this PR, all of that mapping basically goes away; ids stay as primary key ints in the db, but Joist turns them into tagged ids for free, and de-tags the ids before any SQL operations like insert/select/update.

Currently each tag is just the entity name, i.e. `author:1`, `publisher:2`, etc. The `tagName` is stored in the codegen'd `metadata.ts`, so we can easily change it at codegen time, i.e. by config values in `joist-codegen.json`. (Not entirely sure we want that to be configurable, although I suppose this brings up, how to handle table renames? Would prior `oldEntityName:1` tagged ids have to continue to be respected and mapped to the new `newEntityName:1` tag/entity automatically by Joist? Probably, i.e. in case ids have been stored externally in vendor systems / something / something.) 

The existing logic/slice points for "change int primary key to string id" we were already doing for playing nicely in the GQL ecosystem (to match their `ID`) meant this was actually pretty simple to do; i.e. `keyToNumber` and `keyToString` were already called in all the right spots, they just needed to be taught how to tag and un-tag string ids going back/forth to numbers.

While doing this, besides just dropping the tag prefix, i.e. `author:1` --> `1`, the logic also ensures the tag was the expected tag, i.e. if the code thought this should be a `publisher:...` and it got an `author:...` id, then it will immediately fail.

This is essentially the runtime version of our strongly typed/flavor ids, but even more powerful since the checking is based on the id value, so will persist across "hops to strings/GQL results/GQL inputs/query parameters/cookies/whatever".

Note that we still accept un-tagged ids to `.load` and `.find`, so this is not a breaking change in terms of inputs (if for whatever reason code/tests continues to pass the untagged value). However, it is a breaking change in terms of outputs, i.e. `Entity.id` is now `author:1` instead of `1`.

At some point we could disallow passing untagged ids all together.

### DIFF
--- a/packages/integration-tests/src/EntityManager.lens.test.ts
+++ b/packages/integration-tests/src/EntityManager.lens.test.ts
@@ -96,6 +96,6 @@ describe("EntityManager.lens", () => {
     const em = new EntityManager(knex);
     const b1 = await em.load(Book, "1");
     const p1Id = await b1.load((b) => b.author.publisher.idOrFail);
-    expect(p1Id).toEqual("publisher:1");
+    expect(p1Id).toEqual("p:1");
   });
 });
--- a/packages/integration-tests/src/EntityManager.lens.test.ts
+++ b/packages/integration-tests/src/EntityManager.lens.test.ts
@@ -88,4 +88,14 @@ describe("EntityManager.lens", () => {
     const publisherName: string | undefined = await a1.load((a) => a.publisher.name);
     expect(publisherName).toEqual(undefined);
   });
+
+  it("can navigate into getters", async () => {
+    await insertPublisher({ name: "p1" });
+    await insertAuthor({ first_name: "a1", publisher_id: 1 });
+    await insertBook({ title: "b1", author_id: 1 });
+    const em = new EntityManager(knex);
+    const b1 = await em.load(Book, "1");
+    const p1Id = await b1.load((b) => b.author.publisher.idOrFail);
+    expect(p1Id).toEqual("publisher:1");
+  });
 });
--- a/packages/codegen/src/EntityDbMetadata.ts
+++ b/packages/codegen/src/EntityDbMetadata.ts
@@ -127,6 +127,10 @@ export class EntityDbMetadata {
       .map((r) => newManyToManyField(r));
     this.tableName = table.name;
   }
+
+  get name(): string {
+    return this.entity.name;
+  }
 }
 
 function isMultiColumnForeignKey(r: M2ORelation) {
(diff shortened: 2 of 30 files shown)


packages/integration-tests/src/EntityManager.lens.test.ts
```diff
@@ -88,4 +88,14 @@ describe("EntityManager.lens", () => {
     const publisherName: string | undefined = await a1.load((a) => a.publisher.name);
     expect(publisherName).toEqual(undefined);
   });
+
+  it("can navigate into getters", async () => {
```
stephenh, line 92: Unrelated new test case.

stephenh:
> So if I'm understanding this correctly, all GQL calls to an ID will now return <entityName>:<id>?

Yep.

> Wouldn't this require some work on the UIs to handle correctly? For instance to
> avoid URLs like /project/project:1/?

Well, good question. I agree those URLs "look weird" but I think the idea is that the UIs would _not_ try to strip the tag, but the whole point of the tag is to make sure the UI/backend/etc. don't get confused about what id the value is. I.e. seeing `project:1` in the URL is a feature/the purpose of the tags.

That said, yes, you don't usually see URLs that look like that. 

...if our `tagDelimiter` was `/` then something like `/${project.id}` would automatically be the URL `/project/1`. :-) But that would probably confuse react-router quite a bit, because we'd want the path for that URL to still return `project/1` (i.e. `project:1` with the current delimiter) as the `:projectId` path variable, and not "just 1 by itself".

tldr, yes, if we're fine with `/project/project:1/` URLs, this should all be transparent to the frontend/react-router/etc. and "just work". But yeah it "looks weird". Not sure if that's a deal breaker. I don't think we'd want a bunch of boilerplate hacks around it, but maybe there is a way to keep tagged ids ... but ... not have the tag look as weird? ...but the tag being "in the id" is the whole point...so, not sure. Will think about it.

PR #18, opened 2020-06-21
Generate graphql-specific glue code/config

stephenh:
In theory I'm working on some of the bills schema, but got side-tracked by our enum/mapped type/etc. boilerplate in the graphql-codegen config/etc.

In theory this cleans all of that up; we'd get an `enums.graphql` that matches our domain enums, `enumResolvers.ts` that does the enum detail pattern, and a `graphql-codegen-joist.js` that has `const`s for the mappers/enum values (I have a local proof-of-concept in graphql-service that changes `graphql-codegen.yaml` to `graphql-codegen.js` to we can `require` in this new `graphql-codegen-joist.ts` file.)


packages/integration-tests/graphql-codegen-joist.js
```diff
@@ -0,0 +1,14 @@
+const mappers = {
```
stephenh, line 1: The integration-tests don't actually use any of these new files, i.e. it hasn't turned into a graphql project, I'm just using it to watch for the output.

Eventually we'll need to have like N `integration-test/foo` directories where each `foo` is some variation of a project config / schema design / etc., and we could have one of them that was very specifically an apollo graphql project or what not.

packages/integration-tests/schema/enums.graphql
```diff
@@ -0,0 +1,2 @@
+enum PublisherSize { SMALL LARGE }
```
stephenh, line 1: In theory prettier supports graphql, so I'm assuming this would get prettier'd in the downstream project. Didn't bother with that for now/here.

PR #19, opened 2020-06-21
super basic jsonb column support

floodfx:
Taking a shot at super basic jsonb column support.  Two tests, one in `EntityManager` which is passing and one in `JsonDatum` that is failing here:
```
TypeError: Cannot set property 'metadata' of undefined

      407 | };
      408 | 
    > 409 | (JsonDatum as any).metadata = jsonDatumMeta;
          |                            ^
      410 | 
      411 | export const publisherMeta: EntityMetadata<Publisher> = {
      412 |   cstr: Publisher,

      at Object.<anonymous> (src/entities/metadata.ts:409:28)
      at Object.<anonymous> (src/entities/entities.ts:17:1)
```
Not sure why?


packages/integration-tests/migrations/1580658856631_author.ts
```diff
@@ -58,4 +58,9 @@ export function up(b: MigrationBuilder): void {
   });
 
   createManyToManyTable(b, "books_to_tags", "books", "tags");
+
+  createEntityTable(b, "json_data", {
```
stephenh, line 62: Ha, cute that this got turned into `json_datum` plural but the pluralize library.

Fwiw I wouldn't mind just putting these on an existing entity, i.e. make up some sort of `books.errata`...or...something silly like that. Although if it's required/`notNull`, there are a lot of tests that already make the book/author/etc. entities and don't know about a json field.

...although now that I think about it, there are `insertAuthor` / `insertBook` helper methods that basically all of the tests use that can fill in test-factory-style defaults.

Keeping `json_data` for now is cool, just thinking. Maybe actually better as a separate entity.

PR #21, opened 2020-06-23
Generate graphql schema files

stephenh:
We guess at what `type Entity { ... }` and `type SaveEntityInput { ... }` should look like and then use a `.history.json` file to remember what we've auto-added to allow the programmer to add/change/remove customizations on top of our initial guess.


packages/graphql-codegen/src/generateGraphqlCodegen.test.ts
```diff
@@ -0,0 +1,18 @@
+import { EntityDbMetadata, EnumRows } from "joist-codgen";
+import { generateGraphqlCodegen } from "./generateGraphqlCodegen";
+
+describe("generateGraphqlCodegen", () => {
```
stephenh, line 4: Hey look unit tests.

packages/graphql-codegen/src/generateGraphqlSchemaFiles.test.ts
```diff
@@ -0,0 +1,224 @@
+import { snakeCase } from "change-case";
+import { EntityDbMetadata, makeEntity, PrimitiveField } from "joist-codegen";
+import { plural } from "pluralize";
+import { generateGraphqlSchemaFiles } from "./generateGraphqlSchemaFiles";
+import { Fs } from "./utils";
+
+describe("generateGraphqlSchemaFiles", () => {
```
stephenh, line 7: This unit test is turning out to be pretty useful b/c the `.history.json` approach means that just running `npm run codegen` in `integration-tests` can't exercise all of the boundary cases of "this is a not-yet-added field / etc.".

PR #22, opened 2020-06-24
Add EntityManager.transaction API

stephenh:


packages/orm/src/EntityPersister.ts
```diff
@@ -12,21 +12,21 @@ export interface Todo {
   validates: Entity[];
 }
 
-export async function flushEntities(knex: Knex, tx: Transaction, todos: Record<string, Todo>): Promise<void> {
+export async function flushEntities(knex: Knex, todos: Record<string, Todo>): Promise<void> {
```
stephenh, line 15: These `tx` params were able to go away b/c the knex `Transaction` class extends `Knex` itself, so we just pass `tx` as the 1st param and then any `knex.whatever` calls are defacto using the txn.

packages/orm/src/EntityManager.ts
```diff
@@ -522,12 +556,25 @@ export class EntityManager {
...
+          // await knex.commit();
+        });
+      } else {
+        await flushEntities(this.knex, entityTodos);
+        await flushJoinTables(this.knex, joinRowTodos);
+        // Defer to the caller to commit the transaction
+      }
+
+      // TODO: This is really "after flush" if we're being called from a transaction that
+      // is going to make multiple `em.flush()` calls?
```
stephenh, line 575: The wrinkle here is that a valid api usage would be:

```typescript
em.transaction(async () => {
  em.find...
  em.create...
  // sends sql to db, doesn't commit
  await em.flush();

  em.find...
  em.create...
  // send more sql to db
  await em.flush()
}); // lambda is over, now txn commit happens
```

In the current impl, our `afterCommit` hooks would be ran for each `em.flush` call.

I'm pretty sure this is fine for this PR / to fix our current blueprint bug, but is something to think about for handling better/differently/potentially at some point.

packages/orm/src/EntityManager.ts
```diff
@@ -435,6 +435,36 @@ export class EntityManager {
...
   }
 
+  /**
+   * Executes `fn` with a transaction, and automatically calls `flush`/`commit` at the end.
+   *
+   * This ensures both any `.find` as well as `.flush` operations happen within the same
+   * transaction, which is useful for enforcing cross-table/application-level invariants that
+   * cannot be enforced with database-level constraints.
+   */
+  public async transaction<T>(fn: () => Promise<T>): Promise<T> {
```
zgavin, line 445: I'm not sure you'd want every `em.transaction` to always have `isolation level serializable`.  Perhaps expose it as an optional argument that defaults to serializeable?

packages/orm/src/EntityPersister.ts
```diff
@@ -61,21 +61,23 @@ async function assignNewIds(knex: Knex, tx: Transaction, todos: Record<string, T
...
-async function batchInsert(knex: Knex, tx: Transaction, meta: EntityMetadata<any>, entities: Entity[]): Promise<void> {
+async function batchInsert(knex: Knex, meta: EntityMetadata<any>, entities: Entity[]): Promise<void> {
   const rows = entities.map((entity) => {
     const row = {};
     meta.columns.forEach((c) => c.serde.setOnRow(entity.__orm.data, row));
     return row;
   });
   // We don't use the ids that come back from batchInsert b/c we pre-assign ids for both inserts and updates.
-  await knex.batchInsert(meta.tableName, rows).transacting(tx);
+  // We also use `.transacting` b/c even when `knex` is already be a Transaction object,
```
zgavin, line 71: Can remove the word "be" from "is already be a Transaction object"

packages/orm/src/EntityManager.ts
```diff
@@ -474,9 +504,13 @@ export class EntityManager {
...
    * Flushes the SQL for any changed entities to the database.
    *
-   * Currently this `BEGIN`s and `COMMIT`s a new transaction on every call;
-   * we should also support an `EntityManager` itself running all queries
-   * (i.e. including the initial `SELECT`s) in a transaction.
+   * If this is run outside of an existing transaction, it will `BEGIN` and `COMMIT`
+   * a new transaction on every `.flush()` call so that all of the `INSERT`s/etc.
+   * happen atomically.
+   *
+   * This is it called with an existing transaction, i.e. `EntityManager.transaction`
```
zgavin, line 511: Think you just want "If this is called with an existing transaction"

Also should remove the reference to `newEntityManager` here since that seems to have gone away.

packages/orm/src/EntityManager.ts
```diff
@@ -522,12 +556,25 @@ export class EntityManager {
...
     if (Object.keys(entityTodos).length > 0 || Object.keys(joinRowTodos).length > 0) {
-      await this.knex.transaction(async (tx) => {
-        await flushEntities(this.knex, tx, entityTodos);
-        await flushJoinTables(this.knex, tx, joinRowTodos);
-        await tx.commit();
-      });
+      const alreadyInTxn = "commit" in this.knex;
+
+      if (!alreadyInTxn) {
+        await this.knex.transaction(async (knex) => {
```
zgavin, line 562: if we changed `em.transaction` so that nested calls just run their callback without spawning a new transaction, then this code could be simplified to remove the check for a transaction.

packages/orm/src/EntityManager.ts
zgavin: maybe just change the name to `afterFlush` instead of `afterCommit`?

packages/orm/src/EntityManager.ts
stephenh: Yep, agreed...will probably do as a follow up.

packages/orm/src/EntityManager.ts
stephenh: Yep, will fix.

PR #25, opened 2020-06-26
Add hasOneThrough

stephenh:


packages/integration-tests/src/collections/CustomReference.test.ts
```diff
@@ -1,66 +1,74 @@
 import { EntityManager } from "joist-orm";
-import { knex, numberOfQueries, resetQueryCount } from "../setupDbTests";
-import { Author, Book, BookReview } from "../entities";
-import { insertAuthor, insertBook, insertBookReview, insertPublisher } from "../entities/factories";
+import { Author, BookReview, Image, ImageType } from "../entities";
+import { insertAuthor, insertBook, insertBookReview, insertImage } from "../entities/factories";
+import { knex } from "../setupDbTests";
 
 describe("CustomReference", () => {
```
stephenh, line 6: I moved this one to use the new `Image` b/c `BookReview` is using `hasOneThrough` now, so the `hasOneThrough.test.ts` covers those code paths, and the code paths these tests were for are now in the `Image.owner`.

packages/integration-tests/src/entities/BookReview.ts
```diff
@@ -1,24 +1,12 @@
...
-      review.book.get.author.set(author);
-    },
-    isSet: (review) => {
-      return review.book.isSet() && review.book.get.author.isSet();
-    },
-  });
+  readonly author: Reference<BookReview, Author, never> = hasOneThrough(
+    this as BookReview,
+    { book: "author" } as const,
+    (review) => review.book.get.author.get,
```
stephenh, line 8: I'm pretty sure this could be:

```typescript
hasOneThrough(review => review.book.author);
```

Using the lens functionality, but currently the lens eval logic can only do `await`s but with some copy/paste, it could have a synchronous eval...

stephenh: (Updated to use the lens in the latest commits.)

packages/codegen/src/generateEntityCodegenFile.ts
```diff
@@ -237,6 +239,12 @@ export function generateEntityCodegenFile(config: Config, meta: EntityDbMetadata
         return ${loadLens}(this as any as ${entityName}, fn);
       }
 
+      hasOneThrough<U extends ${EntitySym}, N extends undefined | never, V extends U | N>(
```
zgavin, line 242: Do we really want this added to the api of every entity?  Seems like just using the imported `hasOneThrough` would be sufficient.

packages/integration-tests/src/entities/BookReview.ts
```diff
@@ -1,24 +1,9 @@
...
-    get: (review) => review.book.get.author.get,
-    set: (review, author) => {
-      review.book.get.author.set(author);
-    },
-    isSet: (review) => {
-      return review.book.isSet() && review.book.get.author.isSet();
-    },
-  });
+  // Currently this infers as Reference<BookReview, Author, undefined> --> it should be never...
+  readonly author: Reference<BookReview, Author, never> = this.hasOneThrough((review) => review.book.author);
```
zgavin, line 6: Must have been nice deleting all these lines 😆

packages/orm/src/collections/CustomReference.ts
```diff
@@ -1,25 +1,35 @@
...
+import { AbstractRelationImpl } from "./AbstractRelationImpl";
 
-export type CustomReferenceOpts<
-  T extends Entity,
-  U extends Entity,
-  H extends LoadHint<T>,
-  N extends never | undefined
-> = {
+export type CustomReferenceOpts<T extends Entity, U extends Entity, N extends never | undefined> = {
+  // We don't capture the return value of load b/c we purposefully want get to re-calc from `entity`
```
zgavin, line 6: 👍 comment

packages/orm/src/collections/CustomReference.ts
```diff
@@ -1,25 +1,35 @@
...
   load: (entity: T) => Promise<void>;
-  get: (entity: Loaded<T, H>) => U | N;
-  set?: (entity: Loaded<T, H>, other: U) => void;
-  isSet?: (entity: Loaded<T, H>) => boolean;
+  get: (entity: T) => U | N;
+  set?: (entity: T, other: U) => void;
 };
 
-export class CustomReference<T extends Entity, U extends Entity, H extends LoadHint<T>, N extends never | undefined>
+/**
```
zgavin, line 13: 👍  again

packages/codegen/src/generateEntityCodegenFile.ts
stephenh: Yeah, good question; without this, the caller did:

```
reaodnly author: Reference<...> = hasOneThrough(this as BookReview, ...);
```

And the `as BookReview` was killing me that it can't be inferred correctly.

Adding "well, for some reason `this` inference is weird" methods to the codebase class is fairly common at this point, but I agree it is not great. Would also be nice to have an in-joist-but-not-the-generated-code comment about this rationale.

packages/orm/src/collections/CustomReference.ts
stephenh: Yep. :-) I'd had a version for awhile that was capturing the return value and at the "oh fuck, right..." moment thought I'd tell myself not to try that again, because w/o the comment I'm sure I would try it again in about 3-6 months.

stephenh:
> It seems like a railsism when all our other relationships aren't like that. 

Hm, yeah, I thought the rail-ism might actually be a marketing win...

Granted, you're right insofar has we don't have the `hasOne`s b/c Joist just gives those to you.

PR #26, opened 2020-06-28
Implement one to ones

stephenh:


packages/orm/src/collections/ManyToOneReference.ts
```diff
@@ -1,14 +1,26 @@
...
 /**
  * Manages a foreign key from one entity to another, i.e. `Book.author --> Author`.
  *
  * We keep the current `author` / `author_id` value in the `__orm.data` hash, where the
  * current value could be either the (string) author id from the database, or an entity
  * `Author` that the user has set.
+ *
+ * Note that if our `images.author_id` column is unique, this `ManyToOneReference` will essentially
+ * be half of a one-to-one relationship, but we'll keep using this reference on the "owning"
+ * side; the other side, i.e. `Author.image` will use a `OneToOneReference` to point back to us.
```
zgavin, line 23: Was wondering how you were going to do this.  Is there a technical limitation to making it just o2o on both sides?  IMO that would be preferable.  In practice most people are just going to use `entity.relation.get[0]` anyways.  It'd be nice if we could make it both cleaner for the developer and the type system.  Also then all the code related to a o2o could be in o2o.  Seems like it could just generate a different query for load based on which end of the relationship it's on.

README.markdown
```diff
@@ -61,7 +61,6 @@ In general priority/itch order:
 - LargeCollection support
   - I.e. `joist-config.json` entry to mark (say) `publisher.authors` as "too big to never read at once" and use a different/restricted API, i.e. forced paging
 - JSON support, i.e. `toJSON`
-- Cascading deletions for parent/child relationships
```
zgavin, line 64: 😄

docs/misc-features.markdown
```diff
@@ -344,3 +344,15 @@ The `hasOneThrough` DSL is built on Joist's `CustomReferences`, so will also wor
 const review = await em.load(BookReview, "1", { author: "publisher" });
 expect(review.author.get.publisher.get.name).toEqual("p1");
 ```
+
+### Cascading Deletions
```
zgavin, line 348: Thanks, had forgotten this file even existed

packages/orm/src/collections/CustomReference.ts
```diff
@@ -110,13 +97,23 @@ export class CustomReference<T extends Entity, U extends Entity, N extends never
...
+  }
+
   toString(): string {
     return `CustomReference(entity: ${this.entity}, fieldName: ${this.fieldName})`;
   }
 
-  /** Finds this CustomReferences field name by looking in the entity for the key that we're assigned to. */
-  get fieldName(): string {
-    return Object.entries(this.entity).filter((e) => e[1] === this)[0][0];
+  private doGet(opts?: { withDeleted?: boolean }): U | N {
```
zgavin, line 109: Is there some reason these moved around?  Just preference to have private methods at the bottom of the file?

packages/orm/src/collections/OneToManyCollection.ts
```diff
@@ -9,6 +9,10 @@ import { AbstractRelationImpl } from "./AbstractRelationImpl";
 export class OneToManyCollection<T extends Entity, U extends Entity> extends AbstractRelationImpl<U[]>
   implements Collection<T, U> {
   private loaded: U[] | undefined;
+  // We don't need to track removedBeforeLoaded, because if a child is removed in our unloaded state,
```
zgavin, line 12: It took me 3-4 read throughs to understand this comment.

packages/orm/src/collections/OneToOneReference.ts
```diff
@@ -0,0 +1,126 @@
...
+    this.set(other);
+  }
+
+  initializeForNewEntity(): void {
+    this.isLoaded = true;
+  }
+
+  async refreshIfLoaded(): Promise<void> {}
+
+  onEntityDelete(): void {
```
zgavin, line 104: Remove commented code if it's not necessary?

packages/orm/src/index.ts
```diff
@@ -117,7 +118,7 @@ interface Flavoring<FlavorT> {
 }
 export type Flavor<T, FlavorT> = T & Flavoring<FlavorT>;
 
-export function setField(entity: Entity, fieldName: string, newValue: any): void {
+export function setField(entity: Entity, fieldName: string, newValue: any): boolean {
```
zgavin, line 121: Not a big fan of just returning an opaque boolean from a function that isn't implicitly obvious what it is from the function declaration.  

Not sure it's a huge deal, though, since this is basically private anyways.

packages/orm/src/collections/ManyToOneReference.ts
stephenh: > o2o on both sides

Well, it effectively looks like a o2o on both sides, i.e. both sides are `image.book = ...` and `book.image = ...`. This is just the impl caveat that the side that "owns the fk" still uses the ManyToOneReference class.

> In practice most people are just going to use entity.relation.get[0] anyways

Hm, not following?

> Seems like it could just generate a different query for load based 

Hm, potentially, but I think the differences of "one side is based on the column in the table" and "the other side is always a lookup" are big-ish. I.e. it shows up has the "column in table side" could be nullable/not-nullable (based on not null on the column) but the other is always nullable (b/c we can't enforce at the db level that the pk of the current row must exist on a row in the other table). 

I was kinda surprised at how little had to change in ManyToOne so will probably keep for now.

Does seem like a collection refactoring is coming up at some point.

docs/misc-features.markdown
stephenh: Ha, yeah, it is pretty typically for features to show up hear "two or three releases" after they're in.

packages/orm/src/collections/CustomReference.ts
stephenh: Right, should have done as a separate commit, but public API / methods first, impl details hidden below.

packages/orm/src/collections/OneToOneReference.ts
stephenh: True, yeah...I think this is actually a valid reminder that cascade deletion for o2os is not tested / probably doesn't work. I'll change it to that.

PR #28, opened 2020-06-30
Add hasOneDerived.

stephenh:
Needed this for commitment line items having a commitment that is either the change order or the commitment.

### DIFF
--- a/packages/integration-tests/src/entities/BookReview.ts
+++ b/packages/integration-tests/src/entities/BookReview.ts
@@ -1,9 +1,16 @@
-import { Reference } from "joist-orm";
-import { Author, BookReviewCodegen, bookReviewConfig } from "./entities";
+import { hasOneDerived, hasOneThrough, Reference } from "joist-orm";
+import { Author, BookReviewCodegen, bookReviewConfig, Publisher } from "./entities";
 
 export class BookReview extends BookReviewCodegen {
   // Currently this infers as Reference<BookReview, Author, undefined> --> it should be never...
-  readonly author: Reference<BookReview, Author, never> = this.hasOneThrough((review) => review.book.author);
+  readonly author: Reference<BookReview, Author, never> = hasOneThrough((review) => review.book.author);
+
+  // This is kind of silly domain wise, but used as an example of hasOneDerived with a load hint. We don't
+  // technically have any conditional logic in `get` so could use a lens, but we want to test hasOneDerived.
+  readonly publisher: Reference<BookReview, Publisher, undefined> = hasOneDerived(
+    { book: { author: "publisher" } },
+    (review) => review.book.get.author.get.publisher.get,
+  );
 }
 
 // Reviews are only public if the author is over the age of 21
--- a/packages/integration-tests/src/entities/Image.ts
+++ b/packages/integration-tests/src/entities/Image.ts
@@ -4,6 +4,8 @@ import { Author, Book, ImageCodegen, imageConfig, ImageType, Publisher } from ".
 type ImageOwner = Book | Publisher | Author;
 
 export class Image extends ImageCodegen {
+  // We don't use hasOneThrough or hasOneDerived b/c we use the ImageType to do a
+  // selective .load instead of a load hint that probes every possible table.
   readonly owner: Reference<Image, ImageOwner, undefined> = new CustomReference<Image, ImageOwner, undefined>(this, {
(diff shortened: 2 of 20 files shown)


packages/integration-tests/src/entities/Image.ts
```diff
@@ -4,6 +4,8 @@ import { Author, Book, ImageCodegen, imageConfig, ImageType, Publisher } from ".
 type ImageOwner = Book | Publisher | Author;
 
 export class Image extends ImageCodegen {
+  // We don't use hasOneThrough or hasOneDerived b/c we use the ImageType to do a
+  // selective .load instead of a load hint that probes every possible table.
```
stephenh, line 8: Separately from the comment, I'm wondering if this usage could become:

```
this.hasOneDerived(loadFn, getFn)
```

As an overload to the current:

```
this.hasOneDerived(loadHint, getFn);
```

It should be pretty easy to use TS overloading to support both signatures...

I also haven't added the `this.hasOneDerived` yet. Fwiw I've also tried to add these types of methods to the `BaseEntity` but that also trips up inference, which is weird b/c TS has a `this` type that is specifically supposed to support things like that...

packages/integration-tests/src/entities/BookReview.ts
```diff
@@ -1,9 +1,17 @@
...
+import { hasOneDerived, Reference } from "joist-orm";
+import { Author, BookReviewCodegen, bookReviewConfig, Publisher } from "./entities";
 
 export class BookReview extends BookReviewCodegen {
   // Currently this infers as Reference<BookReview, Author, undefined> --> it should be never...
   readonly author: Reference<BookReview, Author, never> = this.hasOneThrough((review) => review.book.author);
+
+  // This is kind of silly domain wise, but used as an example of hasOneDerived with a load hint. We don't
+  // technically have any conditional logic in `get` so could use a lens, but we want to test hasOneDerived.
+  readonly publisher: Reference<BookReview, Publisher, undefined> = hasOneDerived(
```
zgavin, line 10: can we unify how these are called? `hasOneThrough` is an instance method (still not a fan of this) and `hasOneDerived` is just an imported function.

stephenh:
> Not entirely sure what this gives us over just CustomReference other than a load hint instead of a load function

My primary goal is to have it be a function call b/c that seems to make type inference work better.

And, yeah, passing a load hint so that `load` can be dropped.

In general I like keeping `CustomReference` pretty generic and having wrapper methods that have cuter ways of calling it.

packages/integration-tests/src/entities/BookReview.ts
stephenh: Right, hadn't gotten to making it an instance method...

zgavin:
Adding a `loadHint` to `CustomReferenceOpts` that's exclusive with `load` seems like it'd still satisfy your (imo correct) desire for `CustomReference` to stay pretty generic.  eg, something like:

```typescript
export type CustomReferenceOpts<
  T extends Entity,
  U extends Entity,
  H extends LoadHint<T>,
  N extends never | undefined
> = (
  | { loadHint: LoadHint<T>; load?: undefined }
  | {
      loadHint?: undefined;
      load: (entity: T) => Promise<void>;
    }
) & {
  get: (entity: Loaded<T, H>) => U | N;
  set?: (entity: Loaded<T, H>, other: U) => void;
  isSet?: (entity: Loaded<T, H>) => boolean;
};
```

`hasOneDerived` just feels like added bloat to me.  Instead of thinking 'whats the argument to use a load hint instead of a loader on a CustomReference' it's 'whats that random function i need to call to use a load hint'.

zgavin:
Cleaner version of that type definition: 

```typescript
type Without<T, U> = { [P in Exclude<keyof T, keyof U>]?: never };
type Exclusive<T, U> = T | U extends object ? (Without<T, U> & U) | (Without<U, T> & T) : T | U;
export type CustomReferenceOpts<
  T extends Entity,
  U extends Entity,
  H extends LoadHint<T>,
  N extends never | undefined
> = Exclusive<{ loadHint: LoadHint<T> }, { load: (entity: T) => Promise<void> }> & {
  get: (entity: Loaded<T, H>) => U | N;
  set?: (entity: Loaded<T, H>, other: U) => void;
  isSet?: (entity: Loaded<T, H>) => boolean;
};
```

PR #32, opened 2020-07-07
Added CustomCollection

zgavin:
Added `CustomCollection` for collections with complex logic. 
Also added `hasManyThrough` and `hasManyDerived` helper functions to setup CustomCollections.

### DIFF
--- a/packages/codegen/src/EntityDbMetadata.test.ts
+++ b/packages/codegen/src/EntityDbMetadata.test.ts
@@ -1,6 +1,7 @@
 import { collectionName, makeEntity } from "./EntityDbMetadata";
 
 const relationDummy: any = { targetTable: { m2oRelations: [] } };
+
 describe("EntityDbMetadata", () => {
   describe("collectionName", () => {
     it("handles base case", () => {
--- a/packages/codegen/src/EntityDbMetadata.test.ts
+++ b/packages/codegen/src/EntityDbMetadata.test.ts
@@ -1,19 +1,22 @@
 import { collectionName, makeEntity } from "./EntityDbMetadata";
 
 const relationDummy: any = { targetTable: { m2oRelations: [] } };
+const configDummy: any = { relationNameOverrides: {} };
 
 describe("EntityDbMetadata", () => {
   describe("collectionName", () => {
     it("handles base case", () => {
-      expect(collectionName(makeEntity("Author"), makeEntity("Book"), relationDummy)).toEqual("books");
+      expect(collectionName(configDummy, makeEntity("Author"), makeEntity("Book"), relationDummy)).toEqual("books");
     });
 
     it("handles author/mentor", () => {
-      expect(collectionName(makeEntity("Author"), makeEntity("Author"), relationDummy)).toEqual("authors");
+      expect(collectionName(configDummy, makeEntity("Author"), makeEntity("Author"), relationDummy)).toEqual("authors");
     });
 
     it("handles book/book review", () => {
-      expect(collectionName(makeEntity("Book"), makeEntity("BookReview"), relationDummy)).toEqual("reviews");
+      expect(collectionName(configDummy, makeEntity("Book"), makeEntity("BookReview"), relationDummy)).toEqual(
+        "reviews",
+      );
     });
   });
 });
--- a/packages/codegen/src/EntityDbMetadata.test.ts
+++ b/packages/codegen/src/EntityDbMetadata.test.ts
@@ -1,17 +1,18 @@
 import { collectionName, makeEntity } from "./EntityDbMetadata";
 
+const relationDummy: any = { targetTable: { m2oRelations: [] } };
 describe("EntityDbMetadata", () => {
   describe("collectionName", () => {
     it("handles base case", () => {
-      expect(collectionName(makeEntity("Author"), makeEntity("Book"))).toEqual("books");
+      expect(collectionName(makeEntity("Author"), makeEntity("Book"), relationDummy)).toEqual("books");
     });
 
     it("handles author/mentor", () => {
-      expect(collectionName(makeEntity("Author"), makeEntity("Author"))).toEqual("authors");
+      expect(collectionName(makeEntity("Author"), makeEntity("Author"), relationDummy)).toEqual("authors");
     });
 
     it("handles book/book review", () => {
-      expect(collectionName(makeEntity("Book"), makeEntity("BookReview"))).toEqual("reviews");
+      expect(collectionName(makeEntity("Book"), makeEntity("BookReview"), relationDummy)).toEqual("reviews");
     });
   });
 });
--- a/packages/codegen/src/EntityDbMetadata.ts
+++ b/packages/codegen/src/EntityDbMetadata.ts
@@ -3,7 +3,7 @@ import { Column, M2MRelation, M2ORelation, O2MRelation, Table } from "pg-structu
 import pluralize from "pluralize";
 import { imp } from "ts-poet";
 import { SymbolSpec } from "ts-poet/build/SymbolSpecs";
-import { Config, isAsyncDerived, isDerived, isProtected, ormMaintainedFields } from "./config";
+import { Config, isAsyncDerived, isDerived, isProtected, ormMaintainedFields, relationName } from "./config";
 import { ColumnMetaData } from "./generateEntityCodegenFile";
 import { isEnumTable, isJoinTable, mapSimpleDbType, tableToEntityName } from "./utils";
 
@@ -104,19 +104,19 @@ export class EntityDbMetadata {
     this.manyToOnes = table.m2oRelations
       .filter((r) => !isEnumTable(r.targetTable))
       .filter((r) => !isMultiColumnForeignKey(r))
-      .map((r) => newManyToOneField(this.entity, r));
+      .map((r) => newManyToOneField(config, this.entity, r));
     this.oneToManys = table.o2mRelations
       // ManyToMany join tables also show up as OneToMany tables in pg-structure
       .filter((r) => !isJoinTable(r.targetTable))
       .filter((r) => !isMultiColumnForeignKey(r))
       .filter((r) => !isOneToOneRelation(r))
-      .map((r) => newOneToMany(this.entity, r));
+      .map((r) => newOneToMany(config, this.entity, r));
     this.oneToOnes = table.o2mRelations
       // ManyToMany join tables also show up as OneToMany tables in pg-structure
       .filter((r) => !isJoinTable(r.targetTable))
       .filter((r) => !isMultiColumnForeignKey(r))
       .filter((r) => isOneToOneRelation(r))
-      .map((r) => newOneToOne(this.entity, r));
+      .map((r) => newOneToOne(config, this.entity, r));
     this.manyToManys = table.m2mRelations
       // pg-structure is really loose on what it considers a m2m relationship, i.e. any entity
       // that has a foreign key to us, and a foreign key to something else, is automatically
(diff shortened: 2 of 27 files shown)


packages/integration-tests/src/getProperties.test.ts
```diff
@@ -3,42 +3,28 @@ import { getProperties } from "joist-orm";
...
-    expect(getProperties(bookMeta)).toMatchInlineSnapshot(`
-      Array [
-        "advances",
-        "reviews",
-        "author",
-        "image",
-        "tags",
-      ]
-    `);
+    expect(getProperties(bookMeta)).toEqual(expect.arrayContaining(["advances", "reviews", "author", "image", "tags"]));
```
stephenh, line 6: Do we need `arrayContaining`? 

Fwiw I slightly prefer the inline snapshots b/c they are so ridiculously easy to update.

I.e. when doing the refactoring for tagged ids in graphql-service, the tests that had ids in inline snapshots were basically insta-fixed + just post-review the diff to verify they make sense, vs. the ones that had ids in random strings in `.toEqual`'s that I had to hunt and peck update. I probably spent ~2-3 hours updating tests that could have been could have been insta-fixed (if in the extreme scenario which I'm not seriously pitching) if they used snapshots.

So not saying I'd use `toMatchInlineSnapshot` for every assertion ever (...thinking about it...), but just communicating that I specifically like how "agile" they are/pleasant they make updating test assertions while refactoring.

Just musing, but if Jest had a way to auto-update failures to `.toEqual` calls that were constants, that would be pretty slick.

packages/orm/src/collections/hasManyDerived.ts
```diff
@@ -0,0 +1,32 @@
...
+export function hasManyDerived<T extends Entity, U extends Entity, H extends LoadHint<T>>(
+  loadHint: H,
+  opts: HasManyDerivedOpts<T, U, H>,
+): Collection<T, U> {
+  const entity: T = currentlyInstantiatingEntity as T;
+  const { load, get, set, add, remove } = opts;
+  return new CustomCollection<T, U>(entity, {
+    load: load ?? (async (entity) => await getEm(entity).populate(entity, loadHint)),
+    get: () => get(entity as Loaded<T, H>),
+    set: set !== undefined ? (_, values) => set(entity as Loaded<T, H>, values) : undefined,
```
stephenh, line 28: Maybe `fail` the set/add/remove if `undefined`?

packages/orm/src/loadLens.ts
```diff
@@ -75,7 +75,7 @@ function maybeLoad(object: any, path: string): unknown {
  *
  * This assumes you've first evaluated the lens with `loadLens` and now can access it synchronously.
  */
-export function getLens<T, U, V>(start: T, fn: (lens: Lens<T>) => Lens<U, V>): V {
+export function getLens<T, U, V>(start: T, fn: (lens: Lens<T>) => Lens<U, V>): V | V[] {
```
stephenh, line 78: Hm...I think that if the lens "resolved" to a collection, then `V` would already be an array...can this be removed?

packages/orm/src/collections/hasManyThrough.ts
```diff
@@ -0,0 +1,20 @@
...
+ */
+export function hasManyThrough<T extends Entity, U extends Entity>(
+  lens: (lens: Lens<T>) => Lens<U, U[]>,
+): Collection<T, U> {
+  const entity: T = currentlyInstantiatingEntity as T;
+  return new CustomCollection<T, U>(entity, {
+    load: async (entity) => {
+      await loadLens(entity, lens);
+    },
+    get: () => getLens(entity, lens) as U[],
```
stephenh, line 18: I kinda expected this return type of `getLens` to already be `U[]` given the `Lens<U, U[]>` type up above...

packages/orm/src/collections/hasManyDerived.ts
zgavin: it will fail in `CustomCollection`

packages/integration-tests/src/EntityManager.test.ts
```diff
@@ -614,11 +614,9 @@ describe("EntityManager", () => {
     // Then we issue a single SQL query
     expect(numberOfQueries).toEqual(1);
     // And it's the regular/sane query, i.e. not auto-batched
-    expect(queries).toMatchInlineSnapshot(`
-      Array [
-        "select \\"p0\\".* from \\"publishers\\" as \\"p0\\" where \\"p0\\".\\"id\\" = ? order by \\"p0\\".\\"id\\" asc",
-      ]
-    `);
+    expect(queries).toMatchObject([
```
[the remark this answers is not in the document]
zgavin: Yep.

packages/integration-tests/src/getProperties.test.ts
zgavin: For this specific line, no we don't need it.  Just `toEqual` is sufficient.  For the other 5 assertions in the file, we need it since we are testing for specific things and shouldn't need to be exhaustive.

I don't really get the purpose of `toMatchInlineSnapshot`.  It seems like added boilerplate without any syntax checking.  All its use cases are handled by a combination of `toEqual`, `toMatchObject` and `expect.arrayContaining`.

packages/orm/src/loadLens.ts
zgavin: Leftover from an earlier iteration, will remove.

packages/orm/src/collections/hasManyThrough.ts
zgavin: ditto last comment

packages/codegen/src/EntityDbMetadata.test.ts
```diff
@@ -1,17 +1,18 @@
 import { collectionName, makeEntity } from "./EntityDbMetadata";
 
+const relationDummy: any = { targetTable: { m2oRelations: [] } };
```
stephenh, line 3: Newline after this?

packages/integration-tests/src/entities/Publisher.ts
```diff
@@ -1,6 +1,41 @@
-import { PublisherCodegen, publisherConfig } from "./entities";
+import { Image, ImageType, PublisherCodegen, publisherConfig } from "./entities";
+import { Collection, CustomCollection, getEm, Loaded } from "joist-orm";
 
-export class Publisher extends PublisherCodegen {}
+const allImagesHint = { images: [], authors: { image: [], books: "image" } } as const;
```
stephenh, line 4: Newline before the class.

packages/integration-tests/src/getProperties.test.ts
stephenh: > shouldn't need to be exhaustive

Ah sure, I didn't mind it for these, although snapshots asserting unrelated things is generally undesirable, especially when you get into external snapshots like `toMatchSnapshot`.

> I don't really get the purpose of toMatchInlineSnapshot.

The purpose is how damn fast it is b/c you don't type anything. You know I strongly believe in static typing, but it's really common for assertions to be "does this `result` DTO/POJO match-ish what I expect" is the singular intent and having to "encode" that intent with ".toEqual ... type a bunch of syntax chars" or "arrayContaining ... type a bunch of syntax chars" doesn't actually add any value to achieving the singular intent of "does this DTO/tree match-ish what I expect". It's just extra typing to placate "how we've always done things".

(I say this as still a very strong `toMatchSnapshot` skeptic and an initial `toMatchInlineSnapshot` skeptic, but I've since come around on the inline version. You get used to "sure it's a string with some extra backets" and weigh the "small con" of "sure this is not syntax highlighted" with the imo "large pro" of "wow I didn't have to type any of this and can now insta-fix assertions for updated tests" and, eh, I've personally found it useful. At least to the extent that I think wholesale removing it from codebases in unrelated PRs is unnecessary.)

(I also very rarely use "less keystrokes" as an argument given vim bindings/"you're a programmer so get good at typing"/etc. But despite "yes I'm already fast" I find judicious `toMatchineInlineSnapshot` makes me (a tiny bit) faster.)

(Also, I'll repeat the idea that if you want to fork Jest and build a `.toEqual`s that will similarly auto-update assertions that are all consts of hashes/lists, and so achieve the same workflow productivity without the "yuck this is a string", I personally think that'd be pretty cool.)

packages/orm/src/collections/hasManyDerived.ts
stephenh: Ah sure, this is just passing along the props.

packages/codegen/src/EntityDbMetadata.ts
```diff
@@ -216,11 +216,16 @@ function newOneToOne(entity: Entity, r: O2MRelation): OneToOneField {
...
 
 /** Returns the collection name to use on `entity` when referring to `otherEntity`s. */
-export function collectionName(entity: Entity, otherEntity: Entity): string {
-  // TODO Handle multiple fks from otherEntity --> entity
+export function collectionName(entity: Entity, otherEntity: Entity, r: M2ORelation | O2MRelation): string {
   // TODO Handle conflicts in names
   // I.e. if the other side is `child.project_id`, use `children`.
   let fieldName = otherEntity.name;
+  // check if we multiple FKs from otherEntity --> entity and prefix with FK name if so
+  const sourceTable = r instanceof M2ORelation ? r.sourceTable : r.targetTable;
```
stephenh, line 224: I love everything about pg-structure except for their sourceTable/targetTable naming. I swear its backwards. And maybe only backwards some of the time, which is even worse than being backwards all of the time. I'm honestly not sure which is which.

packages/integration-tests/src/getProperties.test.ts
zgavin: You don't need to type anything with `toEqual` either. You can get the actual result as a json string, at least you can in webstorm.

For `toMatchObject` or `arrayContains` you're only looking for a subset of the result anyways and imo that should always be explicitly defined so that intent can be understood.

stephenh: >  You can get the actual result as a json string, at least you can in webstorm and then it's just a copy/paste.

?

stephenh: > you're only looking for a subset of the result anyways

Definitely agreed as a rule, with the disclaimer I think it's like DRY, where 100% adherence is not required / judicious exceptions is fine.

A subset version (i.e. `toMatchObject`-ish semantics) of `toMatchInlineSnapshot` would be great. :-P :-) (Agreed `arrayContains` is too specific to try and auto-update on first-assertion/updating-assertions.)

zgavin: If a `toEquals` test fails for me, Webstorm has a '<click to see difference>' link that shows you what the actual result was as json string.  If you really want to be as quick as possible, you can just copy it, then paste it over the previous expectation.  

IMO, some friction when updating tests is actually a good thing.  If it's too easy then it encourages not actually validating the results.

packages/orm/src/collections/ManyToOneReference.ts
```diff
@@ -136,11 +136,14 @@ export class ManyToOneReference<T extends Entity, U extends Entity, N extends ne
 
   // Internal method used by OneToManyCollection
   setImpl(other: U | N): void {
-    if (this.isLoaded && other === this.loaded) {
+    if (other?.isNewEntity ? other === this.loaded : this.id === other?.id) {
       return;
     }
 
-    const previousLoaded = this.loaded;
+    // we may not be loaded yet, but our previous entity might already be in the UoW
```
stephenh, line 143: Can you add a dedicated test for this to `ManyToOneReference.test.ts`? I could see this breaking / be worth testing explicitly while working on MTOR and not realizing it's only covered implicitly by other tests.

PR #36, opened 2020-07-10
Update partial methods to upsert collections.

stephenh:
We also look for 'delete' and 'remove' markers to either delete or
remove the method from the collection.

Right now 'delete' == `EntityManager.delete` i.e. not soft-deleting.

### DIFF
--- a/packages/orm/src/createOrUpdatePartial.ts
+++ b/packages/orm/src/createOrUpdatePartial.ts
@@ -1,15 +1,5 @@
 import { NullOrDefinedOr } from "./utils";
-import {
-  Entity,
-  EntityConstructor,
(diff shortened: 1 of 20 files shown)


packages/orm/src/createOrUpdatePartial.ts
```diff
@@ -74,6 +71,9 @@ export async function createOrUpdatePartial<T extends Entity, O>(
               return value;
             } else if (isKey(value)) {
               return await em.load(field.otherMetadata().cstr, value);
+            } else if ("id" in value && ("delete" in value || "remove" in value)) {
```
stephenh, line 74: This probably needs some work...

PR #38, opened 2020-07-14
Optimize findExistingInstance and getMetadata.

stephenh:
Profiling a query that loaded ~2,000 entities showed large hot-spots
in `findExistingInstance` + `isEntity` + `getMetadata`. It was taking
~60 seconds and eventually timing out.

Adding an index for `findExistingInstance` dropped the time for the
same query to ~3 seconds and then it fails with a "too many entities"
error.

Kinda surprised at how much of a difference this makes, but when loading
~1000s of entities, findExistingInstance is called in a tight loop and
turns out O(N^2) is not a great idea.

### DIFF
--- a/packages/codegen/src/generateEntityCodegenFile.ts
+++ b/packages/codegen/src/generateEntityCodegenFile.ts
@@ -227,7 +227,7 @@ export function generateEntityCodegenFile(config: Config, meta: EntityDbMetadata
       ${[o2m, m2o, o2o, m2m]}
 
       constructor(em: ${EntityManager}, opts: ${entityName}Opts) {
-        ${hasDefaultValues ? code`super(em, ${metadata}, {...${defaultValuesName}})` : code`super(em, ${metadata})`};
+        super(em, ${metadata}, ${hasDefaultValues ? `${defaultValuesName}` : "{}"}, opts);
         ${setOpts}(this as any as ${entityName}, opts, { calledFromConstructor: true });
       }
 
--- a/packages/orm/src/BaseEntity.ts
+++ b/packages/orm/src/BaseEntity.ts
@@ -21,9 +21,13 @@ export abstract class BaseEntity implements Entity {
   abstract id: string | undefined;
   readonly __orm: EntityOrmField;
 
-  protected constructor(em: EntityManager, metadata: any, data?: Record<string, any>) {
-    this.__orm = { em, metadata, data: data || {}, originalData: {} };
-    em.register(this);
+  protected constructor(em: EntityManager, metadata: any, defaultValues: object, opts: any) {
+    this.__orm = { em, metadata, data: { ...defaultValues }, originalData: {} };
+    // Ensure we have at least id set so the `EntityManager.register` works
+    if (typeof opts === "string") {
+      this.__orm.data["id"] = opts;
+    }
+    em.register(metadata, this);
   }
 
   get idUntagged(): string | undefined {
--- a/packages/orm/src/EntityManager.ts
+++ b/packages/orm/src/EntityManager.ts
@@ -189,6 +189,9 @@ export class EntityManager {
   constructor(public knex: Knex) {}
 
   private _entities: Entity[] = [];
+  // Indexes the currently loaded entities by their tagged ids. This fixes a real-world
+  // performance issue where `findExistingInstance` scanning `_entities` was an `O(n^2)`.
+  private _entityIndex: Map<string, Entity> = new Map();
   private findLoaders: LoaderCache = {};
   private flushSecret: number = 0;
   private _isFlushing: boolean = false;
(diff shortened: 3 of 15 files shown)


packages/orm/src/EntityManager.ts
```diff
@@ -520,17 +521,21 @@ export class EntityManager {
   }
 
   /** Registers a newly-instantiated entity with our EntityManager; only called by entity constructors. */
-  register(entity: Entity): void {
-    if (entity.id && this.findExistingInstance(getMetadata(entity).cstr, entity.id) !== undefined) {
+  register(meta: EntityMetadata<any>, entity: Entity): void {
```
stephenh, line 524: The caller already had `meta` around so passing it through to avoid a `getMetadata` call.

packages/codegen/src/generateEntityCodegenFile.ts
```diff
@@ -227,7 +227,7 @@ export function generateEntityCodegenFile(config: Config, meta: EntityDbMetadata
       ${[o2m, m2o, o2o, m2m]}
 
       constructor(em: ${EntityManager}, opts: ${entityName}Opts) {
-        ${hasDefaultValues ? code`super(em, ${metadata}, {...${defaultValuesName}})` : code`super(em, ${metadata})`};
+        super(em, ${metadata}, ${hasDefaultValues ? `${defaultValuesName}` : "{}"}, opts);
```
stephenh, line 230: I was originally making some real changes here that I had to back out (thought I could call setOpts from the BaseEntity cstr, which I try like once every 3 months and forget that it can't b/c the subclass fields need to be instantiated first), but kept this minor change.

packages/orm/src/EntityManager.ts
```diff
@@ -986,7 +1009,9 @@ export function sameEntity(a: Entity, bMeta: EntityMetadata<any>, bCurrent: Enti
 export function getMetadata<T extends Entity>(entity: T): EntityMetadata<T>;
 export function getMetadata<T extends Entity>(type: EntityConstructor<T>): EntityMetadata<T>;
 export function getMetadata<T extends Entity>(entityOrType: T | EntityConstructor<T>): EntityMetadata<T> {
-  return (isEntity(entityOrType) ? entityOrType.__orm.metadata : (entityOrType as any).metadata) as EntityMetadata<T>;
+  return (typeof entityOrType === "function"
+    ? (entityOrType as any).metadata
+    : entityOrType.__orm.metadata) as EntityMetadata<T>;
```
stephenh, line 1014: Not sure it really matters vs. the `O(N^2)` fix, but flipped it to check "is this entity constructor" first b/c it's a much simpler `typeof` check vs. the ~2-3 probes that `isEntity` does.

packages/orm/src/BaseEntity.ts
```diff
@@ -21,9 +21,13 @@ export abstract class BaseEntity implements Entity {
   abstract id: string | undefined;
   readonly __orm: EntityOrmField;
 
-  protected constructor(em: EntityManager, metadata: any, data?: Record<string, any>) {
-    this.__orm = { em, metadata, data: data || {}, originalData: {} };
-    em.register(this);
+  protected constructor(em: EntityManager, metadata: any, defaultValues: object, opts: any) {
+    this.__orm = { em, metadata, data: { ...defaultValues }, originalData: {} };
+    // Ensure we have at least id set so the `EntityManager.register` works
+    if (typeof opts === "string") {
```
zgavin, line 27: when is opts a string?

packages/orm/src/EntityManager.ts
```diff
@@ -188,6 +188,7 @@ export type LoaderCache = Record<string, DataLoader<any, any>>;
 export class EntityManager {
   constructor(public knex: Knex) {}
 
+  private _entityIndex: Map<Function, Map<string, Entity>> = new Map();
```
zgavin, line 191: does this need to be nested?  could we just use a single level map with tagged ids?

packages/orm/src/BaseEntity.ts
zgavin: oh i see, with the new code path via hydrate.   Not sure I'm a fan of hidden behavior based on the type of an arg, although i suppose it's not publicly exposed since we don't intend for constructors to be directly called or overridden.

stephenh: Yeah, this went from "I'm just hacking this to see if it works" to "...oh huh it does..." to "...well, it's a private api so I guess...".

Agreed it should be better/more explicit eventually, but going to punt for now.

packages/orm/src/EntityManager.ts
stephenh: Oh yeah, good call. Tagged ids ftw.

PR #39, opened 2020-07-14
Strip empty strings to undefined.

stephenh:


packages/orm/src/index.ts
```diff
@@ -193,10 +192,7 @@ export function setOpts<T extends Entity>(
       return;
     }
     // We let optional opts fields be `| null` for convenience, and convert to undefined.
-    const value = _value === null ? undefined : _value;
-    if (value === undefined && requiredKeys.includes(key)) {
-      throw new Error(`${key} is required`);
```
stephenh, line 198: I think this has come up before, but not early-failing now so that they should up as validation errors...

PR #40, opened 2020-07-16
Added support for extra fields on enums

zgavin:


packages/integration-tests/src/entities/ImageType.ts
```diff
@@ -4,32 +4,32 @@ export enum ImageType {
   PublisherImage = "PUBLISHER_IMAGE",
 }
 
-type Details = { id: number; code: ImageType; name: string };
```
stephenh, line 7: Nice, I like the `ImageTypeDetails` name.

packages/codegen/src/generateEnumFile.ts
```diff
@@ -2,45 +2,58 @@ import { Table } from "pg-structure";
...
-    const details: Record<${enumName}, Details> = {
+    const details: Record<${enumName}, ${detailsName}> = {
       ${rows
-        .map(row => {
+        .map((row) => {
           const code = pascalCase(row.code);
           const safeName = row.name.replace(/(["'])/g, "\\$1");
-          return `[${enumName}.${code}]: { id: ${row.id}, code: ${enumName}.${code}, name: '${safeName}' }`;
+          const extras = extraPrimitives
+            .map((p) => `${p.fieldName}: ${JSON.stringify((row as any)[p.columnName])}`)
```
stephenh, line 32: Huh, cute using JSON.stringify to conditionally format/wrap the output.

packages/integration-tests/src/entities/Publisher.ts
```diff
@@ -8,12 +8,18 @@ export class Publisher extends PublisherCodegen {
...
+        .reduce(
+          (images, author) => {
+            images.push(author.image.get!);
+            author.books.get.forEach((book) => images.push(book.image.get!));
+            return images;
+          },
+          [...loaded.images.get],
+        )
+        .filter((imageOrUndefined) => imageOrUndefined !== undefined)
+        .sort((a, b) => ImageTypes.findByCode(a.type)!.sortOrder - ImageTypes.findByCode(b.type)!.sortOrder);
```
stephenh, line 22: I die a little bit each time JS doesn't have `sortBy`.

zgavin: also that the cannonical way to avoid sorting in place is `[...someArray].sort(...)`

zgavin: also that for some reason reduce's arguments are backwards

PR #43, opened 2020-07-20
fix newTestInstance non-required primitive fields

alexkuang:


stephenh:
Huh. I don't know why circleci is not running the build...

alexkuang:
...  Oh, ha.  I didn't seem to have perms to push a branch, so I ended up forking the repo.  I guess I had to set up circle/etc separately on my fork for it to start running.  Which kind of makes sense, but you'd think they'd have better handling for this error case...

stephenh:
@alexkuang ah sure, agreed I expected it to be better handled than that. Added you as a collaborator! :-) 

Once you merge, the release should be pushed out automatically.

PR #51, opened 2020-08-13
Use AsyncLocalStorage instead of contexty.

stephenh:
Running the test suite locally with/without the change, AFAICT the
heap usage/run time is all the same.

---

Before merging, I'm going to locally run the same "1.8mb json response" test that I did with datadog. If this shows the same perf degradation of ~3s --> ~7s, then I'll not merge.

...

Looks like it's still good:

image


packages/orm/src/EntityManager.ts
```diff
@@ -618,39 +612,40 @@ export class EntityManager<C extends HasKnex = HasKnex> {
...
-        recalcDerivedFields(todos);
-        await recalcAsyncDerivedFields(this, todos);
-        await validate(todos);
-        await afterValidation(this.ctx, todos);
-
-        entitiesToFlush.push(...pendingEntities);
-        pendingEntities = this.entities.filter((e) => e.isPendingFlush && !entitiesToFlush.includes(e));
-        this.flushSecret += 1;
+        await new Promise((resolve, reject) => {
+          currentFlushSecret.run({ flushSecret: this.flushSecret }, async () => {
```
stephenh, line 621: They don't technically support this `run` callback being async/a promise, so we use line 620 to promsify this callback.

PR #55, opened 2020-08-19
Bump to v1.

stephenh:


packages/codegen/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "joist-codegen",
-  "version": "0.1.0-bump",
+  "version": "1.0.0-bump",
```
stephenh, line 3: This would end up being `1.0.537`. As we want to bump majors/minors, we'd change the `major`/`minor` vars in `bump.sh`.

floodfx:
1️⃣ 🎉 🍣

PR #56, opened 2020-08-24
Fix 'getting-started' documentation link in README

blimmer:
This link was 404-ing because the file extension was not specified.


stephenh:
Thanks!

PR #59, opened 2020-09-01
Update isSet and deletion for one-to-ones

lucyconklin:


packages/orm/src/collections/OneToOneReference.ts
```diff
@@ -57,8 +57,10 @@ export class OneToOneReference<T extends Entity, U extends Entity> extends Abstr
   }
 
   get isSet(): boolean {
-    // This will failure if we're not loaded yet
-    return this.id !== undefined;
+    if (this.isLoaded) {
+      return this.loaded !== undefined;
+    }
+    throw new Error(`${this.entity}.${this.fieldName} was not loaded`);
```
chr1sjf0x, line 63: question: Why isn't this simply `return this.current() !== undefined`?

I don't think this method is intended to throw an error if it's not set/loaded?

zgavin: This seems close to correct to me.  I think it's just missing a check for a new entity in the if.   A o2o can't know if it's set or not until it's loaded since the FK is on the other entity.  `CustomReference` has the same behavior.

stephenh: @chr1sjf0x there isn't a `current` in this class yet; this is basically an inlined `this.current but error if not loaded`.

> I don't think this method is intended to throw an error if it's not set/loaded?

I think it has to, b/c return `false` in the "not loaded" case would be potentially lying to the caller.

packages/orm/src/collections/OneToOneReference.ts
```diff
@@ -131,12 +133,23 @@ export class OneToOneReference<T extends Entity, U extends Entity> extends Abstr
...
+    }
+  }
+
+  current(opts?: { withDeleted: boolean; }): U | undefined {
+    return this.filterDeleted(this.loaded, opts);
   }
 
-  async onEntityDeletedAndFlushing(): Promise<void> {}
+  async onEntityDeletedAndFlushing(): Promise<void> {
+    if (this.isCascadeDelete) {
```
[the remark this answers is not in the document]
stephenh: @zgavin yeah, I was VC'ing with Lucy to handle this case / do this approach, and nudged her towards this.

Understood what you describe is historically what `onEntityDeletedAndFlushign` is for, but seems like, due to the potentially-un-loaded-ness of a o2o, we need an async-capable place to do the cascade deleting.

What is your proposed alternative? 

...looking at `OneToMany`, I see it's `onEntityDelete` "works" when it's unloaded, but only cascades to `addedBeforeLoaded`. Do you remember if it's a known issue that this wouldn't cascade delete to children that are in the not-yet-loaded-`loaded` array, if it were to be loaded?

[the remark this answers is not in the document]
zgavin: `onEntityDelete` is actually called twice in the life-cycle of a delete operation.   First it's called on the initial `em.delete` call.  This is intended to be synchronous, since `em.delete` is also synchronous.  This is fine, though, because the only objective of this call is to cascade the delete into objects already in the UoW.   The second call is in a `beforeDelete` hook that has itself as a populate hint.  Again, it's fine for this to be sync since the relation should now be loaded. So, there shouldn't really be an issue with the potentially unloadedness of a o2o or any need for an async capable cascade delete code path.

The only caveat here is that o2o might want to search the UoW for an FK pointing to it when it isn't yet loaded to be absolutely sure that it's not already in memory.

packages/orm/src/collections/OneToOneReference.ts
chr1sjf0x: Ah, thanks for the explanation @stephenh. I had not considered that the fk is on the `other` side. 👍

lucyconklin:
Thanks for all the input! Tests have been added, and a bit of cleanup.

packages/orm/src/collections/OneToOneReference.ts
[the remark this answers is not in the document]
stephenh: Ah wow, I'd completely missed that `onEntityDelete` would be called twice...

Lucy, per ^ you should be able to delete `onEntityDeletedAndFlushing` and still have your tests pass.

If you don't mind, would be great to also distill Zach's point ^ into a comment on `onDeleteEntity`, i.e.:

```
// Note this will be called twice, once on em.delete (may not be loaded), and again from
// the `beforeFlush` hook that `cascadeDelete` adds (and will be loaded).
onEntityDelete(): void {
   ...
}
```

packages/integration-tests/src/entities/Author.ts
```diff
@@ -65,6 +65,7 @@ export class Author extends AuthorCodegen {
 }
 
 authorConfig.cascadeDelete("books");
+authorConfig.cascadeDelete("image");
```
stephenh, line 68: Ah, great that our made-up-schema already had a "free to be used for this use case" relation to use. :-)

packages/orm/src/collections/OneToOneReference.ts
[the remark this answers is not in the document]
stephenh: > If i'm understanding your last paragraph, then yes it's a known issue. 

Cool ... wonder if we should how persist the cascade-deleted-ness of `someEntity.someCascadeCollection` and then pre-emptively `em.delete` them on that `.load` call. Not a big deal / would be a separate PR from this one.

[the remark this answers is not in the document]
zgavin: > Lucy, per ^ you should be able to delete onEntityDeletedAndFlushing and still have your tests pass.

That is not actually what I was intending.  The tests may pass, but the behavior is certainly not correct.  `onEntityDeletedAndFlushing` still needs to have an implementation because the relation and entities involved still need to be cleaned up.  It just shouldn't be calling `em.delete`.  I pointed to `OneToMany` as an example because it's pretty much needs the same thing, just instead of an array of `loaded` entities, it's just a single entity.

> wonder if we should how persist the cascade-deleted-ness of someEntity.someCascadeCollection and then pre-emptively em.delete them on that .load call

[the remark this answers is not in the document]
zgavin: > wonder if we should how persist the cascade-deleted-ness of someEntity.someCascadeCollection and then pre-emptively em.delete

In practical terms, i'm not sure it's necessary. Anywhere we are calling `em.delete` we are either immediately calling `em.flush` or we've already pre-loaded all our entities/relations.  This could also result in a significant performance regression because it's not really sufficient to only do this check in the relation.  You'd need to check if an entity should be cascade deleted every time you instantiate one or change `isEntityDeleted` cascade aware, which is a problem since it's called a lot and it'd need to search through the UoW.

packages/orm/src/collections/OneToOneReference.ts
```diff
@@ -79,7 +81,9 @@ export class OneToOneReference<T extends Entity, U extends Entity> extends Abstr
   }
 
   set(other: U): void {
-    ensureNotDeleted(this.entity);
+    if (this.entity.__orm.deleted) {
+      return;
```
zgavin, line 85: This should probably be `ensureNotDeleted(this.entity, { ignore: "pending" });` like it is in the other relations.

A good rule of thumb is to treat `__orm` as protected and not access it from outside an entity.  There are a handful of places where we aren't doing this, but we have been slowly removing those over time.

packages/integration-tests/src/collections/OneToOneReference.test.ts
```diff
@@ -108,4 +108,34 @@ describe("OneToOneReference", () => {
     await a1.image.load();
     expect(a1.image.id).toBeUndefined();
   });
+
```
zgavin, line 111: 👍  for tests

zgavin (changes requested):

zgavin:
Definitely getting there.  I think we just need an implementation for `onEntityDeletedAndFlushing` and it'll be good to go.

packages/orm/src/collections/OneToOneReference.ts
lucyconklin: Ah that makes sense.

zgavin (approved):
👍 Thanks @lucyconklin !

PR #61, opened 2020-09-03
Map numeric columns as numbers.

stephenh:


zgavin (approved):
nice, very glad we can know where exactly publishers are now 👍

PR #64, opened 2020-09-09
Issue-62: Add options to associate optional fields or not

### LINKED ISSUE #62: newTestInstance unnecessarily creates associations for non-required fields
If a field is not required, and not specified in the opts, then newTestInstance should not associate it to an existing entity.

If I have:
```
Author {
  book?: Book
}

Then call:
const a1 = newTestInstance(em, Author, { name: "A1", book: { title: "I'm a book by A1" } });
const a2 = newTestInstance(em, Author, { name: "A2" });

// This will fail
expect(a1.book).not.toEqual(a2.book)
```

chr1sjf0x:
Fixes #62

Attempts to follow existing patters from the FactoryOpts.use example

### DIFF
--- a/packages/integration-tests/src/EntityManager.factories.test.ts
+++ b/packages/integration-tests/src/EntityManager.factories.test.ts
@@ -86,6 +86,26 @@ describe("EntityManager.factories", () => {
     expect(b1.author.get.firstName).toEqual("a3");
   });
 
+  it("can create a author and set mentor automatically", async () => {
+    const em = newEntityManager();
+    // Given there is an existing author
+    const a1 = newAuthor(em);
+    // When we create the second author
+    const a2 = newAuthor(em);
+    // Then we expect it to be mentored by the first
+    expect(a2.mentor.get).toBe(a1);
+  });
+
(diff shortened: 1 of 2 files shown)


packages/integration-tests/src/EntityManager.factories.test.ts
```diff
@@ -86,6 +86,26 @@ describe("EntityManager.factories", () => {
...
+    // Then we expect it to be mentored by the first
+    expect(a2.mentor.get).toBe(a1);
+  });
+
+  it("can create a author and configure to not set the mentor", async () => {
+    const em = newEntityManager();
+    // Given there is an existing author
+    newAuthor(em);
+    // When we create the second author
+    const a2 = newAuthor(em, { associateOptionalEntities: { mentor: false }});
```
stephenh, line 104:
```
    const a2 = newAuthor(em, { mentor: null });
```

Also works for disabling the auto-creation and I think specifies the same intent, i.e. "please keep the `mentor` not set".

chr1sjf0x: Ah yes, thanks you! That does work 👍

PR #65, opened 2020-09-12
Fix date noops still being saved to the db.

stephenh:
If business logic changed a date field, and then changed it back,
`setField` is supposed to realize that the value hasn't actually
changed, and so not treat the entity has dirty.

This was done with '===' which works for primitives, but with Dates
we need need to compare `.getTime`s.


packages/orm/src/index.ts
```diff
@@ -398,3 +398,7 @@ export function getEm(entity: Entity): EntityManager {
 export function getRelations(entity: Entity): AbstractRelationImpl<any>[] {
   return Object.values(entity).filter((v) => v instanceof AbstractRelationImpl);
 }
+
+function equal(a: any, b: any): boolean {
+  return a === b || (a instanceof Date && b instanceof Date && a.getTime() == b.getTime());
```
stephenh, line 403: I was thinking of doing something more generic, like checking for a `a.toJSON` / `b.toJSON` method, and then comparing the result of that...

alexkuang: For something more generic, could eventually adapt something like this: https://github.com/facebook/jest/blob/master/packages/expect/src/jasmineUtils.ts#L65

chr1sjf0x (approved):
👍  LGTM

packages/orm/src/index.ts
stephenh: Interesting, cool find. Yeah, not a bad idea. I was hoping that this could stay simpler.

I've come across the "ah we'll call `toJSON` if you have it" "convention" before...ah right, in graphql servers. I.e. if you give apollo-server/graphqljs some random / non-primitive object as a return value, it will probe for the `toJSON` method and call that.

Like the JS equivalent of type classes I guess. :-P :-) I think they use the term protocols for this, i.e. well-known method names. I think the newer protocols use symbols as keys which are unique-based-on-their-declaration and not just their string name, so are collision resistant... 

Granted, it's cheating a little bit to reuse `toJSON` for something like `toSimpleStructureThatCouldBeStructurallyDiffed` or even `toEqual`... but :shrug: ...

In general I'm surprised there is not a `equals` protocol in JS...

PR #66, opened 2020-09-15
Updated pg-structure to homebound fork that fixes index bug

zgavin:


packages/codegen/package.json
```diff
@@ -10,7 +10,7 @@
     "knex": "^0.21.4",
     "is-plain-object": "^3.0.1",
     "pg": "^8.3.0",
-    "pg-structure": "^5.10.13",
+    "@homebound/pg-structure": "^7.2.4",
```
stephenh, line 13: ...I forgot, does npm let us do `7.2.3-1` How to properly semvers forks is nuanced. I'm good with this though.

yarn.lock
```diff
@@ -247,6 +247,21 @@
     exec-sh "^0.3.2"
     minimist "^1.2.0"
 
+"@homebound/pg-structure@^7.2.4":
```
stephenh, line 250: Hm, I don't read `.lock` file diffs very often, but should we have red lines that show the non-homebound pg-structure isn't pulled in anymore?

stephenh (approved):

yarn.lock
zgavin: yeah weird. good catch

PR #70, opened 2020-10-14
Ignore date sets of same value

blimmer:
This relates to #65. It's the same issue, just another place we need to use the new `equal` method vs. `===`.

### DIFF
--- a/packages/integration-tests/src/setupTestEnv.ts
+++ b/packages/integration-tests/src/setupTestEnv.ts
@@ -1,0 +1,3 @@
+export default function globalSetup() {
+  process.env.TZ = "UTC";
+}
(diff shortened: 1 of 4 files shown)


packages/integration-tests/src/setupTestEnv.ts
```diff
@@ -0,0 +1,3 @@
+export default function globalSetup() {
+  process.env.TZ = "UTC";
```
blimmer, line 2: For my new test, I'm setting a date manually in the DB using a string. Then, I'm setting it to the same value via `new Date()`. This normalizes the timezone across developer machines for the test suite.

zgavin (approved):
lgtm 👍

stephenh (approved):
Awesome! Really appreciate pulling the thread on "why are these still happening?".

blimmer:
No problem!

PR #73, opened 2020-10-28
Add loadAllIfExists

blimmer:
This PR adds a new method, `loadAllIfExists`, which works just like `loadAll`, but it does not throw an `Error` if a given id cannot be found in the database.

### DIFF
--- a/packages/integration-tests/src/EntityManager.test.ts
+++ b/packages/integration-tests/src/EntityManager.test.ts
@@ -25,6 +25,32 @@ describe("EntityManager", () => {
     expect(author.firstName).toEqual("f");
   });
 
+  it("can load all entities by id", async () => {
+    await insertAuthor({ first_name: "a1" });
+    await insertAuthor({ first_name: "a2" });
+
+    const em = newEntityManager();
+    const [author1, author2] = await em.loadAll(Author, ["a:1", "a:2"]);
+    expect(author1.firstName).toEqual("a1");
+    expect(author2.firstName).toEqual("a2");
+  });
+
+  it("fails to load all entities by id when any of the ids do not exist", async () => {
+    await insertAuthor({ first_name: "a1" });
+
+    const em = newEntityManager();
+    await expect(em.loadAll(Author, ["a:1", "a:2"])).rejects.toThrowError("a:2 were not found");
+  });
+
+  it("can load all entities by id without throwing an error when any of the ids do not exist", async () => {
+    await insertAuthor({ first_name: "a1" });
+
+    const em = newEntityManager();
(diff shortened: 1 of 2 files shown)


packages/integration-tests/src/EntityManager.test.ts
```diff
@@ -25,6 +25,32 @@ describe("EntityManager", () => {
     expect(author.firstName).toEqual("f");
   });
 
+  it("can load all entities by id", async () => {
```
blimmer, line 28: These first two tests add coverage for `loadAll`'s behavior.

packages/integration-tests/src/EntityManager.test.ts
```diff
@@ -25,6 +25,32 @@ describe("EntityManager", () => {
...
+  });
+
+  it("fails to load all entities by id when any of the ids do not exist", async () => {
+    await insertAuthor({ first_name: "a1" });
+
+    const em = newEntityManager();
+    await expect(em.loadAll(Author, ["a:1", "a:2"])).rejects.toThrowError("a:2 were not found");
+  });
+
+  it("can load all entities by id without throwing an error when any of the ids do not exist", async () => {
```
blimmer, line 45: And this is the new one.

stephenh (approved):
:sushi:

packages/orm/src/EntityManager.ts
```diff
@@ -459,6 +459,32 @@ export class EntityManager<C extends HasKnex = HasKnex> {
...
+    N extends Narrowable
+  >(type: EntityConstructor<T>, ids: string[], populate: H): Promise<Loaded<T, H>[]>;
+  async loadAllIfExists<T extends Entity>(type: EntityConstructor<T>, _ids: string[], hint?: any): Promise<T[]> {
+    const meta = getMetadata(type);
+    const ids = _ids.map((id) => tagIfNeeded(meta, id));
+    const entities = (
+      await Promise.all(
+        ids.map((id) => {
+          return this.findExistingInstance(id) || this.loaderForEntity(meta).load(id);
+        }),
```
chr1sjf0x, line 479: curious: I haven't looked around here, but I was intuitively thinking this would translate to some sort of `select * from blah where id in [ids]` (and perhaps it is eventually?).

What's the reasoning for doing it one at a time? I guess we are checking the cache first, then loading what we need. Would there be a benefit to loading things that aren't cached like suggested compared to separate queries? Perhaps the query manager is performing that magic under the hoods?

alexkuang: `loaderForEntity` uses `DataLoader` under the hood which handles the batching (I think)

stephenh: Yep

PR #74, opened 2020-11-03
Generate object resolver scaffolding.

stephenh:
This generates `src/resolvers/${entity}/${entity}Resolver[.test].ts`
files.

Historically these files have been scaffolded by our
separate/joist-unaware graphql scaffolding project.

I'd resisted teaching the graphql scaffolding about joist ... however,
I have no qualms about teaching the joist's graphql-codegen plugin
more things.

This works out b/c since joist-codegen runs first, it's resolver
scaffolding (which is more entity-aware) will "win" and the other
graphql-generic scaffolding won't overwrite it (but it will still
create resolver scaffolding for non-joist GQL types as usual).

My primary discomfort is that joist's `integration-tests` project is
getting more GQL cruft when really the core joist functionality is
generally GQL agnostic. Maybe time for a separate integration test
project? Not sure.


chr1sjf0x (approved):
Looks good to me 👍

packages/integration-tests/src/resolvers/entityResolver.ts
```diff
@@ -0,0 +1,3 @@
+import { Entity, EntityMetadata } from "joist-orm";
+
+export function entityResolver<T extends Entity>(entityMetadata: EntityMetadata<T>): any {}
```
zgavin, line 3: Should have taken the chance to change the signature of this to just take an `EntityConstructor` constructor directly.  It's just added boilerplate to take an `EntityMetadata` that's even hardwired into the codegen now.

zgavin: Also sort of confused how consumers that aren't us would actually get an `entityResolver` or understand what it is for.  Seems like we should at least have a default implementation if we're going to be codegenning it?

stephenh: Yeah, so none of these `entityResolver` / etc. utilities are meant for consumption / usage / anything more than "just make sure the currently-generated code compiles for the purposes of the integration test".

I'm lazy punting extracting the utilities "for real" to a future PR / probably some sort of new `joist-graphql-utils` package.

stephenh: And yeah agreed we should change to `EntityConstructor` when we do that.

PR #76, opened 2020-11-03
Generate save resolvers.

stephenh:
Now using the history approach to only generate these files once (and retrofits the previous object resolvers to use the history as well).


packages/integration-tests/src/resolvers/mutations/bookAdvance/saveBookAdvanceResolver.test.ts
```diff
@@ -0,0 +1,20 @@
...
+import { run } from "src/resolvers/testUtils";
+import { saveBookAdvance } from "src/resolvers/mutations/bookAdvance/saveBookAdvanceResolver";
+
+import "src/setupDbTests";
+
+describe("saveBookAdvance", () => {
+  it.withCtx("can create", async (ctx) => {
+    const { em } = ctx;
+    const result = await runSaveBookAdvance(ctx, () => ({}));
+    // const ba = await em.load(BookAdvance, result.BookAdvance);
```
stephenh, line 12: Commenting out b/c these aren't (currently) supposed to do test anything other than the code generation works as expected.

PR #78, opened 2020-11-03
Include extraPrimitives from enum tables in graphql and resolvers

chr1sjf0x:
This takes support for extra columns in enum tables all of the way through the generated graphql code as well as the resolvers.

`enums.graphql`
image

`enumResolvers.ts`
image

### DIFF
--- a/packages/codegen/src/index.ts
+++ b/packages/codegen/src/index.ts
@@ -6,7 +6,7 @@ import { Client } from "pg";
 import { code, Code, def, imp } from "ts-poet";
 import { assignTags } from "./assignTags";
 import { Config, loadConfig, writeConfig } from "./config";
-import { EntityDbMetadata } from "./EntityDbMetadata";
+import { EntityDbMetadata, PrimitiveField } from "./EntityDbMetadata";
 import { generateEntitiesFile } from "./generateEntitiesFile";
 import { generateEntityCodegenFile } from "./generateEntityCodegenFile";
 import { generateEnumFile } from "./generateEnumFile";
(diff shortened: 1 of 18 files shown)


packages/codegen/src/index.ts
```diff
@@ -34,8 +35,15 @@ export interface CodeGenFile {
...
-export type EnumRow = { id: number; code: string; name: string };
+export type EnumTableData = {
+  table: Table;
+  // Pascal case version of table name
+  name: string;
+  rows: EnumRow[];
+  extraPrimitives: PrimitiveField[];
+}
+export type EnumMetadata = Record<string, EnumTableData>;
+export type EnumRow = { id: number; code: string; name: string; [key: string]: any };
```
stephenh, line 46: Cool, nice change

packages/graphql-codegen/src/generateEnumsGraphql.ts
```diff
@@ -18,3 +18,23 @@ export async function generateEnumsGraphql(enums: EnumRows): Promise<CodeGenFile
 
   return { name: "../../schema/enums.graphql", overwrite: true, contents: formatted };
 }
+
+/** Maps db types, i.e. `int`, to GraphQL types, i.e. `Int`. */
+function mapSimpleDbType(dbType: string): string {
```
stephenh, line 23: Seems like we should have this already, is it:

```
function mapFieldTypeToGraphQLType(type: string | SymbolSpec): string | SymbolSpec {
```

Oh, I see, that's from TS type --> db type.

We could export this from `codegen` instead? I believe `graphql-codegen` already accesses exported functions from that package. Like the `config`.

Or basically someone tie together the existing code function that goes db type --> TS type and then the existing TS type --> GQL type.

stephenh (approved):
Cool!

packages/graphql-codegen/src/generateEnumsGraphql.ts
chr1sjf0x: Done, TY 👍

PR #79, opened 2020-11-04
Refactor existing data type converters to reduce duplication

chr1sjf0x:
@stephenh - I tried refactoring the helpers for converting between db types => ts types => graphql types. I generally like this approach, but I'm getting a strange error. Pushing it up here to see if the same problem happens on CircleCI.

Just now realizing it might be `joist-codgen` => `joist-codegen`?
```
Error: Cannot find module 'joist-codgen'
Require stack:
- /Users/homebound/code/joist-ts/packages/graphql-codegen/build/graphqlUtils.js
- /Users/homebound/code/joist-ts/packages/graphql-codegen/build/generateEnumsGraphql.js
- /Users/homebound/code/joist-ts/packages/graphql-codegen/build/index.js
- /Users/homebound/code/joist-ts/packages/codegen/build/index.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:1080:15)
    at Function.Module._load (internal/modules/cjs/loader.js:923:27)

```


packages/graphql-codegen/tsconfig.json
```diff
@@ -15,7 +15,7 @@
     "baseUrl": "./src",
     "paths": {
       "joist-utils": ["../../utils/src"],
-      "joist-codgen": ["../../codegen/src"]
+      "joist-codegen": ["../../codegen/src"]
```
chr1sjf0x, line 18: @stephenh - I presume this was a typo? Let me know if there is some reasoning for this. It resulted in more diffs than I was expecting in the Codegen files.

packages/graphql-codegen/src/graphqlUtils.ts
```diff
@@ -137,3 +139,20 @@ function mergeDocs(existingDoc: DocumentNode, newDocs: [string, DocumentNode][])
...
+      return "Boolean";
+    case "number":
+      return "Int";
+    default:
+      return type;
+  }
+}
+
+export function mapSimpleDbTypeToGraphQLType(type: string): string | SymbolSpec {
+  return mapTypescriptTypeToGraphQLType(mapSimpleDbTypeToTypescriptType(type));
```
stephenh, line 157: Thanks!

packages/graphql-codegen/tsconfig.json
stephenh: Oh wow, yep, good fix.

packages/integration-tests/src/entities/AuthorCodegen.ts
```diff
@@ -1,51 +1,51 @@
 import {
+  Flavor,
```
stephenh, line 2: Fwiw I think this part of the diff is that I snuck in `prettier-organize-imports-plugin` in the last commit, so if you rebase / `yarn install` / run prettier, these `import` diffs should go away...

stephenh (approved):
Thanks for the follow up!

PR #81, opened 2020-11-16
Circular Dependency Fix

zgavin:
Added explicit types for PrimitiveFIeld's columnType / fieldType so they aren't just string and fixed a circular dependency issue in graphql-codegen

### DIFF
--- a/packages/codegen/src/EntityDbMetadata.ts
+++ b/packages/codegen/src/EntityDbMetadata.ts
@@ -28,12 +28,25 @@ export type Entity = {
   optsType: SymbolSpec;
 };
 
+export type DatabaseColumnType =
+  | "boolean"
+  | "int"
+  | "numeric"
+  | "text"
+  | "citext"
+  | "character varying"
+  | "varchar"
+  | "timestamp with time zone"
+  | "date";
+
+export type PrimitiveTypescriptType = "boolean" | "string" | "number" | "Date";
+
 export type PrimitiveField = {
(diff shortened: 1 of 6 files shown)


chr1sjf0x (approved):
TY @zgavin

packages/codegen/src/EntityDbMetadata.ts
```diff
@@ -28,12 +28,25 @@ export type Entity = {
...
+  | "int"
+  | "numeric"
+  | "text"
+  | "citext"
+  | "character varying"
+  | "varchar"
+  | "timestamp with time zone"
+  | "date";
+
+export type PrimitiveTypescriptType = "boolean" | "string" | "number" | "Date";
```
stephenh, line 42: Cool, nice names.

packages/graphql-codegen/src/graphqlUtils.ts
```diff
@@ -1,8 +1,7 @@
 import { DocumentNode, InputObjectTypeDefinitionNode, ObjectTypeDefinitionNode, parse, print, visit } from "graphql";
-import { mapSimpleDbTypeToTypescriptType } from "joist-codegen";
+import { PrimitiveTypescriptType } from "joist-codegen/build/EntityDbMetadata";
```
stephenh, line 2: Hm, does this need to be from the `/build/` directory? Usually just going to the top-level module is more kosher for cross-module imports.

PR #82, opened 2020-11-18
Added utility method to get a constructor from a tagged id

zgavin:


packages/orm/src/index.ts
```diff
@@ -388,9 +389,13 @@ export class ConfigApi<T extends Entity, C> {
   }
 }
 
+const tagToConstructorMap = new Map<string, EntityConstructor<any>>();
```
stephenh, line 392: Wouldn't mind a newline after the `const`/before the `function`.

packages/orm/src/index.ts
```diff
@@ -421,6 +426,11 @@ export function getRelations(entity: Entity): AbstractRelationImpl<any>[] {
   return Object.values(entity).filter((v) => v instanceof AbstractRelationImpl);
 }
 
+export function getConstructorFromTaggedId(id: string): EntityConstructor<any> {
+  const [tag] = id.split(":");
```
stephenh, line 430: There is a `tagDelimiter` `const` in `keys.ts` that in theory "owns" the knowledge of `:`s, maybe put a method in `keys.ts` to get the tag from an id?

stephenh (approved):
Cool

PR #83, opened 2020-12-10
Added reactiveBeforeFlush

zgavin:


packages/orm/src/EntityManager.ts
```diff
@@ -1113,6 +1120,21 @@ async function addReactiveAsyncDerivedValues(todos: Record<string, Todo>): Promi
   await Promise.all(p);
 }
 
+async function addReactiveHooks(todos: Record<string, Todo>): Promise<void> {
+  const p: Promise<void>[] = Object.values(todos).flatMap((todo) => {
+    const entities = [...todo.inserts, ...todo.updates, ...todo.deletes];
+    return todo.metadata.config.__data.reversedHintsForReactiveHooks.map(async (reverseHint) => {
+      (await followReverseHint(entities, reverseHint)).forEach((entity) => {
+        const todo = getTodo(todos, entity);
+        if (!todo.reactiveHooks.includes(entity) && !entity.isDeletedEntity) {
```
stephenh, line 1129: Should we only push this into `reactiveHooks` if the entity is not already in `entities`, i.e. it will have its hooks fired anyway, and we might double fire the hook?

stephenh (approved):
Sweet

PR #84, opened 2020-12-11
added an explicit loader for OneToOneReferences

zgavin:


packages/orm/src/EntityManager.ts
```diff
@@ -745,6 +745,7 @@ export class EntityManager<C extends HasKnex = HasKnex> {
   async refresh(entities: ReadonlyArray<Entity>): Promise<void>;
   async refresh(entityOrListOrUndefined?: Entity | ReadonlyArray<Entity>): Promise<void> {
     this.findLoaders = {};
+    this.__data.loaders = {};
```
zgavin, line 748: I'm not sure why this wasn't here to begin with.  Seems like it would have caused issues for m2m and o2m relations?

packages/orm/src/collections/OneToOneReference.ts
```diff
@@ -35,6 +48,7 @@ export class OneToOneReference<T extends Entity, U extends Entity>
     public otherFieldName: keyof U,
   ) {
     super();
+    this.otherColumnName = otherMeta.columns.find((field) => field.fieldName === otherFieldName)!.columnName;
```
zgavin, line 51: @stephenh is this going to have load order issues? Hoping not to have to make codegen changes.

chr1sjf0x (approved):
TY @zgavin

packages/orm/src/collections/OneToOneReference.ts
stephenh: We should be good. The `metadata` structure should be fully initialized before any entity instances are created.

packages/orm/src/EntityManager.ts
stephenh: Yeah, seems right that it should be here.

stephenh (approved):
Sweet

PR #88, opened 2020-12-29
Add driver interface

stephenh:
Moves all of the SQL calls behind a Driver interface, for running Joist against different databases.

All of the dataloader combining-calls logic is kept inside of EntityManager/entities, so that the driver's methods get already-batched parameters, i.e. "load authors 1, 2, 3" even if there were three separate `em.load(n)` calls.


packages/orm/src/drivers/driver.ts
```diff
@@ -0,0 +1,39 @@
+import Knex from "knex";
+import { JoinRow, ManyToManyCollection } from "../collections/ManyToManyCollection";
+import { OneToManyCollection } from "../collections/OneToManyCollection";
+import { OneToOneReference } from "../collections/OneToOneReference";
+import { Entity, EntityConstructor, EntityManager, EntityMetadata } from "../EntityManager";
+import { FilterAndSettings } from "../QueryBuilder";
+import { JoinRowTodo, Todo } from "../Todo";
+
+/** Isolates all SQL calls that joist needs to make to fetch/save data. */
```
chr1sjf0x, line 9: question: When is `noSQL` support coming?? 😆

chr1sjf0x (approved):
Very cool. Thanks @stephenh

packages/orm/src/drivers/driver.ts
stephenh: :-D PRs accepted. :-) 

Fwiw just the basic load / `loadOneToOne` / `loadManyToOne` / etc. methods wouldn't be very hard in NoSQL--it's the queries like `EntityManager.find` that are a PITA. Basically a wire call per nested layer ... ... which no fair pointing out that is technically how Joist works today anyway. :-)

PR #89, opened 2020-12-30
Use cleaner Narrowable/Const implementation.

stephenh:
From the linked TS issue. The approach is similar to before (...with the caveat that I am still just cargo culting it), but with a new `Const` type that cleans up the `N extends Narrowable` extra/throw-away generic from all of the declaration sites.


chr1sjf0x (approved):
praise: Nice cleanup 👍

PR #96, opened 2021-01-20
Use knex.transaction with a lambda to get the right promise.

stephenh:


packages/orm/src/drivers/PostgresDriver.ts
```diff
@@ -199,25 +199,20 @@ export class PostgresDriver implements Driver {
...
-    const txn = await knex.transaction();
-    em.currentTxnKnex = txn;
-    try {
-      if (isolationLevel) {
-        await txn.raw("set transaction isolation level serializable;");
-      }
-      await beforeTransaction(em, txn);
-      const result = await fn(txn);
-      await afterTransaction(em, txn);
-      await txn.commit();
```
stephenh, line 211: Turns out this `txn.commit` doesn't actually return a rejected promise. You need to do:

```
await txn.commit();
await txn.executionPromise;
```

Instead. Or just use `knex.transaction` with a lambda.

:shrug:

haislip (approved):

stephenh:
Need to follow up with tests...

PR #97, opened 2021-01-25
[Ranch-1048] Return Changed Entities from `em.flush()`

JoshuaScript:
This change makes `em.flush()` return changed entities so that clients can know what specifically changed as a result of their queries

### DIFF
--- a/packages/integration-tests/src/EntityManager.test.ts
+++ b/packages/integration-tests/src/EntityManager.test.ts
@@ -9,7 +9,6 @@ import {
 import { Loaded, setDefaultEntityLimit, setEntityLimit } from "joist-orm";
 import { Author, Book, Publisher, PublisherSize } from "./entities";
 import { knex, newEntityManager, numberOfQueries, queries, resetQueryCount } from "./setupDbTests";
-import {assertNever} from "joist-codegen/build/utils";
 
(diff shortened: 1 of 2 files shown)


packages/integration-tests/src/EntityManager.test.ts
```diff
@@ -9,6 +9,7 @@ import {
 import { Loaded, setDefaultEntityLimit, setEntityLimit } from "joist-orm";
 import { Author, Book, Publisher, PublisherSize } from "./entities";
 import { knex, newEntityManager, numberOfQueries, queries, resetQueryCount } from "./setupDbTests";
+import {assertNever} from "joist-codegen/build/utils";
```
stephenh, line 12: Hm, is there already an `assertNever` inside `integration-tests`? I'd like to avoid the cross-project import especially into the `build` directory. I have admittedly been just copy/pasting `assertNever` around into each project's utils file.

stephenh (approved):
Nice!

packages/integration-tests/src/EntityManager.test.ts
JoshuaScript: Hm, this must have snuck in from an auto-import. I don't need it, going ahead and deleting.

PR #99, opened 2021-01-28
Don't react against fields that are read only.

stephenh:


packages/orm/src/index.ts
```diff
@@ -437,3 +438,34 @@ export function getConstructorFromTaggedId(id: string): EntityConstructor<any> {
...
+  // and OrderOf<T> has id/createdAt/updatedAt (also not in .changed), so just using an index
+  // type for now.
+  changes: { [key: string]: { hasChanged: boolean } };
+}
+
+const cannotBeChangedRule = Symbol("cannotBeChangedRule");
+
+export function cannotBeChanged<T extends Entity & EntityChanges<T>, K extends keyof T & string>(
+  fieldName: K,
+  unless?: (entity: T) => Promise<boolean>,
```
chr1sjf0x, line 455: praise: Nice, I hadn't noticed there was a `unless` option on this before 😆  👍

packages/orm/src/index.ts
```diff
@@ -437,3 +438,34 @@ export function getConstructorFromTaggedId(id: string): EntityConstructor<any> {
...
+): ValidationRule<T> {
+  const fn: ValidationRule<T> = async (entity) => {
+    if (!entity.isNewEntity && entity.changes[fieldName].hasChanged && (!unless || !(await unless(entity)))) {
+      return `${capitalCase(fieldName)} cannot be updated`;
+    }
+    return undefined;
+  };
+  if (!unless) {
+    (fn as any)[cannotBeChangedRule] = fieldName;
+  }
```
chr1sjf0x, line 465: question: Is this indicating that if we specify `unless`, then the rule will still be reactive?

packages/orm/src/index.ts
```diff
@@ -437,3 +438,34 @@ export function getConstructorFromTaggedId(id: string): EntityConstructor<any> {
...
+// The Entity type doesn't model the `changes` structure yet, so do this for now.
+interface EntityChanges<T> {
+  // Ideally we could use Record<keyof T, { hasChanged: boolean }> but we only want
+  // keys that are actually columns. Right now OptsOf<T> has collections (not in .changed)
+  // and OrderOf<T> has id/createdAt/updatedAt (also not in .changed), so just using an index
+  // type for now.
+  changes: { [key: string]: { hasChanged: boolean } };
+}
+
+const cannotBeChangedRule = Symbol("cannotBeChangedRule");
```
chr1sjf0x, line 451: praise: Neat. I haven't seen this use of `Symbol` like this before. I think I `get it` and how it's used, but would also appreciate a @stephenh thought process dump.

chr1sjf0x (approved):

packages/orm/src/index.ts
stephenh: @chr1sjf0x sure, I think it just ensures no conflicts on keys.

Like if we picked a string-based key like `"foo"` but then, in the invariably large/annoying complexities of ~1000-deps-in-`node_modules`, some _other_ library also used `"foo"` for its metadata, we'd stomp on each other.

The symbol just ensures a unique/namespaced lookup on the object.

packages/orm/src/index.ts
stephenh: Yeah, since the reactivity (or lack of reactivity) rules are built once on the initial boot of the system (not at codegen time or at each `em.flush()`), we can't really know for sure if this is always-readonly or always-not-readonly, so have to default to assuming its reactive.

I guess at least for this part of building the initial reactive rule list...

Maybe we could work this conditionally-read-only-ness into the evaluation of the reactive things, but not sure...

PR #100, opened 2021-02-18
Support to `ignore` a field from a Joist Entity

chr1sjf0x:
The goal here is to have the joist codegen process ignore a certain database column, so that it will not be included in the codegen output.

This would be done as a pre-cursor to actually dropping the column from the database, which may need to be performed an a later maintenance window.

- [x] We should fail if the DB column to be ignored is `required`?


packages/codegen/src/EntityDbMetadata.ts
```diff
@@ -50,6 +58,7 @@ export type PrimitiveField = {
   notNull: boolean;
   derived: "orm" | "sync" | "async" | false;
   protected: boolean;
+  softDrop?: boolean;
```
stephenh, line 61: I'm pretty hard -1 on soft drop, I get your explanation of "it's a drop ... but soft" but still just throws me off, would prefer `ignored`. Or other suggestions.

packages/integration-tests/migrations/1580658856631_author.ts
```diff
@@ -40,6 +40,10 @@ export function up(b: MigrationBuilder): void {
     graduated: { type: "date", notNull: false },
     // for testing protected fields
     was_ever_popular: { type: "boolean", notNull: false },
+    // for testing FieldConfig.softDrop
+    soft_drop_used_to_be_useful: { type: "boolean", notNull: false, default: true },
+    soft_drop_legacy_fk_id: foreignKey("publisher_size", { notNull: false }),
```
stephenh, line 45: Nice, thanks for testing all of these use cases

packages/codegen/src/EntityDbMetadata.ts
haislip: `ignored` makes sense.  It _could_ be used when adding a field also (for some reason?), in order to allow this logic to not be dependent on which direction we are going with it.

chr1sjf0x: I have updated this to be `ignore`. I believe it is ready to go now. 👍

packages/codegen/src/config.test.ts
```diff
@@ -0,0 +1,47 @@
+import { Config, defaultConfig, FieldConfig, isFieldIgnored } from "./config";
+import { makeEntity } from "./EntityDbMetadata";
+
+describe("config", () => {
+  describe("isFieldIgnored", () => {
+    // A little hack to test the case of `ignore: false`, since it is actually defined as `ignore?: true`
+    const falseAsTrue = (false as boolean) as true;
+
+    it.each([
+      ["fields key is omitted all together", newAuthorConfig()],
```
stephenh, line 10: Just from reading this, I was unclear what the "then" condition was going to be (it's `false`), i.e. the test name is "config isFieldIgnored fields key is omitted all together" ...

Like are we ignoring here...

Maybe something more like

* "does not ignore when no config is provided"
* "does not ignore when config is set but ignore is unset"
* "does not ignore when ignore is false"
* "does ignore when ignore is true"

I.e. I might work in the next `it` into this each, just b/c it feels like an arbitrary break up for "here are all of the `false` cases" ... when I readed this, where is the true case?... oh here it is, it's the next `it` test. Seems like `it.each` should try to cover all of the boundary cases. 

Granted, this is a trade-off based on the complexity of the mini-DSL/tuples we end up needing to build for `it.each` to accept. And, to be fair, we do having an outer `describe(isFieldIgnored)` to encompass them all. So that's fine, just talking out loud.

stephenh: ...also I'll re-iterate my usual `it.each` bias that I'm not sure this is really a net improvement in readabiliity and long-term dev UX vs. each `it` blocks.

packages/codegen/src/config.ts
```diff
@@ -49,6 +50,15 @@ export function isProtected(config: Config, entity: Entity, fieldName: string):
   return config.entities[entity.name]?.fields?.[fieldName]?.protected === true;
 }
 
+export function isFieldIgnored(config: Config, entity: Entity, fieldName: string, notNull: boolean = false): boolean {
+  const ignore = config.entities[entity.name]?.fields?.[fieldName]?.ignore === true;
+
+  if (ignore && notNull) {
+    fail("notNull fields cannot be ignored. Alter the column to be optional prior to ignoring it.");
```
stephenh, line 57: Interesting, good catch!

packages/integration-tests/migrations/1580658856631_author.ts
```diff
@@ -76,6 +87,9 @@ export function up(b: MigrationBuilder): void {
     name: { type: "varchar(255)", notNull: true },
   });
 
+  // for testing ignore of many to many
```
stephenh, line 90: Great comments, thanks

packages/codegen/src/config.test.ts
```diff
@@ -0,0 +1,47 @@
...
+      ["fields does not specify the field in question", newAuthorConfig({})],
+      ["field is specified but doesn't have ignore key", newAuthorConfig({ shouldNotIgnore: {} })],
```
stephenh, line 12: Fwiw this `shouldNotIgnore` name threw me off just a bit, it almost looks like a config option itself, instead of the (what I finally realized) made-up-field name we used for the test. Maybe using something that is "very obviously a column/field name" like `firstName` would make it more apparent to the reads what this is.

Also "shouldNotIgnore" the name is like mixing the "then" / expected behavior into the code-/data-under-test itself...

packages/integration-tests/joist-codegen.json
```diff
@@ -13,6 +15,14 @@
     "Book": { "tag": "b" },
     "BookAdvance": { "tag": "ba" },
     "BookReview": { "fields": { "isPublic": { "derived": "async" } }, "tag": "br" },
+    "Critic": {
+      "fields": {
+        "ignoreFavouriteBook": { "ignore": true },
+        "ignoreWorstBook": { "ignore": true },
```
stephenh, line 21: Fwiw having `ignore` in the column name itself threw me off ever so slightly in reading some of the test, kinda like the `shouldNotIgnore` field name in the other test.

stephenh (approved):
Only a few minor/ocd nits about naming that made me slow down from "just skimming" but otherwise lgtm! Thanks!

packages/integration-tests/joist-codegen.json
chr1sjf0x: Yea, i could see that. I felt like it was helpful in the migration though to make it clear that you won't be seeing it in the codegen output? I will update the `shouldNotIgnore` field name to something else.

packages/codegen/src/config.test.ts
chr1sjf0x: Cleaned up the messages. TY 👍

PR #105, opened 2021-02-25
[EMPIRE-116] Add cloneEntity function

khalidwilliams:
### [JIRA ticket](https://homebound.atlassian.net/browse/EMPIRE-116)
This PR adds the ability to clone an entity and its nest relations to joist.


packages/integration-tests/src/entities/AuthorCodegen.ts
```diff
@@ -1,51 +1,51 @@
...
   Author,
+  newAuthor,
   authorMeta,
+  Publisher,
+  Image,
   Book,
+  PublisherId,
+  ImageId,
   BookId,
+  PublisherOrder,
```
chr1sjf0x, line 43: request change: It kind of looks like you ran `yarn workspace format`, as requested in the `Readme`, but I think you should revert all these files under `entities`.

It would be nice if we could have these folders be ignored by the formatter, as I ran into that problem aslo.

packages/orm/src/EntityManager.ts
```diff
@@ -398,6 +398,73 @@ export class EntityManager<C = {}> {
     return createOrUpdatePartial(this, type, opts);
   }
 
+  /** Utility to clone an entity and its nested relations, as determined by a populate hint */
+  public async cloneEntity<T extends Entity, H extends LoadHint<T>>(entity: T, hint?: H): Promise<Loaded<T, H>> {
```
chr1sjf0x, line 402: thought: Since this will be used like `em.cloneEntity()`, I wonder if `em.clone()` would be better. Maybe just `clone` will make it more apparent you are going to clone everything on the hint, not just the provided `entity`?

packages/integration-tests/src/EntityManager.test.ts
```diff
@@ -1023,6 +1023,95 @@ describe("EntityManager", () => {
...
+    const a1 = new Author(em, {firstName: "a1"});
+    const b1 = new Book(em, {title: "b1", author: a1});
+    const t1 = new Tag(em, {name: "t1", books: [b1]})
+    await em.flush();
+
+    // When we clone that entity and its nested references
+    const toExecute = async function () { await em.cloneEntity(a1, {books: "tags"})};
+
+    // Then we expect the cloning to fail
+    await expect(toExecute).rejects.toThrow()
```
chr1sjf0x, line 1072: nit: Can you include the message which is thrown please?
```suggestion
    await expect(toExecute).rejects.toThrow("some sort of error")
```

chr1sjf0x (changes requested):
Looks good @khalidwilliams - Minor request changes about the formatting of the files under `entities`

packages/integration-tests/src/EntityManager.test.ts
stephenh: +1

packages/integration-tests/src/EntityManager.test.ts
```diff
@@ -1023,6 +1023,95 @@ describe("EntityManager", () => {
...
+    const a2 =  await em.cloneEntity(a1, {books: "image"});
+    await em.flush();
+
+    // Then we expect the cloned entity to have cloned copies of all its nested references
```
stephenh, line 1092: The flow between this `Then` on line 1092 and the 1st `Then` on line 1087 is throwing me off...

It almost reads like this 2nd `Then` is a left-over copy/paste from a prior test, but it's not...

Can you add like a `// And when ...how is this 2nd cloneEntity different from the first cloneEntity` before teh `const a2` on line 1089?

packages/integration-tests/src/EntityManager.test.ts
```diff
@@ -1023,6 +1023,95 @@ describe("EntityManager", () => {
...
+    const b2 = (await a2.books.load())[0];
+    const i2 = await b2.image.load();
+    expect(i2).toBeTruthy();
+    expect(i2?.id).not.toEqual(i1.id);
+    expect(i2?.fileName).toEqual(i1.fileName);
+    expect(i2?.type).toEqual(i1.type);
```
stephenh, line 1098: Can we add assertions about the overall number of entities in the system? I.e. assert `find(Author, {})` == 2, number of books = 2, number of images = 2 (if that is right I'm kinda guessing).

Just to get a 2nd triangulation for "we created the right number of new entities".

...I'd be tempted to include this "number of authors, books, reviews is now X" so it might be worth a helper method, but defer. 

like `expect(await numberOfAuthors(em)).toEqual(2)` or something.

Maybe `expect(await numberOf(Author, Book, Review)).toEqual([2, 3, 4])` if we wanted to be really cute.

stephenh (commented):
Added a few asks but awesome getting this pulled up into joist, thanks!

packages/orm/src/EntityManager.ts
KoltonG: It might also be helpful to include a usage example!

```
/**
 * Utility to clone an entity and its nested relations, as determined by a populate hint
 *
 * @example
 * // This will duplicate the author and all book entities
 * const duplicateBook = await em.clone(author, "books"); 
 */
```

Following https://jsdoc.app/tags-example.html

packages/integration-tests/src/EntityManager.test.ts
khalidwilliams: This was absolutely a typo from come bad copy-paste 😬 Thanks for catching!

packages/orm/src/EntityManager.ts
khalidwilliams: Yeah @chr1sjf0x I think that could be a good call

packages/integration-tests/src/EntityManager.test.ts
khalidwilliams: I gave it a shot! I just declared the helper at the bottom of this file.

packages/integration-tests/src/EntityManager.test.ts
```diff
@@ -1023,8 +1023,105 @@ describe("EntityManager", () => {
...
+    expect(await a2.books.load()).toHaveLength(0);
+  });
 });
 
 function delay(ms: number): Promise<void> {
   return new Promise((resolve) => setTimeout(resolve, ms));
 }
+
+async function numberOf(em: EntityManager, ...args: EntityConstructor<any>[]): Promise<number[]> {
+  return await Promise.all(
```
stephenh, line 1121: Really minor, you can merge w/o changing, but when you are returning a promise, you don't need to await it, i.e.:

```
return Promise.all(...)
```

Would be fine

packages/integration-tests/src/EntityManager.test.ts
```diff
@@ -1023,8 +1023,105 @@ describe("EntityManager", () => {
...
+    const em = newEntityManager();
+
+    // Given an entity
+    const a1 = new Author(em, { firstName: "a1" });
+    await em.flush();
+
+    // When we clone that entity
+    const a2 = await em.clone(a1);
+
+    // Then we expect the cloned entity to have the same properties as the original
```
stephenh, line 1037: This is checking that a primitive was cloned, could we create a publisher, pass that to the `Author` cstr on line 1031, and then after cloning (with the existing no-populate-hint clone call), add an assertion that `a2.publisher.id === p1.id`? And maybe that `numberOf([Author, Publisher])` is now `[2, 1]`?

Fwiw I think `clone` is already handling this correctly, just thinking I'd like to have it covered by a test, and I didn't see an existing test for this (although cloning the author -> book -> image probably gets really close/probably does...)

packages/integration-tests/src/EntityManager.test.ts
```diff
@@ -1023,8 +1023,105 @@ describe("EntityManager", () => {
...
+    // Given an entity with a reference to another entity with a many-to-many reference
+    const a1 = new Author(em, { firstName: "a1" });
+    const b1 = new Book(em, { title: "b1", author: a1 });
+    const t1 = new Tag(em, { name: "t1", books: [b1] });
+    await em.flush();
+
+    // When we clone that entity and its nested references, which include a many-to-many reference
+    const promise = async function () {
+      await em.clone(a1, { books: "tags" });
+    };
```
stephenh, line 1071:
```suggestion
    const promise = em.clone(a1, { books: "tags" });
```

packages/orm/src/EntityManager.ts
```diff
@@ -398,6 +401,90 @@ export class EntityManager<C = {}> {
...
+   * const duplicatedAuthorAndBooks = await em.clone(author, "books")
+   *
+   * @exmaple
+   * // This will duplicate the author, all their books, and the images for those books
+   * const duplicatedAuthorAndBooksAndImages = await em.clone(author, {books: "image"})
+   */
+  public async clone<T extends Entity, H extends LoadHint<T>>(entity: T, hint?: H): Promise<Loaded<T, H>> {
+    const meta = getMetadata(entity);
+    const { id, ...data } = entity.__orm.data;
+    const clone = this.create(meta.cstr, {} as OptsOf<T>);
```
stephenh, line 425: Fwiw @zgavin I assume doing this didn't work:

```typescript
const { id, ...data } = entity.__orm.data;
const clone = this.create(meta.cstr, data);
```

I.e. passing `data` directly to `create`? Would be nice to have an in-source comment if there is a particularly tricky reason about why that doesn't work, but not a big deal.

packages/orm/src/EntityManager.ts
```diff
@@ -398,6 +401,90 @@ export class EntityManager<C = {}> {
...
+    clone.__orm.data = data;
+    if (hint) {
+      if ((typeof hint as any) === "string") {
```
stephenh, line 428: Interesting, would make a good `normalizeHint` function at some point cc @zgavin ?

packages/orm/src/EntityManager.ts
```diff
@@ -398,6 +401,90 @@ export class EntityManager<C = {}> {
...
+                // Many to One relationships try and fetch their data from the cache if they can,
+                // but the data is pointing to the uncloned entity, resulting in the original relation
+                // being unset.
+                // So we clear this data from the clone to ensure that the relation is set correctly
+                clone.__orm.data[relationName] = undefined;
+                const relationToClone = clone[relationName] as any;
+                relationToClone.set(clonedRelated);
+              }
+            } else {
+              fail(`Irreversible relation: ${relationName}`);
```
stephenh, line 479: Maybe "Uncloneable relation"?

packages/orm/src/EntityManager.ts
```diff
@@ -398,6 +401,90 @@ export class EntityManager<C = {}> {
...
+              | ManyToOneReference<T, any, undefined | never>;
+            if (relation instanceof OneToManyCollection) {
+              const relatedEntities = await relation.load();
+              await Promise.all(
+                relatedEntities.map(async (related) => {
+                  const clonedRelated = await this.clone(related, nested);
+                  // Many to One relationships try and fetch their data from the cache if they can,
+                  // but the data is pointing to the uncloned entity, resulting in the original relation
+                  // being unset.
+                  // So we clear this data from the clone to ensure that the relation is set correctly
```
stephenh, line 448: Could shorten this comment to:

```
// Clear to avoid `set` mutating the original/source entity's relationship
```
 
especially since this comment block is copy/pasted ~3 times (which is otherwise fine)

stephenh (approved):
Few more musings, an ask for a slight tweak/additional coverage for a test, but otherwise looks great!

packages/integration-tests/src/EntityManager.test.ts
khalidwilliams: Ohhh that makes sense. I was worried that I'd end up with a nested promise being returned by `numberOf`

packages/integration-tests/src/EntityManager.test.ts
khalidwilliams: That makes sense to add coverage for, I'll add a publisher in

zgavin (approved):

PR #106, opened 2021-03-05
allow for deleting an object in a transaction

haislip:
Clearing out the original data for deleted objects such that we don't re-run them in follow-on flushes within a transaction.


packages/orm/src/EntityManager.ts
```diff
@@ -760,6 +760,7 @@ export class EntityManager<C = {}> {
           todo.inserts.forEach((e) => this._entityIndex.set(e.id!, e));
           todo.inserts.forEach((e) => (e.__orm.originalData = {}));
           todo.updates.forEach((e) => (e.__orm.originalData = {}));
+          todo.deletes.forEach((e) => (e.__orm.originalData = {}));
```
stephenh, line 763: :-| Great find!

stephenh (approved):

zgavin (approved):
Awesome, thanks for tracking this down 👍

PR #107, opened 2021-03-23
Enhance ValidationError to make it work better with potential operations

chr1sjf0x:
- Make `ValidationError.entity` optional so that we don't need to specify it when making potential operations
- Export `ValidationRuleResult` type to avoid need to make a similar one in graphql-service
- Stamp in the `ValidationError.entity` field as part of `coerceError`

### DIFF
--- a/packages/orm/src/EntityManager.ts
+++ b/packages/orm/src/EntityManager.ts
@@ -1039,15 +1039,15 @@ async function afterCommit(ctx: unknown, todos: Record<string, Todo>): Promise<v
 function coerceError(
   entity: Entity,
   maybeError: string | ValidationError | ValidationError[] | undefined,
-): ValidationError[] {
+): Required<ValidationError>[] {
   if (maybeError === undefined) {
     return [];
   } else if (typeof maybeError === "string") {
     return [{ entity, message: maybeError }];
   } else if (Array.isArray(maybeError)) {
-    return maybeError as ValidationError[];
+    return (maybeError as ValidationError[]).map(ve => ({ entity, ...ve}));
   } else {
-    return [maybeError];
+    return [{ entity, ...maybeError}];
   }
 }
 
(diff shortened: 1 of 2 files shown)


packages/orm/src/EntityManager.ts
```diff
@@ -1039,15 +1039,15 @@ async function afterCommit(ctx: unknown, todos: Record<string, Todo>): Promise<v
 function coerceError(
   entity: Entity,
   maybeError: string | ValidationError | ValidationError[] | undefined,
-): ValidationError[] {
+): Required<ValidationError>[] {
```
stephenh, line 1042: Hm. Feels a little weird to have to add the `Required` here.

Also I kinda like `entity` being required on `ValidationError`...

Just thinking out loud, right now the "hey maybe its string | error | error[]" fuzziness was to allow the rule lambdas the most flexibility / developer ergonomics in what they return as rules.

But that same flexibility is kinda annoying on the caller side of things, when you want something more structured.

Wdyt about making `ValidationError` like an ADT of `EntityValidationError | EntityFieldValidationRule | GenericValidationError` which each have a message, and the entity has the entity, entity field has entity + field, and generic is just hte message.

Or something like that.

Then this `coerceError`'s job is to map the "fuzzy on purpose" validation rule return value to the strict "one of these kinds".

Dunno?

stephenh (commented):
Nice! +1 to moving potential operation-ish things upstream.

packages/orm/src/EntityManager.ts
chr1sjf0x: Agreed that it is kind of strange RE `Required`...

I like your approach. I'll try pushing something up along those lines.

chr1sjf0x: I think this refactor ticks most of the boxes on your suggestion. Happy to discuss further. 👍

stephenh (approved):

PR #115, opened 2021-04-20
Use Collection.set in setOpts even with partial=true.

stephenh:
This fixes wonkiness around "I think I'm exhaustively setting a
colection but I'm not really".

Granted, historically we might have relied upon this as a feature,
although AFAIU what's more likely is that we specifically wanted
to control child entity lifecycle via commands like `delete: true`
and less that we wanted incremental collections specifically.

(I.e. all of our parent + line items mutations, which are our most common
usage of this pattern, post exhaustive collections of the child line items.)

(Also, IMO it's likely that updating collections incrementally was also
convenient b/c the `.add` and `.remove` calls "work" without being
loaded, and `setOpts` is synchronous, so I'm tempted to think this was
also out of convenience than raw need.)

(Separately, ManyToManyCollection.add is probably still broken if the
m2m row already exists, i.e. we need to "on conflict ignore" in that
case.)

### DIFF
--- a/packages/orm/src/createOrUpdatePartial.ts
+++ b/packages/orm/src/createOrUpdatePartial.ts
@@ -48,6 +48,7 @@ export async function createOrUpdatePartial<T extends Entity>(
   const { id, ...others } = opts as any;
   const meta = getMetadata(constructor);
   const isNew = id === null || id === undefined;
+  const collectionsToLoad: string[] = [];
 
   // The values in others might be themselves partials, so walk through and resolve them to entities.
   const p = Object.entries(others).map(async ([key, value]) => {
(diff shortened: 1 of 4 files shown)


packages/orm/src/createOrUpdatePartial.ts
```diff
@@ -138,6 +140,12 @@ export async function createOrUpdatePartial<T extends Entity>(
     return em.createPartial(constructor, _opts);
   } else {
     const entity = await em.load(constructor, id);
+    // For o2m and m2m .set to work, they need to be loaded so that they know what to remove.
+    // Note that we also have the `delete: true` pattern for flagging not only "remove" but "delete",
+    // for a parent's mutation to control the lifecycle of a child entity (i.e. line items).
+    // Musing: Maybe this should happen implicitly, like if a LineItem.parent is set to null, that
+    // LineItem knows to just `em.delete` itself? Instead of relying on hints from GraphQL mutations.
+    await Promise.all(collectionsToLoad.map((fieldName) => (entity as any)[fieldName].load()));
```
stephenh, line 148: Fwiw m2m collections could be optimized, where we don't have to load 1000 m2m rows from the db, just to know a `parent.rows = [a,b]` should delete 998 of them, i.e. we could instead inside of `M2MCollection.set` just tell the UoW "delete my existing m2m rows with bulk sql before doing my inserts", but I wanted to leave that as a follow up optimization.

haislip (approved):

stephenh:
Thanks @haislip for the review; going to merge to pre-flight on a graphql-service PR but will hold off on merging that one until auditing of our current mutations is done.

PR #118, opened 2021-05-04
added codegen for boolean accessors to enum values

zgavin:


packages/integration-tests/src/entities/PublisherCodegen.ts
```diff
@@ -175,6 +180,30 @@ export abstract class PublisherCodegen extends BaseEntity {
     setField(this, "size", size);
   }
 
+  get isSizeSmall(): boolean {
```
stephenh, line 183: Sweet

stephenh (approved):

PR #119, opened 2021-05-05
fix to make the behavior of undefined values consistent for find and findGql

zgavin:


stephenh (approved):
Nice! Add a regression test so I don't break this in the future?

PR #121, opened 2021-05-20
chore(build): rework circleci config (matrix build, test summarization)

blimmer:
This PR reworks the CircleCI Build process to test Joist against multiple versions of Node. It also enables the JUnit test reporter so we get pretty CircleCI test reports.

Screen Shot 2021-05-20 at 10 47 12 AM

Screen Shot 2021-05-20 at 10 54 38 AM

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,7 @@ workflows:
               # We should test against "latest Node 16", but there's an issue with 16.2
               # This will be updated soon, along with a bug report to Node.js
               # node-version: ["lts/erbium", "lts/fermium", "16"]
-              node-version: ["lts/erbium", "lts/fermium", "16.1", "16.2"]
+              node-version: ["lts/erbium", "lts/fermium", "16.1"]
       - release:
           requires:
             - build
@@ -33,6 +33,7 @@ jobs:
     steps:
       - checkout
       - node/install:
+          install-npm: false
           node-version: << parameters.node-version >>
       - run: yarn install
       - run: yarn build
@@ -51,7 +52,13 @@ jobs:
           name: Run Test Suite
           command: |
             cd packages/integration-tests
-            yarn test
+            yarn test --reporters=default --reporters=jest-junit
+          environment:
+            JEST_JUNIT_OUTPUT_DIR: /tmp/reports/junit/
+      - store_test_results:
+          path: /tmp/reports/junit/
+      - store_artifacts:
+          path: /tmp/reports/junit
   release:
     machine:
       image: ubuntu-2004:202104-01
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,7 @@ workflows:
               # We should test against "latest Node 16", but there's an issue with 16.2
               # This will be updated soon, along with a bug report to Node.js
               # node-version: ["lts/erbium", "lts/fermium", "16"]
-              node-version: ["lts/erbium", "lts/fermium", "16.1"]
+              node-version: ["lts/erbium", "lts/fermium", "16.1", "16.2"]
       - release:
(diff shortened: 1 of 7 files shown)


.circleci/config.yml
```diff
@@ -1,21 +1,73 @@
...
 jobs:
   build:
-    docker:
-      - image: circleci/node:14.8.0
+    parameters:
+      node-version:
+        description: Node version to run tests under (set by matrix build config)
+        type: string
+    machine:
+      image: ubuntu-2004:202104-01
```
blimmer, line 32: Instead of doing the docker-in-docker trick, we're now using a "machine" image (VM) so we don't have to build a tests docker container.

.circleci/config.yml
```diff
@@ -1,21 +1,73 @@
...
-      - setup_remote_docker
-      - run: |
-          cd packages/integration-tests
-          docker-compose build db tests
-          docker-compose up db-wait
-          docker-compose up -d tests
-      - run: |
-          cd packages/integration-tests
-          docker-compose exec tests ./tests.sh
-      - run: |
```
blimmer, line 17: This release step is now a separate workflow that depends on `build`. It accomplishes the same thing, it just leans on CircleCI a bit more than this bash `if` statement.

.circleci/config.yml
```diff
@@ -1,21 +1,73 @@
...
-          cd packages/integration-tests
-          if [ "$CIRCLE_BRANCH" = "main" ]; then
-            docker-compose exec tests ./release.sh
-          fi
+      - node/install:
+          install-npm: false
+          node-version: << parameters.node-version >>
```
blimmer, line 37: The machine images have `nvm` pre-installed, so this step pulls the `nvm` node version from the matrix build and configures it as the default.

.circleci/config.yml
```diff
@@ -1,21 +1,73 @@
...
+          path: /tmp/reports/junit/
+      - store_artifacts:
+          path: /tmp/reports/junit
+  release:
+    machine:
+      image: ubuntu-2004:202104-01
+    steps:
+      - checkout
+      - node/install:
+          lts: true
```
blimmer, line 68: I figure we just use the latest LTS version to publish new versions of Joist.

.circleci/config.yml
```diff
@@ -1,21 +1,73 @@
...
+  version: 2
+  workflow:
+    jobs:
+      - build:
+          matrix:
+            parameters:
+              # We should test against "latest Node 16", but there's an issue with 16.2
+              # This will be updated soon, along with a bug report to Node.js
+              # node-version: ["lts/erbium", "lts/fermium", "16"]
+              node-version: ["lts/erbium", "lts/fermium", "16.1"]
```
blimmer, line 16: This felt reasonable - currently supported "latest" LTS versions and "current" (see comment above about why it's not just "latest 16".

.circleci/config.yml
stephenh: Fwiw I kinda liked the bash `if` because it's faster (handwaving on how much faster) than doing the install build/compile again in a new job, but :shrug: not a big deal.

stephenh (approved):

chr1sjf0x (approved):
Nice work @blimmer 👍

.circleci/config.yml
blimmer: Ah, I'm glad you called this out. I assumed the `publish` step would run the build and whatnot, but I'm not sure of that. I went ahead and restored this as part of the `build` step with a conditional block. See [`1996b16` (#121)](https://github.com/stephenh/joist-ts/pull/121/commits/1996b1659fe5aa8c830a049e7bb83e2ac26e480e)

PR #123, opened 2021-05-20
fix: fix release command

blimmer:


.circleci/config.yml
```diff
@@ -61,5 +61,4 @@ jobs:
             - run:
                 name: Release
                 command: |
-                  cd packages/integration-tests
```
blimmer, line 64: Whoops, this was added accidentally - the release script is at the root of the repo.

stephenh (approved):

PR #128, opened 2021-05-26
refactored isNewEntity behavior in afterCommit hooks

zgavin:
`isNewEntity` should now return true during an `afterCommit` if the entity was inserted as part of that flush

### DIFF
--- a/packages/orm/src/EntityManager.ts
+++ b/packages/orm/src/EntityManager.ts
@@ -8,6 +8,7 @@ import { loadDataLoader } from "./dataloaders/loadDataLoader";
 import { Driver } from "./drivers/driver";
 import {
   assertIdsAreTagged,
+  BaseEntity,
   Collection,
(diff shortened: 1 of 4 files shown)


packages/orm/src/EntityManager.ts
```diff
@@ -1117,18 +1121,15 @@ async function afterCommit(ctx: unknown, todos: Record<string, Todo>): Promise<v
   await runHook(ctx, "afterCommit", todos, ["inserts", "updates"]);
 }
 
-function coerceError(
-  entity: Entity,
-  maybeError: ValidationRuleResult<any>,
-): ValidationError[] {
+function coerceError(entity: Entity, maybeError: ValidationRuleResult<any>): ValidationError[] {
```
zgavin, line 1124: think these changes to `coerceError` were just from prettier

stephenh (approved):

PR #129, opened 2021-06-01
added explicit isLoaded to all relation types

zgavin:
also refactored `isReference` / `isCollection` to actually check against the concrete implementation classes instead of just seeing if the object looked like a relation

### DIFF
--- a/packages/orm/src/index.ts
+++ b/packages/orm/src/index.ts
@@ -1,3 +1,11 @@
+import {
+  CustomCollection,
+  CustomReference,
+  ManyToManyCollection,
+  ManyToOneReference,
+  OneToManyCollection,
+  OneToOneReference,
+} from "./collections";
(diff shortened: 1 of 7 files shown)


packages/orm/src/index.ts
```diff
@@ -67,12 +75,14 @@ export interface Reference<T extends Entity, U extends Entity, N extends never |
...
 
+  readonly isLoaded: boolean;
+
   load(opts?: { withDeleted: boolean }): Promise<U | N>;
 
   set(other: U | N): void;
 
   /** Returns `true` if this relation is currently set (i.e. regardless of whether it's loaded, or if it is set but the assigned entity doesn't have an id saved. */
-  isSet: boolean;
+  readonly isSet: boolean;
```
stephenh, line 85: Nice

stephenh (approved):
Nice

PR #130, opened 2021-06-03
chore: re-enable node 16 testing

blimmer:
This fix in Node 16.3.0 should fix compatibility with Joist / jest: https://github.com/nodejs/node/pull/38821

### DIFF
--- a/packages/orm/package.json
+++ b/packages/orm/package.json
@@ -7,6 +7,9 @@
     "url": "https://github.com/stephenh/joist-ts.git",
     "directory": "packages/joist-orm"
   },
+  "engines": {
+    "node": "^12.17.0 || ^13.10.0 || ^14 || >= 16.0.0 < 16.2.0 || >= 16.3.0"
+  },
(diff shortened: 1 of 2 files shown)


packages/orm/package.json
```diff
@@ -7,6 +7,9 @@
     "url": "https://github.com/stephenh/joist-ts.git",
     "directory": "packages/joist-orm"
   },
+  "engines": {
+    "node": "^12.17.0 || ^13.10.0 || ^14 || >= 16.0.0 < 16.2.0 || >= 16.3.0"
```
blimmer, line 11: basically - don't allow installing on 16.2.0. The 12.x and 13.x were pulled from the docs: https://nodejs.org/api/async_hooks.html#async_hooks_class_asynclocalstorage

stephenh: Grumble json and no in-source comments. But +1!

stephenh (approved):

PR #132, opened 2021-06-09
added a em.touch method to force entities into a flush

zgavin:


packages/orm/src/BaseEntity.ts
```diff
@@ -20,6 +20,7 @@ import {
 export abstract class BaseEntity implements Entity {
   abstract id: string | undefined;
   private __isNewEntity: boolean = true;
+  private __isTouched: boolean = false;
```
stephenh, line 23: Maybe squirrel this into `__orm`? Maybe the `__orm` thing wasn't a good idea, we look like 3/4rd inside `__orm`, a few things sneaking out...

packages/orm/src/BaseEntity.ts
```diff
@@ -67,8 +68,12 @@ export abstract class BaseEntity implements Entity {
     return Object.keys(this.__orm.originalData).length > 0;
   }
 
+  get isTouchedEntity(): boolean {
```
stephenh, line 71: I think I would expect `Entity.isTouchedEntity` to become true after calling any setter.

Can we make this more hidden, not part of the public API?

Then I would care less about "what touched as a term might intuitively mean" vs. "in the joist context it specifically meant you called `em.touch`"

packages/integration-tests/src/EntityManager.test.ts
```diff
@@ -1125,6 +1125,17 @@ describe("EntityManager", () => {
...
   });
+
+  it("can touch an entity to force it to be flushed", async () => {
+    await insertAuthor({ first_name: "a1" });
+    const em = newEntityManager();
+    const a1 = await em.load(Author, "1");
+    em.touch(a1);
+    expect(a1["__isTouched"]).toBeTruthy();
+    await em.flush();
+    expect(a1.beforeFlushRan).toBeTruthy();
```
stephenh, line 1136: Should we assert that other things like derived fields ran? I'm fine trusting it that it does happen. Just if it'd be easy.

packages/orm/src/BaseEntity.ts
haislip: hrm... .that's interesting, because you're saying "setting but not changing a value" should be considered touched? which is exactly what we don't want in this case.  If that's true, maybe we call this forceFlush or something instead of touch

packages/integration-tests/src/EntityManager.test.ts
zgavin: I don't think that's necessary.  Actually think i might change this to just check the return of the flush, since it returns `entitiesToFlush`.

stephenh (approved):

PR #133, opened 2021-06-28
[HACKDAY] added `populate` method to Promise to chain off any load / find call

zgavin:
not 100% on this being the right way to go, but it compiles and actually works 🙃

### DIFF
--- a/packages/integration-tests/src/EntityManager.populate.test.ts
+++ b/packages/integration-tests/src/EntityManager.populate.test.ts
@@ -80,6 +80,15 @@ describe("EntityManager.populate", () => {
     expect(book.tags.get.length).toEqual(0);
   });
 
+  it("can populate via promise", async () => {
+    await insertAuthor({ first_name: "a1" });
+    await insertBook({ title: "b1", author_id: 1 });
+    const em = newEntityManager();
+    const book = await em.load(Book, "1").populate(["author", "tags"]);
+    expect(book.author.get.firstName).toEqual("a1");
+    expect(book.tags.get.length).toEqual(0);
+  });
+
   it("can populate a list", async () => {
     await insertAuthor({ first_name: "a1" });
     await insertBook({ title: "b1", author_id: 1 });
--- a/packages/integration-tests/src/EntityManager.populate.test.ts
+++ b/packages/integration-tests/src/EntityManager.populate.test.ts
@@ -80,13 +80,79 @@ describe("EntityManager.populate", () => {
     expect(book.tags.get.length).toEqual(0);
   });
 
-  it("can populate via promise", async () => {
+  it("can populate via promise on direct load", async () => {
+    // Given a book with an author and no tags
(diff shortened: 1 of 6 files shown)


packages/orm/src/Promise.ts
```diff
@@ -0,0 +1,38 @@
+import { Entity, Loaded, LoadHint } from "./EntityManager";
+import { getEm } from "./index";
+
+export {};
+
+declare global {
+  interface Promise<T> {
```
stephenh, line 7: I appreciate the creativity here :-) but yeah, I'm not comfortable with this level of prototype hacking, both a) outside of `graphql-service`, and b) these `populate` methods only allow to promises of entities and not all promises.

packages/integration-tests/src/EntityManager.populate.test.ts
```diff
@@ -80,6 +80,81 @@ describe("EntityManager.populate", () => {
...
+  });
+
+  it("can propagate undefined through a populate promise", async () => {
+    // Given an author with no publisher
+    await insertAuthor({ first_name: "a1" });
+
+    // When we load the author then try to load and populate through publisher
+    const em = newEntityManager();
+    const author = await em.load(Author, "a:1");
+    const publisher = await author.publisher.load().populate("images");
```
stephenh, line 104: And this is what we originally wanted to be `author.publisher.load("images")` but doesn't jive with TS? 

Can you just the return type of `Reference.load` to `Promise<...> & { populate: ... }`, and then monkey-patch `populate` directly onto that single `promise` instead of the global `Promise` prototype?

I guess I wouldn't be surprised if, to do something like:

```
const promise = em.load(....);
promise.populate = ....
return promise;
```

We can't use `async` functions b/c the compiler/runtime rewriting of the function might create it's own promise, and not use the promise that we monkey-patched with a `populate` key.

packages/orm/src/Promise.ts
zgavin: was thinking maybe we just provide the method definitions  in their own interface and a helper function that would add the implementation to `Promise.prototype`.  Then consumers could apply it globally on an opt in basis.

eg we'd define something like this:

```typescript
export interface PromisePopulate<T> {
  populate<T extends Entity, H extends LoadHint<T>>(this: Promise<T>, hint: H): Promise<Loaded<T, H>>;
  ...
}
export function addPopulateToPromise() { ... }
```

and then codebases that use joist could add this somewhere to get the behavior:

```typescript
declare global { interface Promise<T> extends PromisePopulate<T> {} }
addPopulateToPromise();
```

packages/integration-tests/src/EntityManager.populate.test.ts
zgavin: iirc adding a `Loaded` or `LoadHint` anywhere in the definition of `Reference` or `Collection` resulted in circular reference compiler errors

PR #136, opened 2021-08-04
Add incremental collection support via op keyword.

stephenh:
Our collections used to be incremental-by-default, but we removed
that b/c it was unintuitive / violated least-surprise to send in
`collection: [1, 2]` and have that leave an existing `3` in the
collection.

So we changed the behavior to exhaustive only.

This adds back opting in to incremental behavior, if all children
in a given `set` operation have an `op` keyword that, for each
child, specifies whether to include/remove/delete it.


packages/orm/src/createOrUpdatePartial.ts
```diff
@@ -29,7 +29,15 @@ type AllowRelationsToBeIdsOrEntitiesOrPartials<T> = {
   [P in keyof T]: T[P] extends NullOrDefinedOr<infer U>
     ? U extends Array<infer V>
       ? V extends Entity
-        ? Array<V | (DeepPartialOrNull<V> & { delete?: boolean | null; remove?: boolean | null }) | IdOf<V>> | null
+        ? Array<
+            | V
+            | (DeepPartialOrNull<V> & {
+                delete?: boolean | null;
+                remove?: boolean | null;
+                op?: "remove" | "delete" | "include";
```
stephenh, line 37: I went with `include` instead of `add` b/c they're sets so like sending in `op: add` multiple times isn't really re-adding it... :thinking:

packages/integration-tests/src/EntityManager.createOrUpdatePartial.test.ts
```diff
@@ -215,6 +215,57 @@ describe("EntityManager", () => {
...
+    await insertBook({ title: "b1", author_id: 1 });
+    await insertTag({ name: "t1" });
+    await insertTag({ name: "t2" });
+    await insertBookToTag({ tag_id: 1, book_id: 1 });
+    await insertBookToTag({ tag_id: 2, book_id: 1 });
+    const em = newEntityManager();
+    // When we incrementally remove a single tag
+    await em.createOrUpdatePartial(Book, { id: "b:1", tags: [{ id: "t:2", op: "remove" }] });
+    await em.flush();
+    // Then we removed only that one m2m row
```
haislip, line 230: Can we check that this has deleted the proper line? Since we're going from 2 to 1... it's not CERTAIN that this deleted the right one instead of deleting "all but"

haislip (approved):

packages/integration-tests/src/EntityManager.createOrUpdatePartial.test.ts
stephenh: Yep, done

PR #141, opened 2021-09-28
Fix: Support multiple primitive filters as ANDs.

stephenh:
The types supported `{ gte: 1, lte: 5 }` but the parsing did not.


ogrodnek (approved):
thanks @stephenh !

PR #144, opened 2021-09-30
Add jsonb support

stephenh:


packages/codegen/src/generateMetadataFile.ts
```diff
@@ -46,15 +47,27 @@ function generateColumns(dbMetadata: EntityDbMetadata): {
...
+    const { fieldName, columnName, columnType, superstruct } = p;
     const serdeType = columnType === "numeric" ? DecimalToNumberSerde : SimpleSerde;
-    return code`
-      {
-        fieldName: "${fieldName}",
-        columnName: "${columnName}",
-        dbType: "${columnType}",
-        serde: new ${serdeType}("${fieldName}", "${columnName}"),
-      },`;
+    if (superstruct) {
```
stephenh, line 52: If a primitive field is a jsonb/backed by a superstruct, we go through a different serde helper that will apply the superstruct `assert` function to any value from the db, to make sure it matches what we expect.

packages/graphql-codegen/src/generateEnumsGraphql.ts
```diff
@@ -1,14 +1,21 @@
...
+import { formatGraphQL, isJsonbColumn, mapTypescriptTypeToGraphQLType, SupportedTypescriptTypes } from "./graphqlUtils";
 
 /** Generates a `schema/enums.graphql` with GQL enums that match all of our domain enums. */
 export async function generateEnumsGraphql(enums: EnumMetadata): Promise<CodeGenFile> {
   const contents = Object.values(enums)
     .map(({ name, rows, extraPrimitives }) => {
       const enumDecl = `enum ${name} { ${rows.map((r) => r.code).join(" ")} }`;
       const detailDecl = `type ${name}Detail { code: ${name}! name: String! ${extraPrimitives
-        .map((p) => `${camelCase(p.columnName)}: ${mapTypescriptTypeToGraphQLType(p.fieldName, p.fieldType)}!`)
+        .filter((p) => !isJsonbColumn(p))
```
stephenh, line 11: For now just not supporting `jsonb` columns as extra columns on primitives... Should be doable in a future PR.

stephenh: Specifically in the GQL output...I'm not actually sure what they'd look like in the domain model, they might work-ish? Not sure.

haislip (approved):

PR #149, opened 2021-10-04
Use null for o2o defaults.

stephenh:


packages/integration-tests/src/EntityManager.factories.test.ts
```diff
@@ -179,4 +183,22 @@ describe("EntityManager.factories", () => {
     expect(a1.firstName).toEqual("a1");
     expect(a1.lastName).toBeUndefined();
   });
+
+  it("should default children to empty array if created bottom-up", async () => {
```
stephenh, line 187: Forgot that we had a file dedicated to factory tests, so moved old test + new test here.

haislip (approved):

PR #164, opened 2021-11-01
Change hasChanged to be true on create.

stephenh:
It seems like we almost always do `foo.hasChanged || isNewEntity`, i.e. `hasChanged` should just be true for set fields on new entities.

### DIFF
--- a/packages/integration-tests/src/entities/Author.ts
+++ b/packages/integration-tests/src/entities/Author.ts
@@ -89,7 +89,7 @@ authorConfig.addRule((a) => {
 });
 
 authorConfig.addRule((a) => {
-  if (a.changes.lastName.hasChanged) {
+  if (!a.isNewEntity && a.changes.lastName.hasChanged) {
     return "lastName cannot be changed";
   }
 });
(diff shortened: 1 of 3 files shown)


packages/integration-tests/src/entities/Author.ts
```diff
@@ -89,7 +89,7 @@ authorConfig.addRule((a) => {
 });
 
 authorConfig.addRule((a) => {
-  if (a.changes.lastName.hasChanged) {
+  if (!a.isNewEntity && a.changes.lastName.hasChanged) {
```
chr1sjf0x, line 92: While I approve of the changes to this PR, I do worry that bringing this version of joist into our code base may result in hard to find bugs where the old behviour was assumed. 🤔

chr1sjf0x (approved):

packages/integration-tests/src/entities/Author.ts
stephenh: Yeah, we do have ~90 calls to `hasChanged` currently in the codebase, but probably still doable to audit + run the tests and :crossed_fingers: should be good enough. If we miss one or two I think it's worth the long-term UX win.

PR #165, opened 2021-11-01
Pass use entities directly to factories.

stephenh:
Previously we looked them up after a factory has called newTestInstance,
which worked but meant that the factory's internal default behavior
logic might kick in, instead of using the entity passed in `use`.


chr1sjf0x (approved):
Neat approach. TY @stephenh 👍

PR #174, opened 2021-11-08
Fix array columns.

stephenh:
Turns out array columns where just horribly broken because of how
our UPDATE statements work.

We had used a trick where, to update 10 authors in 1 SQL statement,
we send 1 SQL UPDATE and then make a mini-temp-table by sending
in 10 first names, 10 last names, etc., each as an array, and then
`unnest`-ing each array into a given column, i.e. we'd get a temp
table with 10 rows, with each row's first name/last name/etc.

When I added array support, I tried to treat the array column in
this scenario as a "list of lists", i.e. something like 1st entity's
colors is [1], 2nd entity's colors is [2, 3], so send over [[1], [2,3]]
as a 2-dimensional array.

However, 2-dimensional arrays are not sparse, i.e. they have to be
[[1, 4], [2, 3]], which of course just doesn't make sense/can't work
for what I was trying to use them for.

So, instead if enum arrays are involved, we fallback to a different
UPDATE syntax that is still 1 statement, but uses multiple `VALUES
(...)` clauses in the SQL to create the mini-temp-table.

So, the same goal is achieved, just with a slightly longer SQL
statement, as the SQL statement itself needs to repeat the column
list for each entity being updated. But this shouldn't be a big deal,
and is actually how the INSERTs work, as we use a knex.batchInsert
method that uses the same `VALUES (...), (...)` approach.

I'm still keeping the unnest-based approach for when enum array
columns aren't being updated, b/c, dunno, it seems slightly better
given the SQL string is a constant size.

### DIFF
--- a/packages/integration-tests/src/EntityManager.test.ts
+++ b/packages/integration-tests/src/EntityManager.test.ts
@@ -1193,12 +1193,35 @@ describe("EntityManager", () => {
     expect(rows[0].favorite_colors).toEqual([2]);
   });
 
-  it("can save an empty enum array", async () => {
+  it("can create an empty enum array", async () => {
     const em = newEntityManager();
     em.create(Author, { firstName: "a1" });
     await em.flush();
     const rows = await knex.select("*").from("authors");
     expect(rows[0].favorite_colors).toEqual([]);
+  });
+
+  it("can update an empty enum array", async () => {
+    await insertAuthor({ first_name: "f", favorite_colors: [1, 2] });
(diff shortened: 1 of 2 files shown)


DeanGilewicz (approved):
Thanks so much for the explanation!

For additional context: I originally came across this issue when trying to remove a `lineItem` from a `commitment` and attempting to recalculate / set `costTypes` for that `commitment` to an empty array.

Really appreciate the low-level work on this!

PR #175, opened 2021-11-11
feat: Add EntityManager.load(string) overload.

stephenh:


blimmer (approved):

packages/orm/src/EntityManager.ts
```diff
@@ -505,7 +507,16 @@ export class EntityManager<C = {}> {
     id: string,
     populate: Const<H>,
   ): Promise<Loaded<T, H>>;
-  async load<T extends Entity>(type: EntityConstructor<T>, id: string, hint?: any): Promise<T> {
+  async load<T extends Entity>(typeOrId: EntityConstructor<T> | string, id?: string, hint?: any): Promise<T> {
```
chr1sjf0x, line 510: praise: Nice, TY 👍 

question: Do you see this being extended to the other `load` methods, like `loadAll`?

chr1sjf0x (approved):

packages/orm/src/EntityManager.ts
stephenh: @chr1sjf0x ah yeah, kinda forgot about those, I'll go ahead and try to add support for all of them in this PR. If one or two ends up being weird/difficult, I'll hold off, but will do any that are easy this time around. Good call!

stephenh: Hrm, turns out those are a little harder b/c we may end up with ids across multiple entities, and so have a single `loadAll` turn into multiple queries. Which is totally doable, but big enough to punt to the next PR. :-)

PR #181, opened 2021-11-18
Implement optimistic locking.

stephenh:
Per discussions, this is pretty easy for the ORM to keep track of and
just handle for us, and we can't really be sure how often it's happening
today because we can't detect it (although it is almost certainly
basically never).

### DIFF
--- a/packages/orm/src/drivers/PostgresDriver.ts
+++ b/packages/orm/src/drivers/PostgresDriver.ts
@@ -18,6 +18,7 @@ import {
   maybeResolveReferenceToId,
   OneToManyCollection,
   OneToOneReference,
+  tagIds,
 } from "../index";
 import { JoinRow, ManyToManyCollection } from "../relations/ManyToManyCollection";
 import { JoinRowTodo, Todo } from "../Todo";
@@ -227,7 +228,11 @@ export class PostgresDriver implements Driver {
           await batchInsert(knex, meta, todo.inserts);
         }
         if (todo.updates.length > 0) {
-          todo.updates.forEach((e) => (e.__orm.data["updatedAt"] = updatedAt));
+          todo.updates.forEach((e) => {
+            // Should we just go through a setter?
+            e.__orm.originalData["updatedAt"] = e.__orm.data["updatedAt"];
+            e.__orm.data["updatedAt"] = updatedAt;
+          });
           await batchUpdate(knex, meta, todo.updates);
         }
         if (todo.deletes.length > 0) {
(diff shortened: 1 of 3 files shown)


packages/orm/src/drivers/PostgresDriver.ts
```diff
@@ -227,7 +228,11 @@ export class PostgresDriver implements Driver {
           await batchInsert(knex, meta, todo.inserts);
         }
         if (todo.updates.length > 0) {
-          todo.updates.forEach((e) => (e.__orm.data["updatedAt"] = updatedAt));
+          todo.updates.forEach((e) => {
+            // Should we just go through a setter?
+            e.__orm.originalData["updatedAt"] = e.__orm.data["updatedAt"];
+            e.__orm.data["updatedAt"] = updatedAt;
```
stephenh, line 234: Previously when we bumped `updatedAt`, we weren't putting the old value in `originalData`, but we need the old value to do the "WHERE id = ... and updated_at = original_updated_at"

packages/orm/src/drivers/PostgresDriver.ts
```diff
@@ -379,31 +387,50 @@ async function batchUpdate(knex: Knex, meta: EntityMetadata<any>, entities: Enti
...
-                .join(",")})`,
+                .join(", ")}, ?::timestamp)`,
           )
           .join(",")}
-      ) AS data(${columns.map((c) => c.columnName).join(",")})
-      WHERE ${meta.tableName}.id = data.id
+      ) AS data(${columns.map((c) => c.columnName).join(",")},original_updated_at)
+      WHERE
+        ${meta.tableName}.id = data.id
+        AND date_trunc('milliseconds', ${meta.tableName}.updated_at) = data.original_updated_at
```
stephenh, line 396: The `date_trunc` is because PG has higher resolution than the JS dates we have, so need to round to milliseconds.

packages/orm/src/drivers/PostgresDriver.ts
```diff
@@ -379,31 +387,50 @@ async function batchUpdate(knex: Knex, meta: EntityMetadata<any>, entities: Enti
...
+      RETURNING ${meta.tableName}.id
    `;
   } else {
     // Issue 1 UPDATE statement with bindings having an array for each column's new values
     bindings = columns.map(() => []);
+    // Add 1 more column for original_updated_at
+    const originalUpdatedAt: Date[] = [];
+    bindings.push(originalUpdatedAt);
```
stephenh, line 404: Our `UPDATE`s work by basically uploading a mini-table of the changed rows, and this added `originalUpdatedAt` as a new "column" in that mini-table (`data`), which we don't use in the `SET ...` statements, but we do use in the `WHERE` clause (just like the `id` column).

chr1sjf0x (approved):
Cool, TY 👍

PR #197, opened 2021-11-30
Add ManyToOneReference.id setter

stephenh:


packages/integration-tests/src/entities/Book.test.ts
```diff
@@ -16,4 +16,35 @@ describe("Book", () => {
...
+    // Given an author with a loaded list of books
+    const a1 = em.create(Author, { firstName: "a1" });
+    const b1 = em.create(Book, { title: "b1", author: a1 });
+    // And the book is initially in the author's loaded collection
+    expect(a1.books.get).toEqual([b1]);
+    // When we make a new author and move to the book to it
+    const a2 = em.create(Author, { firstName: "a2" });
+    b1.author.set(a2);
+    // Then both a1 and a2 book collections are correct
+    expect(a1.books.get).toEqual([]);
```
stephenh, line 31: Without the fix in this PR, the `books` still had `b1` in it, even though it was supposed to have moved over to `a2`.

packages/orm/src/index.ts
```diff
@@ -141,17 +141,15 @@ interface Flavoring<FlavorT> {
 }
 export type Flavor<T, FlavorT> = T & Flavoring<FlavorT>;
 
-export function setField(entity: Entity, fieldName: string, newValue: any): boolean {
+export function setField<T extends Entity>(entity: T, fieldName: keyof T, newValue: any): boolean {
```
stephenh, line 144: Small typing improvement while here

packages/orm/src/relations/ManyToOneReference.ts
```diff
@@ -89,14 +93,23 @@ export class ManyToOneReference<T extends Entity, U extends Entity, N extends ne
...
   }
 
-  get id(): IdOf<U> | undefined {
+  get id(): IdOf<U> | N {
     ensureNotDeleted(this.entity, { ignore: "pending" });
-    return maybeResolveReferenceToId(this.current()) as IdOf<U> | undefined;
+    return maybeResolveReferenceToId(this.current()) as IdOf<U> | N;
+  }
+
+  set id(id: IdOf<U> | N) {
```
stephenh, line 101: This is new. Technically our `Reference.id` was not previously marked as `readonly`, but kinda surprising, TS didn't complain that we didn't have a setter :thinking: :shrug:

packages/orm/src/relations/ManyToOneReference.ts
```diff
@@ -149,40 +162,42 @@ export class ManyToOneReference<T extends Entity, U extends Entity, N extends ne
...
   // Internal method used by OneToManyCollection
   setImpl(other: U | N): void {
+    // If other is new (i.e. has no id), we only noop/early exit if it matches our loaded reference.
+    // Otherwise, noop/early exit based on id comparison (b/c we may not be loaded yet).
     if (other?.isNewEntity ? other === this.loaded : this.id === other?.id) {
       return;
     }
-
-    // we may not be loaded yet, but our previous entity might already be in the UoW
-    const previousLoaded = this.loaded ?? this.maybeFindExisting();
```
stephenh, line 164: Well shoot, I thought there was a bug here that `previousLoaded` was, on `main`, getting set only _after_ the field had already changed, but it looks like `main` was correct, and I just had some mis-refactored code on my `aggregate-roots` branch.

Hrm. Well, I think the code is still cleaner now, and we learned to have a `ManyToOneReference.id` setter, so going to keep the PR, but I guess not technically a bug fix.

PR #212, opened 2021-12-21
fix: Make o2o types safer.

stephenh:
Removes some of the unsafe-for-o2o methods out of `Reference` (i.e. `id`, `isSet`) and moves them into a new `ManyToOneReference` interface.

### DIFF
--- a/packages/integration-tests/src/relations/OneToOneReference.test.ts
+++ b/packages/integration-tests/src/relations/OneToOneReference.test.ts
@@ -100,14 +100,14 @@ describe("OneToOneReference", () => {
     expect(a1.image.get?.fileName).toEqual("f1");
   });
 
-  it("id fails if not loaded", async () => {
-    await insertAuthor({ first_name: "a1" });
-    const em = newEntityManager();
-    const a1 = await em.load(Author, "1");
-    expect(() => a1.image.id).toThrow("Author:1.image was not loaded");
-    await a1.image.load();
-    expect(a1.image.id).toBeUndefined();
-  });
+  // it("id fails if not loaded", async () => {
+  //   await insertAuthor({ first_name: "a1" });
+  //   const em = newEntityManager();
+  //   const a1 = await em.load(Author, "1");
+  //   expect(() => a1.image.id).toThrow("Author:1.image was not loaded");
+  //   await a1.image.load();
+  //   expect(a1.image.id).toBeUndefined();
+  // });
 
   it("can cascade delete", async () => {
     await insertAuthor({ first_name: "a1" });
@@ -115,7 +115,7 @@ describe("OneToOneReference", () => {
     const a1 = await em.load(Author, "1", "image");
     await insertImage({ type_id: 2, file_name: "f1", author_id: 1 });
     await em.refresh();
-    expect(a1.image.isSet).toBeTruthy();
+    // expect(a1.image.isSet).toBeTruthy();
     em.delete(a1);
     await em.flush();
     expect((await knex.select("*").from("images")).length).toEqual(0);
@@ -125,7 +125,7 @@ describe("OneToOneReference", () => {
     await insertAuthor({ first_name: "a1" });
     const em = newEntityManager();
     const a1 = await em.load(Author, "1", "image");
-    expect(a1.image.isSet).toBeFalsy();
+    // expect(a1.image.isSet).toBeFalsy();
     em.delete(a1);
     await em.flush();
     expect((await knex.select("*").from("authors")).length).toEqual(0);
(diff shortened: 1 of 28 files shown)


packages/orm/src/index.ts
```diff
@@ -319,7 +319,7 @@ export class ConfigApi<T extends Entity, C> {
     }
   }
 
-  beforeDelete<H extends LoadHint<T>>(populate: H, fn: HookFn<Loaded<T, H>, C>): void;
+  beforeDelete<H extends LoadHint<T>>(popuate: H, fn: HookFn<Loaded<T, H>, C>): void;
```
zgavin, line 322: not sure why this changed? `populate` is now mispelled

packages/orm/src/relations/Reference.ts
```diff
@@ -1,66 +1,60 @@
...
+import { Relation, RelationT, RelationU } from "./Relation";
 
-const H = Symbol();
+// Exported and used in sub-interfaces/types per https://stackoverflow.com/a/70437874/355031
+export const ReferenceN = Symbol();
+
+// We could go back to `N extends never | undefined` once https://github.com/microsoft/TypeScript/issues/47213
+// is resolved so that `MarkLoaded` can keep matching on `Reference` and it works with sub-interfaces.
+export type MaybeUndefined<T, N extends NullOrNotNull> = N extends "null" ? T | undefined : T;
+export type NullOrNotNull = "null" | "not-null";
```
zgavin, line 14:
```suggestion
export type Nullable = "nullable"
export type NotNullable = "not-nullable"
export type MaybeNullable = Nullable | NotNullable
```

then you could do just `MaybeNullable` instead of `NullOrNotNull`.  Less concise, but seems like it'd show intent better and not require hardcoded strings all over.

packages/orm/src/relations/Reference.ts
```diff
@@ -1,66 +1,60 @@
...
 
 /**
  * A many-to-one / foreign key from `T` to `U`, i.e. book to author.
  *
  * The `N` generic is for whether the field is optional (i.e. the foreign key column is
  * nullable). If it is optional, `N` will be `undefined`, which makes the return types
- * `U | undefined`. If it is not optional, `N` will be `never`, making the return types
- * `U | never` which becomes just `U`.
+ * `U | undefined`. If it is not optional, `N` will be `U`, making the return types
+ * `U | U` which becomes just `U`.
```
zgavin, line 22: comment needs more updates i think?

packages/integration-tests/src/relations/OneToOneReference.test.ts
```diff
@@ -100,22 +100,22 @@ describe("OneToOneReference", () => {
...
-    expect(a1.image.id).toBeUndefined();
-  });
+  // it("id fails if not loaded", async () => {
+  //   await insertAuthor({ first_name: "a1" });
+  //   const em = newEntityManager();
+  //   const a1 = await em.load(Author, "1");
+  //   expect(() => a1.image.id).toThrow("Author:1.image was not loaded");
+  //   await a1.image.load();
+  //   expect(a1.image.id).toBeUndefined();
+  // });
```
zgavin, line 110: delete this?

packages/integration-tests/src/relations/OneToOneReference.test.ts
```diff
@@ -100,22 +100,22 @@ describe("OneToOneReference", () => {
...
 
   it("can cascade delete", async () => {
     await insertAuthor({ first_name: "a1" });
     const em = newEntityManager();
     const a1 = await em.load(Author, "1", "image");
     await insertImage({ type_id: 2, file_name: "f1", author_id: 1 });
     await em.refresh();
-    expect(a1.image.isSet).toBeTruthy();
+    // expect(a1.image.isSet).toBeTruthy();
```
zgavin, line 118: `a1.image` seems like it should be `Loaded...` and this can be uncommented?

zgavin (approved):

packages/orm/src/index.ts
stephenh: Yep, fixed in #213

packages/integration-tests/src/relations/OneToOneReference.test.ts
stephenh: My intention was to restore this pre-merging; went ahead and did that in #213 . It required added a dedicated `LoadedOneToOneReference` that adds the `id` / `isSet` / etc. methods that are now only available on the unloaded `ManyToOneReference`.

packages/integration-tests/src/relations/OneToOneReference.test.ts
stephenh: Yep, restored in #213

stephenh:
#213 is turning out well enough that I'm going to close this one for that one... We can come back to the `"null" | "not-null"` approach later if/as needed.

PR #272, opened 2022-03-07
fix: Refactor factories to use null less.

### LINKED ISSUE #268: Refactor out the factory null marker
When we do:

```
newAuthor({ firstName: "a1", books: [{}] });
```

We currently "have" to have a single `setOpts` / constructor call and so end up doing:

```
em.create(Author, { firstName: "a1", books: [b1]);
```

The wrinkle is that we've created `b1` before the `Author`, but we tell the `Book` that it's okay to not have a parent by passing:

```
newBook({ author: null });
```

But really we could just split the Author creation up into two phases:

1. Phase one is all primitive fields, i.e. `em.create(Author, { firstName: "a1" })`
2. Phase two is creating the downstream entities i.e. with `newBook({ author: a1 })`
3. Phase three is then setting them into the author `author.set({ books: [b1] })`

We may not actually need the 3rd set if the two-way collection percolation gets the `author.books` collection correctly populated anyway.

The upshot of this is:

* Just an internal refactoring which would be nice, but also
* Factories like `newBook` would not have to observe the `author: null` marker which today actually happens _after_ the entity is supposedly created with all-required-fields set, i.e. we have a factory like:

```
  const hc = newTestInstance(em, HomeownerContract, {
    lineItems: [{}],
    ...opts,
  });
  // We break this out so that our `hc.projectStage.get` will be an actual,
  // post-factory-opts-fuzziness entity.
  if (opts.andSigned) {
    hc.document.set(
      newDocument(em, {
        documentType: DocumentType.PreconAgreement,
        project: hc.projectStage.get.project.id,
        uploaded: true,
      }),
    );
  }
  return hc;
```

But `hc.projectStage.get` may actually be the `null` marker.

stephenh:
When doing:

```
newAuthor(em, { firstName: "a1", books: [{}] });
```

We used to assume we needed to create the `books` before the `Author` existed, so that we could do a single:

```
em.create(Author, { firstName: ..., books: ... });
```

But since the `Book` itself requires an author, we'd end up passing `{ author: null }` to the Book factory/constructor, to tell it "we'll have an author for you soon".

But instead we can just split up the opts into "initial opts" that are actual columns on the entity, and then do the o2m/m2m collections later, i.e.:

```
// initial opts
const a = em.create(Author, { firstName: ... });
// additional opts, which can use the instantiated Author
... make newBook(em, { author: a });
a.setOpts({ books: [b1] });
```

And now the book factory / constructor doesn't see the `author: null` marker.

Fixes #268


packages/orm/src/newTestInstance.ts
```diff
@@ -181,16 +187,15 @@ function resolveFactoryOpt<T extends Entity>(
       }
       // Otherwise fall though to making a new entity via the factory
     }
-    // Find the opposite side, to see if it's an o2o or o2m pointing back at us
-    const otherField = field.otherMetadata().fields[field.otherFieldName];
+    if (maybeEntity === undefined) {
+      // If this is image.author (m2o) but the other-side is a o2o, pass null instead of []
+      maybeEntity = (field.otherMetadata().fields[field.otherFieldName].kind === "o2o" ? null : []) as any;
+    }
```
zgavin, line 193:
```suggestion
    if (maybeEntity === undefined) {
      // If this is image.author (m2o) but the other-side is a o2o, pass null instead of []
      maybeEntity = (field.otherMetadata().fields[field.otherFieldName].kind === "o2o" ? null : []) as any;
    }
```
```suggestion
    // If this is image.author (m2o) but the other-side is a o2o, pass null instead of []
    maybeEntity ??= (field.otherMetadata().fields[field.otherFieldName].kind === "o2o" ? null : []) as any;
```

zgavin (approved):

packages/orm/src/newTestInstance.ts
stephenh: Ah nice, updated

PR #281, opened 2022-03-26
fix: Fix native enums not working.

stephenh:
The columns weren't actually in the runtime metadata, so were effectively ignored.

For now just telling the runtime metadata that they're primitive columns.


mmahalwy:
oops 😓

stephenh:
@mmahalwy np! :-)

PR #286, opened 2022-03-30
Resolve #273 Added isLoaded, ensureLoaded, and ensureLoadedThen

### LINKED ISSUE #273: Add isLoaded(entity, typeHint)
For doing runtime checking of whether the entity does match the type hint, i.e. to branching in an overload method that takes a "maybe loaded/maybe not".

Should we also have a `ensureLoaded(entity, typeHint): Promise<Loaded<T, H>>` that is technically async that _might_ call `populate`, but first uses `isLoaded(entity, typeHint)` to short-circuit and avoid calling `populate`?

With the idea that, if `isLoaded` which is sync returned true, there would only be a single `await` bounce on the event loop, and not one for each entity/level of the populate tree.

zgavin:


packages/orm/src/loaded.ts
```diff
@@ -117,3 +117,38 @@ export type NestedLoadHint<T extends Entity> = {
     ? {}
     : never;
 };
+
+/** recursively checks if the relations from a load hint are loaded on an entity */
+export function isLoaded<T extends Entity, H extends LoadHint<T>>(entity: T, hint: H): entity is Loaded<T, H> {
```
stephenh, line 122: Nice! Really pretty short & sweet

packages/orm/src/loaded.ts
```diff
@@ -117,3 +117,38 @@ export type NestedLoadHint<T extends Entity> = {
...
+
+export function ensureLoaded<T extends Entity, H extends LoadHint<T>, R>(entity: T, hint: H): Promise<Loaded<T, H>> {
+  return isLoaded(entity, hint) ? entity : (entity as any).populate(hint);
+}
+
+export function ensureLoadedThen<T extends Entity, H extends LoadHint<T>, R>(
+  entity: T,
+  hint: H,
+  fn: (loaded: Loaded<T, H>) => R,
+): R | Promise<R> {
```
stephenh, line 152: I was kind of expecting this to just be `Promise<R>` but don't have a clear reason/articulation for why it should/shouldn't be :shrug:

stephenh (approved):
:100:

packages/orm/src/loaded.ts
zgavin: the point of `ensureLoadedThen` is to be a convenience function for our methods that return `MaybePromise` and have async and sync overloads

PR #289, opened 2022-04-03
feat: Loosen restrictions on timestamp columns.

### LINKED ISSUE #288: Loosen restriction on entity tables

stephenh:
This PR:

* Changes the default behavior for `isEntityTable` to (with the default config) be any table with an `id` column (that is not also a code table)
* Adds a `joist-codegen.json` `timestampColumns` config for projects (like ours) to explicitly opt in to either differently-named timestamp columns, or to make timestamp columns required for entity tables
* Adds a new `schema-misc` project for testing tables w/different conventions, i.e. an `authors` table with `createdAt` and `updatedAt` columns (note not snake case) and a `book` table with no timestamp columns at all

Fixes #288

### DIFF
--- a/docs/docs/features/fast-database-resets.md
+++ b/docs/docs/features/fast-database-resets.md
@@ -9,7 +9,7 @@ To reset the database between each unit test, Joist generates a `flush_database`
 await knex.select(knex.raw("flush_database()"));
 ```
 
-This is generated at the end of the `joist-migation-utils` set only if `ADD_FLUSH_DATABASE` environment variable is set, i.e. this function should never exist in your production database. It is only for local testing.
+This is generated at the end of the `joist-codegen`, which should only be invoked against local development databases, i.e. this function should never exist in your production database. It is only for local testing.
 
 ### What About Per-Test Transactions?
 
--- a/packages/codegen/src/EntityDbMetadata.ts
+++ b/packages/codegen/src/EntityDbMetadata.ts
@@ -4,6 +4,7 @@ import { plural, singular } from "pluralize";
 import { imp, Import } from "ts-poet";
 import {
   Config,
+  getTimestampConfig,
   isAsyncDerived,
   isDerived,
   isFieldIgnored,
@@ -45,6 +46,7 @@ export type DatabaseColumnType =
   | "character varying"
   | "varchar"
   | "timestamp with time zone"
+  | "timestamp without time zone"
   | "date"
   | "jsonb";
 
@@ -168,6 +170,8 @@ export class EntityDbMetadata {
   polymorphics: PolymorphicField[];
   tableName: string;
   tagName: string;
+  createdAt: PrimitiveField | undefined;
+  updatedAt: PrimitiveField | undefined;
 
   constructor(config: Config, table: Table, enums: EnumMetadata = {}) {
     this.entity = makeEntity(tableToEntityName(config, table));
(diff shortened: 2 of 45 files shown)


docs/docs/features/fast-database-resets.md
```diff
@@ -9,7 +9,7 @@ To reset the database between each unit test, Joist generates a `flush_database`
 await knex.select(knex.raw("flush_database()"));
 ```
 
-This is generated at the end of the `joist-migation-utils` set only if `ADD_FLUSH_DATABASE` environment variable is set, i.e. this function should never exist in your production database. It is only for local testing.
+This is generated at the end of the `joist-codegen`, which should only be invoked against local development databases, i.e. this function should never exist in your production database. It is only for local testing.
```
stephenh, line 12: I moved `flush_database` to be created as the last step of codegen, instead of the last step of migrations.

The primary reason is that previously we'd copy/pasted the `isEntityTable` code between `joist-codegen` and `joist-migration-utils`, and having different definitions of `isEntityTable` for each of those is bad.

I could have moved `isEntityTable` into `joist=utils` or something, but dunno, actually kind of like running it as the last step of codegen, b/c it's already something that is "something you'd only run locally", and also it means we can easily re-use the pg-structure-driven metadata, instead of having to load it once for the migrations (that only used it for creating the flush function) and once for codegen (which of course just inherently needs it).

packages/codegen/src/EntityDbMetadata.ts
```diff
@@ -241,6 +245,10 @@ export class EntityDbMetadata {
 
     this.tableName = table.name;
     this.tagName = config.entities[this.entity.name]?.tag;
+
+    const { createdAtConf, updatedAtConf } = getTimestampConfig(config);
+    this.createdAt = this.primitives.find((f) => createdAtConf.names.includes(f.columnName));
+    this.updatedAt = this.primitives.find((f) => updatedAtConf.names.includes(f.columnName));
```
stephenh, line 251: Baking into the metadata what, if any, the created at / updated at field names are for each entity.

In our app, these will all be the same; I toyed with detecting "literally all of the entities use the same updated at / created at" and then "putting that config in just one spot", instead of putting it in every metadata instance, but, dunno, I couldn't find a great spot for it.

And this doesn't seem that bad.

packages/codegen/src/generateFlushDatabaseFunction.ts
```diff
@@ -0,0 +1,30 @@
+import { Client } from "pg";
+import { Db } from "pg-structure";
+import { Config } from "./config";
+import { isEntityTable, isJoinTable } from "./utils";
+
+/** Creates a `flush_database` stored procedure to truncate all of the tables between tests. */
+export async function createFlushDbFunction(db: Db, client: Client, config: Config): Promise<void> {
```
stephenh, line 7: Nothing changed, just moved in from `joist-migration-utils`.

Actually another reason this makes sense is b/c users who bring-their-own-migration library (so don't need/want `joist-migration-utils`) would still want to use `flush_database`...

packages/tests/schema-misc/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "version": "1.0.0",
+  "license": "MIT",
+  "private": true,
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/stephenh/joist-ts.git",
+    "directory": "packages/tests/schema-misc"
+  },
+  "scripts": {
+    "migrate": "DATABASE_URL=postgres://joist:local@localhost:5435/schema_misc node-pg-migrate up --decamelize=false --migrations-dir=./migrations",
```
stephenh, line 12: I'm running `node-pg-migrate` directly to turn off `decamelize` so that we can test camel-cased names in the db.

PR #291, opened 2022-04-06
feat: [SC-14739] Improved newTestInstance support for polymorphic references

chr1sjf0x:
Adds a `maybeNewPoly` helper function to make working with polymorphic references easier.

image


packages/orm/src/newTestInstance.ts
```diff
@@ -308,8 +336,34 @@ export function maybeNew<T extends Entity>(opts?: ActualFactoryOpts<T>): Factory
...
+ *     parent: maybeNewPoly<FavoriteThingParent>(Author, {}, Book, Publisher),
+ *     ...opts,
+ *   });
+ * }
+ * ```
+ */
+export function maybeNewPoly<T extends Entity>(
+  cstr: EntityConstructor<T>,
+  opts?: ActualFactoryOpts<T>,
+  ...orExisting: EntityConstructor<T>[]
```
chr1sjf0x, line 359: thought: I'm not married to this signature, and would be open to suggestions.

In particular, it combines the concepts of `this is what I want to make` and `it is also the default thing to look for first`, which probably makes sense in most cases 🤷 

But also, it would be nice if the opts were forced down to be the actual opts of the constructor you specify.

packages/integration-tests/migrations/1580658856631_author.ts
```diff
@@ -158,4 +158,17 @@ export function up(b: MigrationBuilder): void {
     parent_book_review_id: foreignKey("book_reviews", { notNull: false, unique: true }),
     text: "text",
   });
+
+  // testing for a required polymorphic field
```
stephenh, line 162: I get this is "a favorite thing can be owned by book/publisher/author"...what's the `author_favorites` for? Turning it into a m2m?

chr1sjf0x: I've revised this PR to not add any new entities, per my comment [here](https://github.com/stephenh/joist-ts/pull/291/files#r844309866). I feel like this muddied the water unnecessarily.

packages/orm/src/newTestInstance.ts
stephenh: Does throwing in another generic help:

```
export function maybeNewPoly<T extends Entity, F extends T>(
  cstr: EntityConstructor<F>,
  opts?: ActualFactoryOpts<F>,
  ...orExisting: EntityConstructor<T>[]
```

Not sure if TS is going to narrow `opts` to be explicitly the union type from cstr...

stephenh: Just thinking out loud, the "array of constructors on the end" seems a little opaque, wdyt of things like:

```
maybeNewPoly<Parent>({
  order: [Foo, Bar, Zaz],
  ifNew: [Bar, { name: "bar props" }]
]);
```

Or maybe that's too much, and:


```
maybeNewPoly<Parent>(
  Bar, { name: "bar props" },
  { order: [Foo, Bar, Zaz] }
]);
```

Which is the existing signature, but opts-izing the last arg solely for readability purposes...wdyt?

stephenh (approved):
Nice! Great tests...

packages/integration-tests/migrations/1580658856631_author.ts
```diff
@@ -158,4 +158,17 @@ export function up(b: MigrationBuilder): void {
...
+  createEntityTable(b, "favorite_things", {
+    parent_book_id: foreignKey("books", { notNull: false }),
+    parent_publisher_id: foreignKey("publishers", { notNull: false }),
+    parent_author_id: foreignKey("authors", { notNull: false }),
+  });
+
+  // testing for a required polymorphic field
```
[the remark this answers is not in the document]
stephenh: Thanks for doubling back on this and avoiding new entities! I've done that several times myself, start out adding new entities and then realizing "ah damn, I could just use...". :-)

packages/orm/src/newTestInstance.ts
chr1sjf0x: Yea, I had fool around with adding extra generic types, but it didn't help and made it a bit annoying since one would have to specify the 2 parameters, it wouldn't infer it.

What about this signature:
```
export function maybeNewPoly<T extends Entity>(
  ifNewCstr: EntityConstructor<T>,
  opts?: { ifNewOpts: ActualFactoryOpts<T>; existingOrderSearch: EntityConstructor<T>[] },
): FactoryEntityOpt<T> {
  // Return a marker that resolveFactoryOpt will look for
  return new MaybeNew<T>((opts?.ifNewOpts || {}) as any, opts?.existingOrderSearch ?? [ifNewCstr]) as any;
}
```

Then you could do things like:
```
maybeNewPoly(Activity); 
// Which would be equivalent to 
maybeNewPoly(Activity, { ifNewOpts: {}, existingSearchOrder: [Activity] });
```

chr1sjf0x: Her's the 2 type param version, which at least gives folks the option for better typing 🤷 
image

chr1sjf0x: I'm kind of liking the 2 generic version now more.

@stephenh, if you agree, I'll update the rest of the PR
image

chr1sjf0x: Updated, TY for the feedback @stephenh 👍

PR #292, opened 2022-04-07
fix: Correct semantic release publishCmd

chr1sjf0x:
Update the command to correctly publish to npm.


stephenh (approved):
:crossed_fingers: thanks!

PR #304, opened 2022-04-18
fix: Use joist util fail

khalidwilliams:
This PR brings the `fail` utils function in to `keys.ts` to prevent a naming conflict with `jest`'s `fail()` function.


stephenh (approved):
Thanks!

PR #307, opened 2022-04-26
feat: Implement DeepNew for factories and tests

### LINKED ISSUE #255: Allow tests to have always loaded entities
In tests, it would be a large DX win to be able to traverse the object graph w/all relations already loaded.

I.e. we could have factories return `DeepNew<...>` versions of entities that just assumed all relations are always loaded.

This would require either:

1) the entire DB would already be loaded into the `EntityManager` (which in tests, the data set is usually small enough for this to be doable), or

2) the `DeepNew<...>` would be just a heuristic that is "usually true", but might still fail at runtime if the test/code-under-test did not actually trigger relation to be loaded. For tests that use a single `EntityManager`, this heuristic would likely always be true. For tests that use multiple `EntityManager`s, i.e. resolver tests, it would probably be "usually true", but not 100%.

---

Just talking out loud, we could envision a `EntityManager.loadEntireDatabase` that just pulled the entire db into the memory. A naive implementation, that just did one `SELECT` per table + one `SELECT` per o2m/m2m/etc. relation, seems like it'd be incredibly inefficient and not doable.

A fancy implementation idea would be to leverage the `flush_database` trick and use a `select_all_tables` stored procedure that returned a `SELECT *` across all tables as a `UNION`. This would give us a single wire call + result set with "all the things", which would be a huge mess, but we could have a dedicated/internal method that knew how to decipher the mega-result set into a) individual entities and b) to force/preload all of the o2m/m2m/o2o collections appropriately.

stephenh:
Fixes #255 by:

* Adding a `DeepNew<T>` type that marks all relations as recursively loaded
* Add an `EntityManager.refresh({ deepLoad?: true })` that will crawl existing entities and recursively fully load any relation/entity we find
* Updating factories to return a `DeepNew<T>` version of their entity

This allows tests to use a `DeepNew` and, as long as they're either:

a) Working within only a single test EM where all entities are new and defacto fully loaded, or
b) Using `em.refresh({ deepLoad: true })` either directly or inside a `run`-style utility method

Then they can have sync access to any relation within an entity that is reachable.

I suppose as a downside, doing `refresh({ deepLoad: true })` _all the time_ in a widely-used utility method like `run` will result in more work being done to fully load literally any reachable relation, vs. previously relations merely within the UoW. But I don't think it should result in that much more work.

### DIFF
--- a/packages/codegen/src/generateEntityCodegenFile.ts
+++ b/packages/codegen/src/generateEntityCodegenFile.ts
@@ -410,8 +410,8 @@ export function generateEntityCodegenFile(config: Config, meta: EntityDbMetadata
       populate<H extends ${LoadHint}<${entityName}>>(opts: { hint: H, forceReload?: boolean }): Promise<${Loaded}<${entityName}, H>>;
       populate<H extends ${LoadHint}<${entityName}>, V>(hint: H, fn: (${tagName}: Loaded<${entityName}, H>) => V): Promise<V>;
       populate<H extends ${LoadHint}<${entityName}>, V>(opts: { hint: H, forceReload?: boolean }, fn: (${tagName}: Loaded<${entityName}, H>) => V): Promise<V>;
-      populate<H extends ${LoadHint}<${entityName}>, V>(hint: H, fn?: (${tagName}: Loaded<${entityName}, H>) => V): Promise<${Loaded}<${entityName}, H> | V> {
-        return this.em.populate(this as any as ${entityName}, hint as any, fn);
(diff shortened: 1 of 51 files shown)


packages/codegen/src/generateEntityCodegenFile.ts
```diff
@@ -407,9 +407,11 @@ export function generateEntityCodegenFile(config: Config, meta: EntityDbMetadata
       }
 
       populate<H extends ${LoadHint}<${entityName}>>(hint: H): Promise<${Loaded}<${entityName}, H>>;
+      populate<H extends ${LoadHint}<${entityName}>>(opts: { hint: H, forceReload?: boolean }): Promise<${Loaded}<${entityName}, H>>;
```
stephenh, line 410: Adding overloads so that `CustomCollection`s can pass in `forceReload` opt if they need/want

PR #312, opened 2022-05-01
fix: add numeric types

mmahalwy:


stephenh:
@mmahalwy this is great, definitely something we need to flush out / fully support.

I'd like to have some `integration-test` coverage of these, otherwise it's very likely I'll unintentionally break them while refactoring at some point. :-)

Can you add a new table to the `integration-test` migration, something like `author_stats` that just has a bunch of numeric columns? They don't even need realistic names, like column `a` is `smallint`, column `b` is `integer`, column `c` is `bigint`, etc., just go down the list making a column "next letter of the alphabet" --> "next numeric type you want to test".

And then have like an `AuthorStats.test.ts` that creates a new `AuthorStat` entity with like `a: 1`, `b: 2`, `c: 3` dummy values for each field, and make sure the entity can save & load?

mmahalwy:
@stephenh i made some of those changes but not sure why im getting error with `makeRunResolver`. The code shows that it should generate it but i dont see it with other tests

stephenh:
@mmahalwy shoot! Sorry about that; we recently refactored how our GraphQL scaffolding works, but b/c no one had added a new entity yet, it didn't catch that our super-hand-coded / should-be-codegen-d GraphQL types were out of sync.

Hope it's okay, but I pushed up a commit that I _think_ brings everything up to speed.

stephenh (approved):
Looks great, thank you!

PR #315, opened 2022-05-04
feat: add ability to disable tags

mmahalwy:


packages/integration-tests/src/entities/metadata.ts
```diff
@@ -14,9 +14,10 @@ export const authorMeta: EntityMetadata<Author> = {
   type: "Author",
   idType: "int",
   tagName: "a",
+  disableTags: false,
   tableName: "authors",
   fields: {
-    id: { kind: "primaryKey", fieldName: "id", fieldIdName: undefined, required: true, serde: new KeySerde("a", "id", "id", "int") },
+    id: { kind: "primaryKey", fieldName: "id", fieldIdName: undefined, required: true, serde: new KeySerde("a", "id", "id", "int", false) },
```
mmahalwy, line 20: imho, i recommend we change `KeySerde` to accept one argument that's an object

packages/integration-tests/src/entities/metadata.ts
```diff
@@ -14,9 +14,10 @@ export const authorMeta: EntityMetadata<Author> = {
   type: "Author",
   idType: "int",
   tagName: "a",
+  disableTags: false,
```
mmahalwy, line 17: while the config is for all, we can scope the option to be on a table level. lmk what you think

packages/codegen/src/EntityDbMetadata.ts
```diff
@@ -170,6 +170,7 @@ export class EntityDbMetadata {
   polymorphics: PolymorphicField[];
   tableName: string;
   tagName: string;
+  disableTags: boolean;
```
stephenh, line 173: Instead of introducing `disableTags`, maybe make `tagName: string | undefined`, and we can treat `tagName: undefined` as "don't use a tag".

packages/codegen/src/config.ts
```diff
@@ -56,6 +56,7 @@ export interface Config {
   codegenPlugins: string[];
   entities: Record<string, EntityConfig>;
   ignoredTables?: string[];
+  disableTags?: boolean;
```
stephenh, line 59: Maybe `taggedIds?: boolean` and we'll default it to true?

packages/orm/src/keys.ts
```diff
@@ -6,7 +6,7 @@ const tagDelimiter = ":";
 
 // I'm not entirely sure this is still necessary, but use a small subset of EntityMetadata so
 // that this file doesn't have to import the type and potentially create import cycles.
-type HasTagName = { tagName: string; idType: "int" | "uuid" };
+type HasTagName = { tagName: string; idType: "int" | "uuid"; disableTags: boolean };
```
stephenh, line 9: Same suggestion, maybe just name `tagName: string | undefined`?

stephenh (commented):
Cool, looks like a great start, thanks! Just a few suggestions on how to pass things through.

I'm a little said because tagged ids are _such a great idea_! that it'd be great to just always use them, but I totally get for doing migrations/just other codebases, skipping tags would be required for adoption.

Maybe pick an entity in `integration-tests` to set the `tagName: undefined`? Like maybe the new `AuthorStat` entity, or really any entity that isn't from the core Author/Book/BookReview/Publisher entities that already have a lot of tests.

Then we can make sure loading/saving w/o the tags works fine.

mmahalwy:
Agreed with the `tagName: string | undefined`. I wanted to do that to begin but wasnt sure. I like it better cause it allows the user to pick and choose which models (maybe there's an opportunity for rolling adoption)

mmahalwy:
@stephenh mighjt need your help here. getting stuck at `assertIdsAreTagged`

packages/orm/src/EntityManager.ts
```diff
@@ -180,8 +180,8 @@ export class EntityManager<C = {}> {
   }
 
   /** Looks up `id` in the list of already-loaded entities. */
-  getEntity<T extends Entity>(id: IdOf<T>): T | undefined {
-    return this._entityIndex.get(id) as T | undefined;
+  getEntity<T extends Entity>(tableName: string, id: IdOf<T>): T | undefined {
```
stephenh, line 183: Our codebase uses this method a lot, and just in general I think/hope that eventually most Joist apps will use tagged ids, and so should be allowed to use this method.

Can we put it back, and then add an `assertTaggedId`-style check that will fail at runtime if its called w/an untagged id?

packages/orm/src/EntityManager.ts
```diff
@@ -180,8 +180,8 @@ export class EntityManager<C = {}> {
...
+    return this._entityIndex.get(`${tableName}:${id}`) as T | undefined;
```
stephenh, line 184: Instead of using the `tableName`, I'm thinking we could try something like:

* Add a `BaseEntity.entityId` getter
* If an entity is using tagged ids, `entityId` returns the tagged id as-is
* If an entity is not using tagged ids, `entityId` returns `${constructor.name}:${id}` i.e. `AuthorStat:1`

Then internally in the places we had been using `entity.id` with the assumption it was tagged, we could use `entity.entityId`.

Maybe the only con is that historically we've been cautious about adding too many extra methods onto entity's, b/c typically that is something that purists hate on ORMs about, but :shrug: IMO I could see `entity.entityId` being useful for apps themselves to use, and it seems very unlikely to conflict with columns the user is storing in the database, so should be fine.

packages/orm/src/EntityManager.ts
```diff
@@ -426,7 +426,8 @@ export class EntityManager<C = {}> {
     }
     const meta = getMetadata(type);
     const tagged = tagId(meta, id);
-    const entity = this.findExistingInstance<T>(tagged) || (await loadDataLoader(this, meta).load(tagged));
+    const entity =
+      this.findExistingInstance<T>(meta.tableName, tagged) || (await loadDataLoader(this, meta).load(tagged));
```
stephenh, line 430:
```
const tagged = entityId(meta, id);
this.findExistingInstance(tagged) || ...load(tagged);
```

packages/orm/src/EntityManager.ts
```diff
@@ -833,7 +835,7 @@ export class EntityManager<C = {}> {
     const meta = getMetadata(type);
     const id = keyToString(meta, row["id"]) || fail("No id column was available");
     // See if this is already in our UoW
-    let entity = this.findExistingInstance(id) as T;
+    let entity = this.findExistingInstance(meta.tableName, id) as T;
```
stephenh, line 838: So like for this call-site we could do something like:

```
const meta = getMetadata(type);
const id = keyToString(...);
const eid = entityId(meta, id); // returns `a:1` or `AuthorStat:1`
this.findExistingInstance(eid)
```

packages/orm/src/EntityManager.ts
stephenh: ...a variation on ^ would be that we keep a lot of the tagged id code as-is, so like `AuthorStat` would still get an `as` tag, and `tagName` would go back to being just `string` and not `string | undefined`, but then:

* We allow the user to control whether `someEntity.id` returns the tagged id or the normal id (and that's it)
* We add a `BaseEntity.taggedId` that always returns the tagged id
* Internally we move most callers of `entity.id` that today are assuming "it will always be a tagged id" to use `entity.taggedId`.

I think I like this alternative.... It is similar to the `entity.entityId` idea, of having a way to "always get a unique-per-entity id", but simplifies it by letting that always be a tagged id, even if the user "doesn't want tagged ids" returned from their `.id` method.

It also would mean most of the config / assign tag changes in this PR could go away...

And on the config side we've have something new like:

```
{ id: "tagged" | "untagged" }
```

In the `joist-codegen.json`. And you could either set it at the top-level for all entities, or if you wanted on an entity-by-entity base...

@mmahalwy dunno, do you mind maybe keeping this PR as-is and just doing a brand new branch that spikes this approach?

We could probably do something in `generateEntityCodegenFile` that, when generating the `.id` field, looks at the config `id === tagged / untagged` and output an extra `deTag` to quickly strip the prefix before returning it to the user.

Wdyt?

mmahalwy: @stephenh that works for me. We can do that.

So to sum it up, the idea is:
1. If you want untagged, it will wrap the id getter with the detag function
2. we will create another getting for getting the tagged id, using it regardless if choosing tagged or untagged

I do think we should have this top level, not per model.

stephenh: Yep! 

> I do think we should have this top level, not per model.

Yeah, that sounds good to me, I agree having it change per-model would be pretty weird.

We might have to find a few more places to unwrap (I'm thinking like `ManyToOneReference.id`?) but `Entity.id` is surely the big one that most users would want to have change.

stephenh: Also if it's only top-level, we'll need a new test project, similar to uuid-ids with ~one-two entities and id: untagged set.

mmahalwy: feels like we should consider nesting into a `testing-apps` instead of having multiple test projects under `packages`?

mmahalwy: yep thats what i mean @stephenh

PR #319, opened 2022-05-06
fix: Make BaseEntity.toJSON more Prisma-like.

### LINKED ISSUE #314: API returning entities does not include null fields
Take example of joist vs prisma:
```ts
app.get('/testing', async (req, res) => {
  const users = await em.find(User, {});

  res.send(users);
});
app.get('/testing2', async (req, res) => {
  const users = await prisma.user.findMany();

  res.send(users);
});
```

Because we are using `res.send`, we aren't calling each field directly. This causes fields that are `null` to be missing. See:
joist:
```json
[
  {
    "role": "ADMIN",
    "id": "u:0c1b33c7-897a-4ca3-a116-9560d02ff48e",
    "createdAt": "2022-04-14T03:14:53.828Z",
    "updatedAt": "2022-04-14T03:14:53.829Z",
    "firstName": "Costas",
    "lastName": "Hadjipateras",
    "email": "costas@withshepherd.com",
    "phoneNumber": "(203) 246 2006",
    "imageUrl": "",
    "title": "Head of Underwriting"
  }
]
```

Prisma:
```json
[
  {
    "id": "a7635524-423b-4324-b88d-1544c2dd31e3",
    "firstName": "Steve",
    "lastName": "Buonpane",
    "email": "steve@withshepherd.com",
    "phoneNumber": "(631) 513 5253",
    "imageUrl": "",
    "role": "ADMIN",
    "title": "Chief Insurance Officer",
    "createdAt": "2022-04-13T23:14:53.828Z",
    "updatedAt": "2022-04-13T23:14:53.829Z",
    "lastLoggedIn": null,
    "slackUserId": null
  }
]
```

This is obviously not a problem with GraphQL. 

A potential fix is having a helper function that turns an entity into an object.

stephenh:
Fixes #314

### DIFF
--- a/packages/integration-tests/src/EntityManager.test.ts
+++ b/packages/integration-tests/src/EntityManager.test.ts
@@ -926,10 +926,25 @@ describe("EntityManager", () => {
     const em = newEntityManager();
     const a1 = await em.load(Author, "1");
     a1.publisher.set(em.create(Publisher, { name: "p1" }));
-    expect(a1.toJSON()).toMatchObject({
+    await em.flush();
+    expect(a1.toJSON()).toEqual({
       id: "a:1",
+      address: null,
+      age: null,
+      createdAt: expect.anything(),
+      currentDraftBook: null,
+      favoriteColors: [],
+      favoriteShape: null,
       firstName: "a1",
-      publisher: "Publisher#1",
+      graduated: null,
+      initials: "a",
+      isPopular: null,
+      lastName: null,
+      mentor: null,
+      numberOfBooks: null,
+      publisher: "p:1",
+      updatedAt: expect.anything(),
+      wasEverPopular: null,
     });
   });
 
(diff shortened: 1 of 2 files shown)


packages/integration-tests/src/EntityManager.test.ts
```diff
@@ -926,10 +926,25 @@ describe("EntityManager", () => {
     const em = newEntityManager();
     const a1 = await em.load(Author, "1");
     a1.publisher.set(em.create(Publisher, { name: "p1" }));
-    expect(a1.toJSON()).toMatchObject({
+    await em.flush();
+    expect(a1.toJSON()).toEqual({
```
stephenh, line 930: @mmahalwy does this look good?

mmahalwy: yep!

mmahalwy (approved):
ty!

PR #322, opened 2022-05-13
feat: detag ids with config

mmahalwy:


mmahalwy:
wip but looking for initial feedback

stephenh (approved):

stephenh:
@mmahalwy this looks like a good start! Going to merge and see how it goes.

Fwiw for `id` --> `taggedId` fixes going forward, would like to have regression tests for as much of those as possible in the `untagged-ids` project otherwise I'm almost certain we'll have regression as we tweak/refactor our id approach going forward (i.e. to support numbers).

I thought a little bit about "should we duplicate the _entire_ / existing integration-tests suite"? Which I don't think is the right answer, so I think for now just copy/pasting/re-creating a subset of tests that we know are actually needed to coverage untagged id behavior seems like the most pragmatic / albeit tedious approach.

I'm also flip-flopping on the name `taggedId` --> `idTagged`, b/c we had an existing `idUntagged`, and so `idTagged` lines up with that. So I went ahead and pushed that to the PR before merging...

So :shrug: I guess try this out and let us know how it goes. Thanks!

mmahalwy:
thanks @stephenh for bringing this across the finish line!

PR #332, opened 2022-05-24
fix: untagged association

mmahalwy:


packages/orm/src/relations/OneToOneReference.ts
```diff
@@ -124,8 +124,8 @@ export class OneToOneReferenceImpl<T extends Entity, U extends Entity>
   async load(opts: { withDeleted?: boolean; forceReload?: boolean } = {}): Promise<U | undefined> {
     ensureNotDeleted(this.entity, { ignore: "pending" });
     if (!this._isLoaded || opts.forceReload) {
-      if (!this.entity.isNewEntity) {
-        this.loaded = await oneToOneDataLoader(this.entity.em, this).load(this.entity.idOrFail);
+      if (!this.entity.isNewEntity && this.entity.idTagged) {
+        this.loaded = await oneToOneDataLoader(this.entity.em, this).load(this.entity.idTagged);
```
mmahalwy, line 128: @stephenh can i get thoughts on this

stephenh: Hm, @mmahalwy yeah, looks like we should add an `idTaggedOrFail`.

I was futzing with adding a type-guard is `isNewEntity`, i.e. within an `if (isNewEntity)` check, ideally we could narrow the type of `idTagged` to just string, but it looks like TypeScript doesn't support type guards on getters, which is odd... :shrug: 

I suppose we could also do an `idTagged!`, with the hope that eventually TS will let us do the type guard approach, but dunno, I think probably `idTaggedOrFail` is best for now...

mmahalwy: im cool with `idTaggedOrFail`. Should be an easy add. LMK

stephenh: Cool, yep, let's do that. Thanks!

mmahalwy: ill add it to this PR

mmahalwy:
@stephenh updated

stephenh:
@mmahalwy this looks great! I think there is a build error where `BaseEntity` needs to re-declare that it (abstractly) implements `idTaggedOrFail`...seems kinda odd/not really sure why, but I've seen it before when adding methods to `Entity`.

stephenh:
@mmahalwy looks like the code gen files have been updated for the `tests/*` projects, but maybe not the main `integration-tests` project?

Fwiw I believe a top-level `yarn migrate` + `yarn codegen` should run it for all 4 projects.

mmahalwy:
@stephenh fixed

packages/tests/untagged-ids/src/entities/Author.test.ts
```diff
@@ -26,4 +26,17 @@ describe("Author", () => {
...
   });
+
+  it("can load author and books", async () => {
+    const em = newEntityManager();
+    const a1 = newAuthor(em);
+    const a2 = newAuthor(em);
+    const b1 = newBook(em, { author: a1 });
+    const b2 = newBook(em, { author: a2 });
+    await em.flush();
+    await em.refresh();
```
stephenh, line 37: I'm going to go ahead and merge the PR, to get this released :rocket: , but fwiw the `em.refresh()`s are probably unnecessary here...

Typically `refresh` is only necessary if some other database connection/transaction/em has modified the data that `em` had already loaded, and it now wants to reload.

90% of our usage of `refresh` is in test cases to do things like:

* Start one `em1` to setup test data
* Create a 2nd "clean" `em2` and invoke the endpoint-under-test (this ensure the endpoint's production code is not someone thrown off by `em1`-s state)
* Let the endpoint-under-test do an `em2.flush`
* Back in the test, run `em1.refresh` so that it can see whatever mutations the endpoint made, and then do assertions like "ah yeah, the author name changed from fred to bob".

If the idea of the tests was to have the `await em.find(Author, ...)` have more of a clean slate, then just creating a 2nd/net-new em is probably the best way to do that.

stephenh (approved):

packages/tests/untagged-ids/src/entities/Author.test.ts
mmahalwy: yeah i the intent was a new clean slate. thanks for the heads up

PR #338, opened 2022-05-26
fix: Fix cloning polymorphic references.

### LINKED ISSUE #333: Add em.clone support for polymorphic associations
Issue: We want to add `em.clone` support for polymorphic associations 

Example: We want to clone a `ScheduleTemplate` and all of its `units/activities/tasks` however this causes an issue because the clone is not getting the correct entity constructor for the polymorphic parent relationship on `Schedule`.

We want to be able to do this:
Code Example: `em.clone(scheduleTemplate, {units: {activities: {tasks: {})`

stephenh:
Fixes #333

### DIFF
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -1,4 +1,5 @@
 module.exports = {
   trailingComma: "all",
   printWidth: 120,
+  plugins: ["prettier-plugin-organize-imports"],
 };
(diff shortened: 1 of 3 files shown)


chr1sjf0x (approved):
Looks great, TY 🍣

PR #346, opened 2022-06-01
feat: support for camel cased table names and columns

mmahalwy:


.circleci/config.yml
```diff
@@ -48,6 +48,9 @@ jobs:
             cd ../untagged-ids
             ./run.sh ../../migration-utils/build/index.js
             ./run.sh ../../codegen/build/index.js
+            cd ../camel-cased
```
stephenh, line 51: I'm pretty good with a new test directory/db; fwiw just talking out loud, I wonder if we could add a table or two to the existing `schema-misc` that had a few camel cased names?

stephenh:
Looks like this existing `schema-misc` test is failing:

```
➤ YN0000:       Object {
➤ YN0000:     -   "authorId": Anything,
➤ YN0000:     +   "author": Object {
➤ YN0000:     +     "createdAt": 2022-06-01T04:16:45.256Z,
➤ YN0000:     +     "firstName": "firstName",
➤ YN0000:     +     "id": null,
➤ YN0000:     +     "lastName": null,
➤ YN0000:     +     "updatedAt": 2022-06-01T04:16:45.256Z,
➤ YN0000:     +   },
➤ YN0000:         "title": "title",
➤ YN0000:       }
➤ YN0000: 
➤ YN0000:        8 |     const b = newBook(em);
➤ YN0000:        9 |     // Then we did not set any updatedAt/createdAt values
➤ YN0000:     > 10 |     expect(b.__orm.data).toEqual({
➤ YN0000:          |                          ^
➤ YN0000:       11 |       authorId: expect.anything(),
➤ YN0000:       12 |       title: "title",
➤ YN0000:       13 |     });
```

Oh, but it's probably a good thing, b/c I think this PR actually fixes the `__orm.data.authorId` key to now be `__orm.data.author`. Nice!

stephenh (approved):
See comment musing if we should reuse `schema-misc` for this, but lgtm! Thanks!

.circleci/config.yml
stephenh: Yep, that's what I was thinking. Granted, we don't have a great written-down heuristic for "when / when not to create a new testing directory"...kind of figuring it out as we go. :-)

mmahalwy:
@stephenh updated

stephenh:
@mmahalwy nice! Just looks like that one `Book.test.ts` is still failing.

stephenh:
@mmahalwy ack! One more failure, later in the same test I think. :-)

mmahalwy:
@stephenh hopefully this does it 😓

stephenh:
There we go! :-D Merging! Thanks!

PR #349, opened 2022-06-03
feat: Add field-level immutability to the metadata.

stephenh:
In theory this will let us short-circuit field-level validation rules,
should those exist in the future.


packages/integration-tests/src/entities/Author.ts
```diff
@@ -45,7 +45,7 @@ export class Author extends AuthorCodegen {
   public afterCommitRan = false;
   public afterCommitIdIsSet = false;
   public afterCommitIsNewEntity = false;
-  public ageForBeforeFlush?: number;
+  public setGradatedInFlush?: boolean;
```
chr1sjf0x, line 48: quibble: 
```suggestion
  public setGraduatedInFlush?: boolean;
```

packages/integration-tests/src/entities/Author.ts
```diff
@@ -122,15 +122,15 @@ config.addRule("books", async (a) => {
 });
 
 // Example of cannotBeUpdated
-config.addRule(cannotBeUpdated("age", (a) => !!a.age && a.age < 100));
+config.addRule(cannotBeUpdated("age"));
```
chr1sjf0x, line 125: thought: I was half expecting this to look something like 
```suggestion
config.cannotBeChanged("age");
```
But after further digging I see how you are going about it...

And you plan to do use this marker in the future for possible short circuiting 🤔  Interesting.

chr1sjf0x (approved):

packages/integration-tests/src/entities/Author.ts
stephenh: Fixed, thanks!

PR #351, opened 2022-06-05
feat: Field-level validation rules

stephenh:
This changes the `config.addRule` API from using load hints:

```
addRule<H extends LoadHint<T>>(populate: H, rule: ValidationRule<Loaded<T, H>>): void;
```

To using a new `ReactiveHint` that is field-level:

```
addRule<H extends ReactiveHint<T>>(hint: H, rule: ValidationRule<Reacted<T, H>>): void;
```

I.e. specifically an old rule of `Author` running some logic anytime one of it's child books changes:

```
// Example of reactive rule being fired on Book change
// config.addRule("books", async (a) => {
```

Can now say specifically which fields in the `Book` (and the `Author` itself) it wants to access:

```
config.addRule({ books: ["title"], firstName: {} }, async (a) => {
```

The upshot is that any hint-driven rules are now executed only if the fields they include actually change.

To enforce "make sure you put the fields you access in the hint" (otherwise you might write logic using a value that you won't react against), the rule lambda now takes a new `Reacted<Entity, someHint>` mapped type. Similar to how the existing `Loaded<Entity, someHint>` would bless certain relations as `.get`-able, `Reacted` does the same for fields, such that if a field is not in a hint, the rule is not able to access to it.

To ease migration + satisfy valid special circumstances, lambdas can access fields w/o triggering reactivity in two ways:

1. Including the field in the hint but with a "read only" `:ro` suffix, i.e. `{ books: ["title:ro"] }`
2. Accessing the `Reacted.entity` property, which is an escape hatch to the full entity with all of its fields & methods.

Another up-shot of bifurcating validation rules into "simple / hint-less / non-reactive" and "hint/reactive" is that we run all of the simple rules first, and since those include the required rules, it means that the 2nd pass any lambdas for custom rules will not see partially valid entities (at least in terms of type-safety on relations).


packages/integration-tests/src/relations/OneToManyLargeCollection.test.ts
```diff
@@ -33,8 +33,7 @@ describe("OneToManyLargeCollection", () => {
   });
 
   it("cannot be a load hint", async () => {
-    // ...actually this does work, b/c Tag doesn't have any other loadable collections,
-    // so NestedLoadHint becomes {}, and turns out anything extends {}...
+    // @ts-expect-error
     const hint: LoadHint<Tag> = { publishers: {} };
```
stephenh, line 37: This is actually supposed to be a compile error, and the reworking of `LoadHint` somehow "fixed" it to actually be a compile error. No idea why. :shrug:  :-)

packages/orm/src/changes.ts
```diff
@@ -17,8 +17,6 @@ export interface ManyToOneFieldStatus<T extends Entity> extends FieldStatus<IdOf
   originalEntity: Promise<T | undefined>;
 }
 
-type ExcludeNever<T> = Pick<T, { [P in keyof T]: T[P] extends never ? never : P }[keyof T]>;
```
stephenh, line 20: The key mapping feature of TS 4.1 (i.e. mapping `foo` -> `fooChanged`) also means `ExcludeNever` is not necessary anymore b/c if you do:

```
[P in keyof T as ...some condition... extends never ? never : K]: ...
```

Any keys that the `as` evaluates to `never` will just be dropped from the type.

A little verbose, but neat. I rewrote some of the `LoadHint` / `Loadable` things to use this.

packages/orm/src/relations/hasAsyncProperty.ts
```diff
@@ -23,7 +23,7 @@ export interface LoadedProperty<T extends Entity, V> {
  * with `someProperty.get`.
  */
 export function hasAsyncProperty<T extends Entity, H extends LoadHint<T>, V>(
-  loadHint: H,
+  loadHint: Const<H>,
```
stephenh, line 26: Not entirely sure why, but one of the `Loaded` refactorings meant this `loadHint` was losing it's `const`-ness, so the `fn` lambda was being typed as `Loaded<Book, LoadHint<Book>>` which is too generic, instead of the specific const `Loaded<Book, "authors">` or what not.

We already use this `Const` helper to `const`-ize the params in `EntityManager.populate`, but for some reason it wasn't necessary here before... :shrug:

packages/integration-tests/src/reactiveHints.test.ts
```diff
@@ -0,0 +1,186 @@
...
+    }
+  });
+});
+
+// it("can type-check", () => {
+//   // author hint that reruns on book review changes
+//   const hint = {
+//     "book:rx": { "reviews:rx": ["title:rx"] },
+//   };
+// });
```
chr1sjf0x, line 186: quibble: Remove?
```suggestion
```

packages/integration-tests/src/reactiveHints.test.ts
```diff
@@ -0,0 +1,186 @@
...
+    it("works with parent m2o and primitive field names", () => {
+      expect(convertToLoadHint(getMetadata(Author), { publisher: "name" })).toStrictEqual({ publisher: {} });
+      expect(convertToLoadHint(getMetadata(Book), { author: "firstName" })).toStrictEqual({ author: {} });
+      expect(convertToLoadHint(getMetadata(BookReview), { book: { author: "firstName" } })).toStrictEqual({
+        book: { author: {} },
+      });
+    });
+  });
+
+  describe("typings", () => {
```
chr1sjf0x, line 112: curious: Since there are no `it` blocks, or expects, I'm guessing these are ensuring that the code passes type checking?

packages/integration-tests/src/reactiveHints.test.ts
```diff
@@ -0,0 +1,186 @@
+import { Author, Book, BookReview, Image, Publisher } from "@src/entities";
+import { getMetadata, Loaded, LoadHint, Reacted, ReactiveHint, reverseHint, reverseReactiveHint } from "joist-orm";
+import { convertToLoadHint } from "joist-orm/build/src/reactiveHints";
+
+describe("reactiveHints", () => {
```
chr1sjf0x, line 5: praise: great tests 👍

packages/orm/src/EntityManager.ts
```diff
@@ -672,7 +673,10 @@ export class EntityManager<C = {}> {
           await recalcAsyncDerivedFields(this, todos);
 
           if (!skipValidation) {
-            await validate(todos);
+            // Run simple rules first b/c it includes not-null/required rules, so that then when we run
+            // `validateReactiveRules` next, the lambdas won't see invalid entities.
```
chr1sjf0x, line 677: curious: Why not? Will they be stripped out of the `todos` array?

chr1sjf0x (approved):
Looks great, neat feature.

packages/orm/src/EntityManager.ts
stephenh: If any simple rules fail, we'll blow up the `flush` with a `ValidationError`, so just won't get to the 2nd step of calling the hint-based rules/lambdas.

packages/integration-tests/src/EntityManager.reactiveRules.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+    // Then we invoke both rules on initial save
+    expect(b.firstNameRuleInvoked).toBe(1);
+    expect(b.favoriteColorsRuleInvoked).toBe(1);
+    // And when the Author.firstName changes
+    a.firstName = "a2";
+    await em.flush();
+    // Then only the firstName validation rule runs again
+    expect(b.firstNameRuleInvoked).toBe(2);
+    expect(b.favoriteColorsRuleInvoked).toBe(1);
+    // And when we change something else about the author
```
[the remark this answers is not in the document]
stephenh: Cool, added!

stephenh:
Thanks for the review @chr1sjf0x !

PR #352, opened 2022-06-08
fix: support jest 28

blimmer:
In Jest v28, a newer version of `expect` is used that does not define `expect/build/jestMatchersObject` in the `files` portion of their `package.json`. Therefore, when you try to `require` it, it throws an error:

```
    Cannot find module 'expect/build/jestMatchersObject' from 'src/utils/toIncludeDisabledReason.ts'

      11 |   expectedDisabledReason: PotentialOperationReason | string,
      12 | ) {
    > 13 |   const { getMatchers } = require("expect/build/jestMatchersObject");
         |                           ^
      14 |   if (
      15 |     !potentialOperation.disabledReasons?.find((r) =>
      16 |       typeof expectedDisabledReason === "string"

      at Resolver._throwModNotFoundError (node_modules/jest-resolve/build/resolver.js:491:11)
      at Object.require (src/utils/toIncludeDisabledReason.ts:13:27)
      at __EXTERNAL_MATCHER_TRAP__ (node_modules/expect/build/index.js:346:30)
      at Object.throwingMatcher [as toIncludeDisabledReason] (node_modules/expect/build/index.js:347:15)
```

To work around this problem, we can import the matchers exposed through the `files` array: https://github.com/facebook/jest/blob/96dae9947b85dc0dcc0636af9516958ea32be752/packages/expect/package.json#L18

I've confirmed this works on Jest v27 and v28, so it shouldn't be a breaking change. I've also used `patch-package` to confirm this fix on the project I was upgrading to Jest v28.


packages/codegen/package.json
```diff
@@ -28,9 +28,9 @@
   },
```
blimmer, line 28: These `package.json` updates were made by running

```
yarn up jest @types/jest @swc/jest jest-junit
```

packages/test-utils/src/toMatchEntity.ts
```diff
@@ -71,10 +73,8 @@ export async function toMatchEntity<T>(actual: Entity, expected: MatchedEntity<T
     }
   }
 
-  // Blatantly grab `toMatchObject` from the guts of expect
-  const { getMatchers } = require("expect/build/jestMatchersObject");
```
blimmer, line 75: Was there a reason the `require` was being done in the method vs. as a top-level import?

stephenh: Ah hm, not that I recall :thinking: , probably just my at-the-time way of doing an import w/o knowing I could `@ts-ignore` it.

PR #354, opened 2022-06-20
feat: em.findGql support passing { ne: null } to nullable foreign keys

chr1sjf0x:
Ensures that code such as this passes type checking:
```
const authors = await em.findGql(Author, { publisher: { ne: null } });
```

### DIFF
--- a/packages/codegen/src/generateEntityCodegenFile.ts
+++ b/packages/codegen/src/generateEntityCodegenFile.ts
@@ -646,14 +646,16 @@ function generateGraphQLFilterFields(meta: EntityDbMetadata): Code[] {
   const pgEnums = meta.pgEnums.map(({ fieldName, enumType }) => {
     return code`${fieldName}?: ${EnumGraphQLFilter}<${enumType}>;`;
   });
-  const m2o = meta.manyToOnes.map(({ fieldName, otherEntity }) => {
-    return code`${fieldName}?: ${EntityGraphQLFilter}<${otherEntity.type}, ${otherEntity.idType}, ${GraphQLFilterOf}<${otherEntity.type}>>;`;
+  const m2o = meta.manyToOnes.map(({ fieldName, otherEntity, notNull }) => {
+    return code`${fieldName}?: ${EntityGraphQLFilter}<${otherEntity.type}, ${otherEntity.idType}, ${GraphQLFilterOf}<${
+      otherEntity.type
+    }>, ${nullOrNever(notNull)}>;`;
   });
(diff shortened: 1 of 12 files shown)


packages/codegen/src/generateEntityCodegenFile.ts
```diff
@@ -646,14 +646,16 @@ function generateGraphQLFilterFields(meta: EntityDbMetadata): Code[] {
...
+      otherEntity.type
+    }>, ${nullOrNever(notNull)}>;`;
   });
   const o2o = meta.oneToOnes.map(({ fieldName, otherEntity }) => {
-    return code`${fieldName}?: ${EntityGraphQLFilter}<${otherEntity.type}, ${otherEntity.idType}, ${GraphQLFilterOf}<${otherEntity.type}>>;`;
+    return code`${fieldName}?: ${EntityGraphQLFilter}<${otherEntity.type}, ${otherEntity.idType}, ${GraphQLFilterOf}<${otherEntity.type}>, null | undefined>;`;
   });
   const polys = meta.polymorphics.map(({ fieldName, fieldType }) => {
-    return code`${fieldName}?: ${EntityGraphQLFilter}<${fieldType}, ${IdOf}<${fieldType}>, never>;`;
+    return code`${fieldName}?: ${EntityGraphQLFilter}<${fieldType}, ${IdOf}<${fieldType}>, never, null | undefined>;`;
```
stephenh, line 658: Thought: I think polys can be either nullable or not-nullable, so we could probably do a conditional here as well?

stephenh (approved):
Great!

packages/codegen/src/generateEntityCodegenFile.ts
chr1sjf0x: I was following the example from `EntityFilter` above, which assumed `null | undefined`
image

I could try adding that in, but feel like I would like to add some more test cases around both. Looks like we currently only have a `notNull` polymorphic relationship.

stephenh: Cool, yeah sgtm to punt. If anything this is the more permissive version.

chr1sjf0x: Looked into that a bit. I think it warrants a separate follow up PR, since this work as-is will unblock some work we need to do.

stephenh: Cool! Feel free to merge/release when you're ready

PR #356, opened 2022-06-21
feat: [SC-15953] em.clone enhancements

chr1sjf0x:
Enhancements to `em.clone` to allow for:
1) Optional `skipIf` function to avoid cloning an entity deep in the object graph if desired (ex: don't clone soft-deleted entities)
2) Optional `postClone` function to be called for each `original/clone` pair
3) Optional ability to pass an array of objects to be clones, instead of just 1

### DIFF
--- a/packages/integration-tests/src/EntityManager.clone.test.ts
+++ b/packages/integration-tests/src/EntityManager.clone.test.ts
@@ -1,0 +1,270 @@
+import { insertAuthor } from "@src/entities/inserts";
+import { EntityConstructor, EntityManager } from "joist-orm";
+import { Author, Book, Comment, Image, ImageType, newAuthor, newBook, newPublisher, Publisher, Tag } from "./entities";
+import { newEntityManager } from "./setupDbTests";
+
+describe("EntityManager", () => {
+  it("can clone entities", async () => {
+    const em = newEntityManager();
+
+    // Given an entity
+    const p1 = newPublisher(em, { name: "p1" });
+    const a1 = new Author(em, { firstName: "a1", publisher: p1 });
+    await em.flush();
+
+    // When we clone that entity
+    const a2 = await em.clone(a1);
+    await em.flush();
+
+    // Then we expect the cloned entity to have the same properties as the original
+    expect(a2.firstName).toEqual(a1.firstName);
+    expect(a2.publisher.idOrFail).toEqual(p1.id);
+    expect(a2.id).not.toEqual(a1.id);
+    expect(await numberOf(em, Author, Publisher)).toEqual([2, 1]);
+    expect(p1.authors.get).toEqual([a1, a2]);
+  });
+
+  it("can clone entities and referenced entities", async () => {
+    const em = newEntityManager();
+
+    // Given an entity with a reference to another entity
+    const a1 = newAuthor(em, { firstName: "a1" });
+    const b1 = newBook(em, { title: "b1", author: a1 });
+    // And the author itself points to the book we'll clone
+    a1.currentDraftBook.set(b1);
+    await em.flush();
+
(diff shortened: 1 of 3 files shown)


packages/integration-tests/src/EntityManager.clone.test.ts
```diff
@@ -0,0 +1,270 @@
+import { insertAuthor } from "@src/entities/inserts";
+import { EntityConstructor, EntityManager } from "joist-orm";
+import { Author, Book, Comment, Image, ImageType, newAuthor, newBook, newPublisher, Publisher, Tag } from "./entities";
+import { newEntityManager } from "./setupDbTests";
+
+describe("EntityManager", () => {
```
chr1sjf0x, line 6: FYI - I thought these tests were getting sufficiently long to split out from `EntityManager.test.ts`. New test cases start at line 170

packages/integration-tests/src/EntityManager.clone.test.ts
```diff
@@ -0,0 +1,270 @@
...
+    const a2 = await em.clone(a1, { hint: "comments" });
+    await em.flush();
+    // Then we expect the cloned entity to have cloned copies of all its nested references
+    expect(a2.comments.get.length).toBe(2);
+    expect(a2.comments.get[0].id).toBe("comment:3");
+    expect(a2.comments.get[1].id).toBe("comment:4");
+    expect(a2.comments.get[0].parent.get).toBe(a2);
+  });
+
+  it("can clone a collection of entities", async () => {
```
chr1sjf0x, line 170: FYI - New test cases start here.

packages/integration-tests/src/EntityManager.clone.test.ts
```diff
@@ -0,0 +1,270 @@
...
+
+    // Given an entity with a reference to another entity
+    const a1 = newAuthor(em, { firstName: "a1" });
+    const b1 = newBook(em, { title: "b1", author: a1 });
+    // And the author itself points to the book we'll clone
+    a1.currentDraftBook.set(b1);
+    await em.flush();
+
+    // When we clone that entity and its reference
+    const a2 = await em.clone(a1, { hint: "books" });
```
chr1sjf0x, line 38: FYI - This is a diff from the moved test. `hint` is now in the `opts` instead of the second argument. Seemed like the best path forward with `hint` already being an optional argument. Open to other signature suggestions.

packages/integration-tests/src/EntityManager.clone.test.ts
```diff
@@ -0,0 +1,270 @@
...
+  describe("skipIf", () => {
+    it("can skip the main entity you asked to clone", async () => {
+      const em = newEntityManager();
+      // Given an entity to clone
+      const p1 = newPublisher(em, { name: "p1" });
+      await em.flush();
+      // When I attempt to clone it, but ask to skip the entity
+      await expect(em.clone(p1, { skipIf: (e) => e === p1 }))
+        // Then I expect it to fail
+        .rejects.toThrow("no entities were cloned given the provided option");
```
chr1sjf0x, line 229: though: This seemed like the `best` thing to do in this case, but open to suggestions

stephenh (approved):
Nice!

packages/integration-tests/src/EntityManager.clone.test.ts
chr1sjf0x: TY - I meant to change that 👍 
```suggestion
describe("EntityManager.clone", () => {
```

packages/integration-tests/src/EntityManager.clone.test.ts
chr1sjf0x: 👍

PR #361, opened 2022-06-24
feat: Skip reacting to immutable fields.

stephenh:


packages/integration-tests/src/entities/Author.ts
```diff
@@ -122,14 +123,19 @@ config.addRule("books", (a) => {
   }
 });
 
-// Example of rule that won't get run unless field is set
+// Example of rule that is always run even if the field is not set
```
stephenh, line 126: This behavior changed in the main field-level validation PR, but comment wasn't updated

packages/integration-tests/src/reactiveHints.test.ts
```diff
@@ -93,6 +93,26 @@ describe("reactiveHints", () => {
     ]);
   });
 
+  it("can do read-only string hint", () => {
```
stephenh, line 96: These are some extra tests I happened to have locally, unrelated to the current change

packages/integration-tests/src/entities/Author.ts
```diff
@@ -122,14 +123,19 @@ config.addRule("books", (a) => {
...
 config.addRule("mentor", (a) => {
   a.entity.mentorRuleInvoked++;
 });
 
 // Example of cannotBeUpdated
 config.addRule(cannotBeUpdated("age"));
 
+// Example of a rule against an immutaable field
```
chr1sjf0x, line 134: quibble: 
```suggestion
// Example of a rule against an immutable field
```

chr1sjf0x (approved):
Looks good 👍  🍣

PR #373, opened 2022-07-16
fix: Correct clearing of o2m/m2o fks when entity data references the entity instead of ids

chr1sjf0x:
I've traced this down to the scenario when `this.entity.__orm.data[this.fieldName]` returns an actual entity instance, as opposed to just the ID of the referenced entity.

image

### DIFF
--- a/packages/integration-tests/src/relations/OneToManyCollection.test.ts
+++ b/packages/integration-tests/src/relations/OneToManyCollection.test.ts
@@ -182,6 +182,22 @@ describe("OneToManyCollection", () => {
     expect(p1.authors.get.length).toEqual(0);
   });
 
+  it("removes deleted entities from other foreign key", async () => {
+    // Given an author with a publisher
+    await insertPublisher({ name: "p1" });
+    await insertAuthor({ first_name: "a1", publisher_id: 1 });
+    const em = newEntityManager();
+    // And the a1.publishers collection is loaded
+    const p1 = await em.load(Publisher, "1", "authors");
+    const a1 = p1.authors.get[0];
+    expect(p1.authors.get.length).toEqual(1);
+    // When we delete the publisher
+    em.delete(p1);
+    await em.flush();
+    // Then the author.publisher is cleared
+    expect(a1.publisher.isSet).toEqual(false);
+  });
+
   it("respects deleted entities before the collection loaded", async () => {
     // Given an author with a publisher
     await insertPublisher({ name: "p1" });
--- a/packages/orm/src/relations/OneToManyCollection.ts
+++ b/packages/orm/src/relations/OneToManyCollection.ts
@@ -199,7 +199,7 @@ export class OneToManyCollection<T extends Entity, U extends Entity>
     const current = await this.load({ withDeleted: true });
     current.forEach((other) => {
       const m2o = this.getOtherRelation(other);
-      if (maybeResolveReferenceToId(m2o.current()) === this.entity.id) {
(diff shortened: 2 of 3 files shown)


packages/integration-tests/src/relations/OneToManyCollection.test.ts
```diff
@@ -182,6 +182,21 @@ describe("OneToManyCollection", () => {
     expect(p1.authors.get.length).toEqual(0);
   });
 
+  it("removes deleted entities from other foreign key", async () => {
+    // Given an publisher with an author
+    const em = newEntityManager();
+    // And the publisher and authors are loaded
+    const p1 = newPublisher(em, { authors: [{}] });
```
chr1sjf0x, line 189: @stephenh - FYI: I couldn't figure out how to get `this.entity.__orm.data[this.fieldName]` to be the actual `Publisher` entity, and not `"p:1"` without using the `newPublisher` factory 🤔

If you have thoughts on how to repro this issue using `insertPublisher` and `em.load`, let me know and I can update.

packages/orm/src/relations/OneToManyCollection.ts
```diff
@@ -199,7 +199,7 @@ export class OneToManyCollection<T extends Entity, U extends Entity>
     const current = await this.load({ withDeleted: true });
     current.forEach((other) => {
       const m2o = this.getOtherRelation(other);
-      if (maybeResolveReferenceToId(m2o.current()) === this.entity.id) {
+      if (maybeResolveReferenceToId(m2o.current({ withDeleted: true })) === this.entity.id) {
```
chr1sjf0x, line 202: thought: I wonder if an audit of use cases of `current` for other missing `withDeleted` clauses is warranted.

packages/integration-tests/src/relations/OneToManyCollection.test.ts
stephenh: Yeah, I've pretty recently thought a similar "...wait, `data[fieldName]` is an entity" bugs I believe in either clone or changes.

I think the rub is that `data[fieldName]` is basically always an id/string _unless_ the `data[fieldName]` entity was brand new and didn't have an id, so "of course" needs to stay an object.

So, I think this is why these things "only show up when using factories" is that it's less the factory per se, and more about "oh this is a test that created the entities and, even if it's done an `em.flush`, is still using the same entities the test itself directly created".

packages/orm/src/relations/OneToManyCollection.ts
stephenh: Likely...fwiw I believe `current` is an internal method, so I wouldn't be against draconian methods like _requiring_ `withDeleted` just for good measure. (Dunno, that would get really ugly, but just talking out loud.)

stephenh (approved):
Wow, that's subtle.

What's interesting about that current method:

```
  current(opts?: { withDeleted?: boolean }): U | string | N {
    const current = this.entity.__orm.data[this.fieldName];
    if (current !== undefined && isEntity(current)) {
      return this.filterDeleted(current as U, opts);
    }
    return current;
  }
```

Is that per the comment on factories, a) the `withDeleted` param only actually works for when `data[fieldName]` is an entity, and b) I think it is actually rarely an entity (only for the mentioned pointed-to-entity-was-new case).

Makes me think that it's inaccurate for `current` to advertise a `withDeleted` capability, when it will really only enforce that in rare/at least not majority circumstances...

But great find!

PR #383, opened 2022-07-28
fix: Deleted entities should trigger async derived fields.

stephenh:


zovington (approved):

packages/orm/src/EntityManager.ts
```diff
@@ -1013,11 +1013,14 @@ async function addReactiveValidations(todos: Record<string, Todo>): Promise<void
  */
 async function addReactiveAsyncDerivedValues(todos: Record<string, Todo>): Promise<void> {
   const p: Promise<void>[] = Object.values(todos).flatMap((todo) => {
-    const entities = [...todo.inserts, ...todo.updates];
+    const entities = [...todo.inserts, ...todo.updates, ...todo.deletes];
```
stephenh, line 1016: What's kind of embarrassing about this is that the `addReactiveValidations` function right above this had the same/correct `...todo.deletes`, but this ~80%-the-same chunk of code just didn't get updated/fixed to include it as well.

PR #389, opened 2022-08-07
feat: Add databaseUrl to joist-codegen

### LINKED ISSUE #382: Support databaseUrl in joist-codegen.json
For users who really don't want to use environment variables.

stephenh:
Fixes #382


stephenh:
Fyi @mmahalwy you'd mentioned a way to avoid the `run.sh` script; at least for `npm run joist-codegen`, this should break that dependency on "`DATABASE_URL` _must_ be in the env".

Tests and production code are still an exercise left to the reader. :-)

PR #396, opened 2022-08-19
feat: Add EntityManager.loadLens.

stephenh:
This allows applying a lens-based navigation to multiple starting
entities, instead of a single entity via `Entity.load`, i.e.:

```
const publishers = await em.loadLens(books, b => b.author.publisher);
```

Vs. today you can only use the `b => b.author.publisher` syntax if you're
starting at a single book and do `await book.load(b => ...)`.

### DIFF
--- a/packages/orm/src/EntityManager.ts
+++ b/packages/orm/src/EntityManager.ts
@@ -549,8 +549,22 @@ export class EntityManager<C = {}> {
    * Results are unique, i.e. if doing `em.loadLens([b1, b2], b => b.author.publisher)` point to the
    * same `Publisher`, it will only be returned as a single value.
    */
-  public async loadLens<T extends Entity, U, V>(entities: T[], fn: (lens: Lens<T>) => Lens<U, V>): Promise<V> {
-    return loadLens(entities, fn);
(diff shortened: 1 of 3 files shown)


packages/orm/src/EntityManager.ts
```diff
@@ -541,6 +543,16 @@ export class EntityManager<C = {}> {
     return entities as T[];
   }
 
+  /**
+   * Loads entities found, when starting at `entities`, via the "path" given by the `fn` lens function.
+   *
+   * Results are unique, i.e. if doing `em.loadLens([b1, b2], b => b.author.publisher)` point to the
+   * same `Publisher`, it will only be returned as a single value.
+   */
+  public async loadLens<T extends Entity, U, V>(entities: T[], fn: (lens: Lens<T>) => Lens<U, V>): Promise<V> {
```
stephenh, line 552: I don't love the name `loadLens`; I'd originally thought about overloading `loadAll` like:

```
const pubs = await em.loadAll(books, b => b.author.publisher);
```

But we've already got a `loadAll` that is `loadAll(Book, ids)` and it seemed like this was too different to be an overload.

chr1sjf0x (commented):
cool 👍

PR #414, opened 2022-09-08
feat: Enhanced support for reversing polymorphic references and reacting to changes through them

chr1sjf0x:
Makes it possible to have an async derived persisted field which reacts to changes through a polymorphic references.

Includes enhanced support for polymorphic relationships via `changes.originalEntity` as well.


packages/integration-tests/src/reactiveHints.test.ts
```diff
@@ -52,6 +52,14 @@ describe("reactiveHints", () => {
     ]);
   });
 
+  it("can do via polymorphic reference", () => {
+    expect(reverseReactiveHint(Author, { books: { comments: "text" } })).toEqual([
+      { entity: Author, fields: [], path: [] },
+      { entity: Book, fields: ["author"], path: ["author"] },
+      { entity: Comment, fields: ["parent", "text"], path: ["parent@Book", "author"] },
```
chr1sjf0x, line 59: @stephenh: This allows us to differentiate the type of the polymorphic entity these reactions should go through

packages/orm/src/changes.ts
```diff
@@ -79,7 +80,7 @@ export function newChangesProxy<T extends Entity>(entity: T): Changes<T> {
           if (isEntity(originalValue)) {
             return Promise.resolve(originalValue);
           } else if (isId(originalValue)) {
-            return entity.em.load((entity as any)[p].otherMeta.cstr, originalValue);
```
chr1sjf0x, line 82: @stephenh: FYI - this line didn't work for polymorphic relationships before.

PR #418, opened 2022-09-14
fix: Fix EntityManager.populate not checking loadMany

stephenh:


packages/orm/src/EntityManager.ts
```diff
@@ -612,67 +613,62 @@ export class EntityManager<C = {}> {
...
+
         // We disable caching if `forceReload` is enabled to ensure we re-touch any previously-loaded
         // collections; otherwise using `cache: false` all the time does reduce the effectiveness of
         // the fan-in and results in duplicate populate calls.
         { cache: !opts.forceReload },
       ),
     );
 
-    // Purposefully use `then` instead of `async` as an optimization
-    return loader.loadMany(list).then(() => (!fn ? (entityOrList as any) : fn(entityOrList as any)));
```
stephenh, line 675: This `loadMany` is returning a `Promise<V | Error>` and we weren't looking for errors, i.e. if an `AsyncProperty` calc blows up.

packages/integration-tests/src/EntityManager.clone.test.ts
```diff
@@ -24,6 +36,22 @@ describe("EntityManager.clone", () => {
...
+    const a1 = new Author(em, { firstName: "a1", publisher: p1 });
+    await em.flush();
+
+    // When we clone it in a 2nd UoW
+    const em2 = newEntityManager();
+    const a2 = await em2.load(Author, a1.idOrFail);
+    const a3 = await em2.clone(a2);
+
+    // Then the a3.publisher loaded state is correct
+    expect(a3.publisher.isLoaded).toBe(false);
```
stephenh, line 52: The "only call `.load` if `!isLoaded`" optimization to EntityManager broke `M2O.get` for entities that were cloned, b/c the cloned value would:

1) See its `M2O.initializeNewEntity` set `isLoaded` to true
2) But the value dropped into `clone.__orm.data` could end up being a string

Fixing this turned into a large refactoring of `em.clone` to not longer just write directly to `__orm.data`, but to go through `em.create` / the constructor with a set of opts just like any other caller would.

I think this is much cleaner, albeit it required teaching `PolymorphicReference` how to be initialized via `em.create` with non-entity/string ids (similar to how we taught `ManyToOneReference` to do the same a few months ago).

packages/integration-tests/src/EntityManager.clone.test.ts
```diff
@@ -176,6 +182,20 @@ describe("EntityManager.clone", () => {
...
+    // Given an entity with a polymorphic parent
+    const c1 = newComment(em, { parent: {} });
+    expect(c1.parent.get).toBeInstanceOf(Author);
+    await em.flush();
+    // When we clone it in a new UoW
+    const em2 = newEntityManager();
+    const c2 = await em2.load(Comment, c1.idOrFail);
+    const c3 = await em2.clone(c2);
+    // Then the parent can be loaded
+    expect(await c3.parent.load()).toBeInstanceOf(Author);
```
stephenh, line 196: This test was breaking until the "teach PolymorphicReference about having ids set" fix (given em.clones new pass-opts-to-the-cstr strategy).

packages/orm/src/EntityManager.ts
```diff
@@ -612,67 +630,62 @@ export class EntityManager<C = {}> {
...
-              } else {
-                throw new Error(`Unexpected hint ${hint}`);
-              }
+          const hints = Object.entries(normalizeHint(hintOpt as any));
+
+          // One breadth-width pass to ensure each relation is loaded
+          const loadPromises = list.flatMap((entity) => {
+            return hints.map(([key]) => {
+              const relation = (entity as any)[key];
+              return relation.isLoaded && !opts.forceReload ? undefined : (relation.load(opts) as Promise<any>);
```
stephenh, line 659: This `relation.isLoaded` check is what surfaced the clone bug in `ManyToOneReference`.

packages/orm/src/EntityManager.ts
```diff
@@ -612,67 +630,62 @@ export class EntityManager<C = {}> {
...
+
         // We disable caching if `forceReload` is enabled to ensure we re-touch any previously-loaded
         // collections; otherwise using `cache: false` all the time does reduce the effectiveness of
         // the fan-in and results in duplicate populate calls.
         { cache: !opts.forceReload },
       ),
     );
 
-    // Purposefully use `then` instead of `async` as an optimization
-    return loader.loadMany(list).then(() => (!fn ? (entityOrList as any) : fn(entityOrList as any)));
```
stephenh, line 675: This `loadMany()` not checking its return value is the original bug that started this rewrite of `populate`.

What would happen is that `.load()` on an async property would blow up, and we would ignore it, but then since the promise was rejected, it would ~early return (i.e. not `allSettled` behavior), and code that used the `populate` results would start calling `.get` into not-yet-loaded relations, and get "not loaded" errors.

PR #422, opened 2022-09-21
fix: Support PersistedAsyncProperties in toMatchEntity.

stephenh:


TylerR909 (approved):
I like the way this is

PR #424, opened 2022-09-21
fix: Do not reuse entities that have unique constraints.

stephenh:


packages/integration-tests/src/EntityManager.factories.test.ts
```diff
@@ -299,9 +299,13 @@ describe("EntityManager.factories", () => {
...
     // When we create an entity that has a m2o to author
-    const i = newImage(em);
+    const i1 = newImage(em);
     // Then the m2o reused the existing entity
-    expect(i.author.get).toEqual(a);
+    expect(i1.author.get).toEqual(a);
+    // But if we make a _2nd_ image
+    const i2 = newImage(em);
+    // Then we don't use the same author b/c of the unique constraint
+    expect(i2.author.get).toBeUndefined();
```
TylerR909, line 308: So if I did `threeOf(() => newImage(em))` then asserted against them it'd be:

```
i1.author.get === a
i2.author.get === undefined
i3.author.get === undefined
```

I'd think we'd just want new unique things. 😕

stephenh: That will happen if the fields are required:

```
        // Otherwise, only make a new entity only if the field is required
        if (field.required) {
          return [fieldName, resolveFactoryOpt(em, opts, field, undefined, undefined)];
        }
```

Right below this; the lines being updated are for the factories seeing "ah, this field is not _required_ but you probably want this one because there is only one anyway".

PR #432, opened 2022-10-01
feat: Support async properties in reactive hints.

stephenh:
This includes both non-persisted `AsyncProperty` and `PersistedAsyncProperty`s.

`PersistedAsyncProperty` basically worked for free, b/c a rule depending on persisted field would already get the transitiveness basically for free:

1) new `Book` recalcs the `Author.numberOfBooks` persisted property,
2) `Author`'s reactive validations see `numberOfBooks` changing and queue the `Publisher` rule to rerun.

(Note that `em.flush`s design of "loop over all pending entities first, then do a single validate pass" worked out really well to avoid ordering issues, i.e. we don't have to worry about the Author's reactive validations (step 2) having already been evaluated before the step 1 recalc ran, which if it did happen would mean missing that `numberOfBook`s _will_ change but just hasn't yet.)

For `AsyncProperty`, we don't have the persisted property (`numberOfBooks`) to implicitly stitch together the step 1 "this reactive hint triggers when the property has changed" (from `Book` -> `Author.numberOfBooks2`) with step 2 "this reactive hint triggers the validation rule" (from `Author.numberOfBooks2` -> `Publisher`), so we just explicitly merge them together into a single reactive hint.

I.e. in `reverseReactiveHint` (e.g. while creating the `Author.numberOfBooks2` -> `Publisher` hint for the `Publisher` rule) if we notice an async property (`numberOfBooks2`), we just copy/paste that property's reactive hint (`Book` -> `Author.numberOfBooks2`) into the current reactive hint, and so end up with a singular `Book` -> `Author` -> `Publisher` hint.

So, `Book` ends up with a reaction that, when Book changes, re-evals the `Publisher` validation rule, even though `Publisher`'s validation rule hint (`{ authors: numberOfBooks2 }`) doesn't explicitly show/know that `Author.numberOfBooks2` accessed `Book`.

### DIFF
--- a/packages/orm/src/reactiveHints.ts
+++ b/packages/orm/src/reactiveHints.ts
@@ -1,6 +1,7 @@
 import { Entity } from "./Entity";
 import { EntityConstructor, FieldsOf } from "./EntityManager";
 import { EntityMetadata, getMetadata } from "./EntityMetadata";
+import { getProperties } from "./getProperties";
 import { Loadable, Loaded, LoadHint } from "./loadHints";
 import { NormalizeHint, normalizeHint, suffixRe, SuffixSeperator } from "./normalizeHints";
 import { Collection, LoadedCollection, LoadedReference, OneToOneReference, Reference } from "./relations";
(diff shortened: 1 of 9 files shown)


packages/orm/src/reactiveHints.ts
```diff
@@ -59,49 +70,59 @@ export function reverseReactiveHint<T extends Entity>(
...
     const key = keyMaybeSuffix.replace(suffixRe, "");
-    const field = meta.fields[key] || fail(`Invalid hint ${entityType.name} ${JSON.stringify(hint)}`);
-    const isReadOnly = !!keyMaybeSuffix.match(suffixRe) || field.immutable;
-    switch (field.kind) {
-      case "m2o": {
-        if (!isReadOnly) {
-          primitives.push(field.fieldName);
+    // const field = meta.fields[key] || fail(`Invalid hint ${entityType.name} ${JSON.stringify(hint)}`);
+    const field = meta.fields[key];
+    const isReadOnly = !!keyMaybeSuffix.match(suffixRe) || (field && field.immutable);
```
stephenh, line 75: Use the "hide whitespace" option b/c ~90% of this diff is just indented code

PR #434, opened 2022-10-04
fix: Keep deleted children createOrUpdatePartial.

stephenh:


chr1sjf0x (approved):
TY @stephenh - Appreciate the quick fix 👍 

I may look into whether there is a way to get a similar behaviour when using the `non-incremental` version of updating collections 🤔

PR #435, opened 2022-10-04
feat: Calling o2m.set(values) deletes owned children

stephenh:
Where ownership is determined by cascade deletion from the parent to the child.


chr1sjf0x (approved):
Great, TY 👍

PR #452, opened 2022-10-23
feat: Add EntityManager.assignNewIds capability

chr1sjf0x:
For cases where you need to know the ID of an entity prior to it being flushed, you can now call `EntityManager.assignNewIds`.

### DIFF
--- a/packages/orm/src/drivers/InMemoryDriver.ts
+++ b/packages/orm/src/drivers/InMemoryDriver.ts
@@ -72,17 +72,23 @@ export class InMemoryDriver implements Driver {
     });
   }
 
+  async assignNewIds(em: EntityManager, todos: Record<string, Todo>): Promise<void> {
+    // do our version of assign ids
+    Object.entries(todos).forEach(([_, todo]) => {
+      todo.inserts
+        .filter((e) => e.id === undefined)
+        .forEach((i) => {
+          const id = this.nextId(todo.metadata.tableName);
+          i.__orm.data["id"] = keyToString(todo.metadata, id);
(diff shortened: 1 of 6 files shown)


packages/orm/src/drivers/InMemoryDriver.ts
```diff
@@ -72,17 +72,23 @@ export class InMemoryDriver implements Driver {
...
-        const id = this.nextId(todo.metadata.tableName);
-        i.__orm.data["id"] = keyToString(todo.metadata, id);
-        this.rowsOfTable(todo.metadata.tableName)[id] = {};
-      });
+      todo.inserts
+        .filter((e) => e.id === undefined)
+        .forEach((i) => {
+          const id = this.nextId(todo.metadata.tableName);
+          i.__orm.data["id"] = keyToString(todo.metadata, id);
+          this.rowsOfTable(todo.metadata.tableName)[id] = {};
```
chr1sjf0x, line 83: @stephenh - FWIW, I'm not 100% sure if adding the row here is correct or tested. But I'm also under the impression the `InMemoryDriver` is not used/maintained?

Let me know if I should care more about this and test it more thoroughly.

stephenh: Yeah, no need to worry about that.

In theory the test suite has been setup to be able to run against N drivers, like the PG driver and InMemoryDriver, but the InMemoryDriver:

a) is passing like ~80-90% of tests but not yet 100 and
b) yeah it's not actively used so :shrug: don't want to actively remove it, but yeah it's still a backburner thing for now

stephenh (approved):

stephenh:
Cool, looks good!

PR #457, opened 2022-10-23
fix: Ensure sameEntity works even if an ID has been assigned to a new entity

chr1sjf0x:
This fixes a max stack trace exceeded observed when using the original commit in graphql-service


stephenh (approved):

stephenh:
Thanks!

PR #458, opened 2022-10-24
fix: loadHints.isNew was checking for idTagged===undefined

chr1sjf0x:
Switched to using `entity.isNewEntity`


stephenh (approved):
Ah shoot! Good fix!

PR #462, opened 2022-10-29
fix: Issue where isLoaded does not correctly handle nullable loaded references

chr1sjf0x:
Discovered due to this failing test in GQL service
image

When the `changeOrder` was loaded but null, we would still attempt to recurse into `isLoaded` again giving the message above.


stephenh (approved):
Great, thanks!

PR #469, opened 2022-11-15
fix: Correctly handle partial unique indexes

chr1sjf0x:
Here is the fix discussed for partial unique indexes resulting in o2o collection instead of o2m.

The key point to observe in the diffs is that now there are no changes to the metadata. Publisher still has an `authors: Author[]` collection.

### DIFF
--- a/packages/integration-tests/migrations/1580658856631_author.ts
+++ b/packages/integration-tests/migrations/1580658856631_author.ts
@@ -82,6 +82,10 @@ export function up(b: MigrationBuilder): void {
     address: { type: "jsonb", notNull: false },
   });
 
+  // A publisher can only have one author named `Jim`, but still have other authors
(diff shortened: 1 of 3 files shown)


packages/integration-tests/migrations/1580658856631_author.ts
```diff
@@ -82,6 +82,10 @@ export function up(b: MigrationBuilder): void {
     address: { type: "jsonb", notNull: false },
   });
 
+  // A publisher can only have one author named `Jim`, but still have other authors
+  // Verifies that partial unique indexes do not result in o2o collections
+  b.createIndex("authors", ["publisher_id"], { unique: true, where: "first_name = 'Jim'" });
```
stephenh, line 87: Nice, thanks for the repro!

stephenh (approved):
Nice!

PR #474, opened 2022-11-28
feat: Add minValueRule, maxValueRule and rangeValueRule

KoltonG:
Add three new number-type rules which are:

- `minValueRule`:    A given field value cannot be smaller than a given value
- `maxValueRule`:   A given field value cannot be greater than a given value
- `rangeValueRule`: minValueRule && maxValueRule


packages/integration-tests/src/entities/Author.ts
```diff
@@ -12,6 +12,7 @@ import {
   PersistedAsyncProperty,
   Reference,
 } from "joist-orm";
+import { rangeValueRule } from "joist-orm/src/rules";
```
stephenh, line 15: Suggestion: Ideally this would be just `from "joist-orm"`, like we should go ahead and add these rules to the `src/index.ts` as exports.

packages/artifacts/junit-integration-tests.xml
```diff
@@ -0,0 +1,3 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<testsuites name="jest tests" tests="0" failures="0" errors="0" time="0.527">
+</testsuites>
```
KoltonG, line 3: @stephenh not sure why this was created. Does this look correct?

packages/integration-tests/migrations/1580658856631_author.ts
```diff
@@ -176,6 +176,8 @@ export function up(b: MigrationBuilder): void {
   createEntityTable(b, "author_stats", {
     smallint: { type: "smallint", notNull: true },
     integer: { type: "integer", notNull: true },
+    // Nullable integer
+    integer_null: { type: "integer", notNull: false },
```
KoltonG, line 180: Created this to test nullable numeric fields.

packages/artifacts/junit-integration-tests.xml
stephenh: Hm, no, not sure where this came from... The `junit...xml` files are only used by CircleCI to do test reporting.

packages/orm/src/rules.ts
```diff
@@ -87,3 +87,101 @@ function errorMessage(errors: ValidationError[]): string {
...
+        return `${field} must be a number`;
+      }
+
+      // Show an error when the value is smaller than the minimum value
+      if (value < minValue) {
+        return `${field} must be greater than or equal to ${minValue}`;
+      }
+    }
+  };
+}
```
KoltonG, line 122: I kept getting failing tests when a passing rule would return an `undefined` value. Should a passing rule not return anything?

KoltonG: I found out the issue was that I was not running `yarn build` after each change of this file 🙈

stephenh (approved):
Looks good!

PR #477, opened 2022-12-13
feat: Automatically filter soft-deleted entities.

stephenh:


packages/orm/src/relations/ManyToOneReference.ts
```diff
@@ -276,8 +276,17 @@ export class ManyToOneReferenceImpl<T extends Entity, U extends Entity, N extend
...
   }
 
+  /** Removes pending-hard-delete or soft-deleted entities, unless explicitly asked for. */
   private filterDeleted(entity: U | N, opts?: { withDeleted?: boolean }): U | N {
-    return opts?.withDeleted === true || entity === undefined || !entity.isDeletedEntity ? entity : (undefined as N);
+    if (
+      entity &&
+      (!opts || !opts.withDeleted) &&
+      (entity.isDeletedEntity ||
+        (this.#otherMeta.timestampFields.deletedAt && (entity as any)[this.#otherMeta.timestampFields.deletedAt]))
```
zgavin, line 285: imo this should just be codegenned as `isSoftDeletedEntity`.  Since it's a property it'd be undefined for entities without soft deletedness anyways

packages/orm/src/relations/OneToManyCollection.ts
```diff
@@ -251,8 +251,16 @@ export class OneToManyCollection<T extends Entity, U extends Entity>
...
+  /** Removes pending-hard-delete or soft-deleted entities, unless explicitly asked for. */
   private filterDeleted(entities: U[], opts?: { withDeleted?: boolean }): U[] {
-    return opts?.withDeleted === true ? [...entities] : entities.filter((e) => !e.isDeletedEntity);
+    return opts?.withDeleted === true
+      ? [...entities]
+      : entities.filter(
+          (e) =>
+            !e.isDeletedEntity &&
+            (this.#otherMeta.timestampFields.deletedAt === undefined ||
+              !(e as any)[this.#otherMeta.timestampFields.deletedAt]),
```
zgavin, line 262: ditto with previous comment

packages/orm/src/relations/ManyToManyCollection.ts
```diff
@@ -65,8 +65,16 @@ export class ManyToManyCollection<T extends Entity, U extends Entity>
...
+  /** Removes pending-hard-delete or soft-deleted entities, unless explicitly asked for. */
   private filterDeleted(entities: U[], opts?: { withDeleted?: boolean }): U[] {
-    return opts?.withDeleted === true ? [...entities] : entities.filter((e) => !e.isDeletedEntity);
+    return opts?.withDeleted === true
+      ? [...entities]
+      : entities.filter(
+          (e) =>
+            !e.isDeletedEntity &&
+            (this.#otherMeta.timestampFields.deletedAt === undefined ||
+              !(e as any)[this.#otherMeta.timestampFields.deletedAt]),
```
zgavin, line 76: ditto again

zgavin (approved):
Should add a `withSoftDeleted` option to `filterDeleted` and extend that to `load`/`get`.

Also could update `em.delete` to check the entity for something like `shouldSoftDelete` and soft delete instead of normal delete depending on that.

packages/orm/src/relations/ManyToOneReference.ts
stephenh: Nice! Good call, will update.

stephenh:
> Should add a withSoftDeleted option to filterDeleted and extend that to load/get.

Yeah, I was kind of hoping to avoid increasing the API, by having the existing `withDeleted` be good enough, even though it is technically ambiguous. I might go with that as an opinionated decision until it becomes obvious that its incorrect.

> Also could update em.delete to check the entity for something like shouldSoftDelete
> and soft delete instead of normal delete depending on that.

I'm a little worried about loosening the mental model of `em.delete`-means-always-`DELETE`s...

Maybe a new `EntityManager` method or `opts` to `em.delete` to opt-in to special soft delete handling?

I'd forgotten about our adhoc infra for that:

* `ProjectItem.softDelete`
* `ProjectItem.isSoftDeleted` (already added as `isSoftDeletedEntity`)
* `isSoftDeletable` / `isHardDeletable`
* `ProjectItem.deleteOrSoftDelete`

I'm unsure how to work `isSoftDeletable` into Joist's conventions; like should _every_ table with a `deleted_at` be expected to impl `isSoftDeletable` method? And then Joist codegen would always add the `softDelete` / `deleteOrSoftDelete` methods?

Seems like maybe something the user should opt-in to on an entity-by-entity basis?...

I think I'm going to ship this core soft-delete support for now, and follow up with others later/as-needed.

PR #520, opened 2023-02-03
feat: include inherited fields in graphql files

0ces:
@stephenh I saw your `allFields` (maybe you intended to create this behavior too) in the metadata generation and tried going with that approach but it seem simpler to do this, this is also works for the inheritance-chain someone may use.

I wish GQL had some kind of inheritance, maybe it could be possible using the `extend type ...` but it doesn't fits as good.


stephenh (approved):
This is great @0ces ! 

Yeah, agreed GraphQL's inheritance is kinda "meh". They technically have interfaces, so we could do like:

```graphql
interface PublisherLike {
  name: String!
}
```

And then have each type implement the interface, but much to @zgavin 's chagrin, we end up having to repeat the fields anyway, i.e.:

```graphql
type SmallPublisher implements PublisherLike {
  " Still have to type this out..."
  name: String!
}
```

So maybe someday we could codegen the `PublisherLike` interface, because that would allow a server to have like `publishers: [PublisherLike!]!` and clients could do like:

```
query {
  publishers {
    ...on SmallPublisher { city }
    ...on LargePublisher { country }
   }
}
```

PR #522, opened 2023-02-03
fix: Re-fixes the o2m loading issue w/o a scan.

stephenh:
Also adds a test for persisted fields depending on other persisted fields.


packages/integration-tests/src/EntityManager.reactiveRules.test.tsx
```diff
@@ -316,5 +323,29 @@ describe("EntityManager.reactiveRules", () => {
...
+        await em.flush();
+        expect(await select("authors")).toMatchObject([{ id: 1, number_of_public_reviews: 1 }]);
+      });
+
+      it.withCtx("calculates on new review", async ({ em }) => {
+        newAuthor(em, { age: 40, graduated: new Date(), books: [{ reviews: [] }] });
+        await em.flush();
+        // Use a new em to ensure nothing is cached
+        const em2 = newEntityManager();
+        await em2.create(BookReview, { book: "b:1", rating: 1 });
```
stephenh, line 345: @chr1sjf0x I was able to reproduce a derived field not recalculating, but it was only if I use `book: "b:1"` here and not `book: await load book`.

It ends up being broken for the "lazy loaded o2ms i.e. author.books do not see books that were pointed at the author, before the author was even loaded" bug that I fixed ~2 weeks ago in a naive / q&d way and caused a perf issue.

So I have refixed that bug, now w/o the bad perf, and it fixed this test.

...however:

a) that bug didn't really have anything to do with "derived fields on derived fields", this test written with `book: "b:1"` just happened to get tripped up by it, and

b) I'm 90% sure the `book: "b:1"` should not affect us at Homebound, b/c we use `em.createOrUpdatePartial`, which was written before `book: "b:1"` was allowed, and so it `em.load`s any m2o.set before doing the set. So like all of our save resolvers should be doing `book: loadedBook` and not `book: "b:1".

So, I'm kinda stuck; the test shows that, in theory, derived values that depend on my derived value should work/get recalced...

Well, hold on...let me try a few more things.

PR #523, opened 2023-02-07
feat: find by inherited fields

0ces:
TBH this took me more time to accomplish than I initially thought, I'm open to suggestions/improvements.

@stephenh I saw some `TODO` comments when implementing this, was something like this the goal?

### DIFF
--- a/packages/integration-tests/src/entities/LargePublisher.test.ts
+++ b/packages/integration-tests/src/entities/LargePublisher.test.ts
@@ -1,0 +1,17 @@
+import { LargePublisher } from "./LargePublisher";
+import { newLargePublisher } from "./LargePublisher.factories";
+import { newPublisherGroup } from "./PublisherGroup.factories";
+import { newSmallPublisher } from "./SmallPublisher.factories";
+
+describe("LargePublisher", () => {
+  it.withCtx("can be find by inherited fields", async (ctx) => {
+    const { em } = ctx;
+    // Given a large publisher, a small publisher and a group of publishers
+    const lp = newLargePublisher(em, { name: "lp", latitude: 1, country: "US" });
+    await em.flush();
+    // When we find it by name
+    const result = await em.findOneOrFail(LargePublisher, { name: "lp", latitude: 1, country: "US" });
+    // Then we get the large publisher
+    expect(result).toEqual(lp);
+  });
+});
--- a/packages/orm/src/QueryBuilder.ts
+++ b/packages/orm/src/QueryBuilder.ts
@@ -2,7 +2,7 @@ import { groupBy } from "joist-utils";
 import { Knex } from "knex";
 import { Entity, isEntity } from "./Entity";
 import { EntityConstructor, entityLimit, FilterOf, OrderOf } from "./EntityManager";
-import { EntityMetadata, Field, getMetadata, PolymorphicField } from "./EntityMetadata";
+import { EntityMetadata, getMetadata, PolymorphicField } from "./EntityMetadata";
 import {
   addTablePerClassJoinsAndClassTag,
   asConcreteCstr,
--- a/packages/orm/src/QueryBuilder.ts
+++ b/packages/orm/src/QueryBuilder.ts
@@ -2,7 +2,7 @@ import { groupBy } from "joist-utils";
 import { Knex } from "knex";
 import { Entity, isEntity } from "./Entity";
 import { EntityConstructor, entityLimit, FilterOf, OrderOf } from "./EntityManager";
-import { EntityMetadata, getMetadata, PolymorphicField } from "./EntityMetadata";
+import { EntityMetadata, Field, getMetadata, PolymorphicField } from "./EntityMetadata";
 import {
   addTablePerClassJoinsAndClassTag,
   asConcreteCstr,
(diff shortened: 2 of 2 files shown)


packages/orm/src/QueryBuilder.ts
```diff
@@ -318,9 +341,16 @@ export function buildQuery<T extends Entity>(
...
-        // TODO Currently we only support base-type WHEREs if the sub-type is the main `em.find`
-        // const maybeBaseAlias = field.alias;
-        query = hasClause ? addPrimitiveClause(query, alias, column, clause) : query;
+        // If the field is a base type field we add the clause to the base type table
+        const maybeBaseType = baseTypeFields[key];
+        const baseMeta = meta.baseTypes.find((b) => b.type === maybeBaseType)!;
+        if (maybeBaseType && hasClause && baseMeta) {
+          const baseTypeIndex = meta.baseTypes.indexOf(baseMeta);
+          const baseAlias = `b${baseTypeIndex}`;
+          addClauses(baseMeta, baseAlias, baseTypeTree[maybeBaseType] as any, order);
```
0ces, line 350: Curious to test this with something like this:

```ts
/**
  Where the Animal type will be something like:
  type Animal {
    name: String!
  }
*/


const human = newHuman(em, { name: "Jane" }); // Inherits from Animal
const dog = newDog(em, { name: "Zeus", owner: human }); // Also inherits from Animal

const result = await em.findOne(Dog, { name:"Zeus", owner: { name: "Jane" } });
```

Maybe worth testing this to see if my approach with the object-trees works or if it needs some tweaks.

packages/orm/src/QueryBuilder.ts
```diff
@@ -318,9 +341,16 @@ export function buildQuery<T extends Entity>(
...
         const column = serde.columns[0];
-        // TODO Currently we only support base-type WHEREs if the sub-type is the main `em.find`
-        // const maybeBaseAlias = field.alias;
-        query = hasClause ? addPrimitiveClause(query, alias, column, clause) : query;
+        // If the field is a base type field we add the clause to the base type table
+        const maybeBaseType = baseTypeFields[key];
+        const baseMeta = meta.baseTypes.find((b) => b.type === maybeBaseType)!;
+        if (maybeBaseType && hasClause && baseMeta) {
+          const baseTypeIndex = meta.baseTypes.indexOf(baseMeta);
+          const baseAlias = `b${baseTypeIndex}`;
```
stephenh, line 349: Hey @0ces , thanks for the PR!

I think the wrinkle with my original approach (the commented out line 322, `const maybeBaseAlias = field.alias`) is that it breaks when a table is included in the query twice.

I believe the `mapBaseTypeFields` code is doing what I'd basically done in the `field.alias` query, which is saying:

* We have 3 tables, `LargePublisher`, `SmallPublisher`, and `Publisher`
* We know that we'll always join in `large_publishers` as `lp0` and `small_publishers` as `sp0` and `publishers` as `p0`
* For each of the fields, we'll just set `city.alias = lp0`, `name.alias = p0`
* Then we add `WHERE` clauses for a key, we'll just use its already-known alias i.e. `p0.name = ...` or `lp0.city = ...`

So the ^ logic is what my `field.alias` approach tried, and I believe what your code is doing as well, just at runtime instead of using `field.alias`?

The wrinkle is that sometimes we might be searching against different publishers, like if `Author` had both `authors.first_publisher_id` and `authors.current_publisher_id`, a query could look like:

```
em.find(Authors, { firstPublisher: { name: ... }, currentPublisher: { name: ... } });
```

And now our assumption about `name` is always `p0.name` is incorrect because it actually needs to be `p1` sometimes and `p2` sometimes.

Which is what the current `alias` local variable already handles, but `field.alias` does not--it's too static.

Fwiw I'm pretty sure I realized ^ only after I tried line 322 and it broke just 1 test somewhere; but it looks like your PR is passing all tests, so I'm kinda surprised.

...I dunno, I haven't fully digested the approach in this PR yet, but wdyt of my initial description of the problem?

stephenh: (As a disclaimer, `QueryBuilder` is probably the "most in need of a rewrite" part of Joist, so it is admittedly hard to follow/hard to maintain...it's on the todo list for a face lift/rewrite, just haven't gotten to it...)

packages/integration-tests/src/entities/LargePublisher.test.ts
```diff
@@ -0,0 +1,17 @@
+import { LargePublisher } from "./LargePublisher";
+import { newLargePublisher } from "./LargePublisher.factories";
+import { newPublisherGroup } from "./PublisherGroup.factories";
+import { newSmallPublisher } from "./SmallPublisher.factories";
+
+describe("LargePublisher", () => {
+  it.withCtx("can be find by inherited fields", async (ctx) => {
```
stephenh, line 7: Suggestion: Can we put this in `Inheritance.test.ts`?

We do have a fair amount of tests in like `Author.test.ts` or what not, but lately I've been trying to organize the tests by feature instead of by entity.

So like if we futz with something related to inheritance, we can easily rerun all of the inheritance-related tests.

packages/orm/src/QueryBuilder.ts
stephenh: Just off the top of my head, I think this line:

```
const baseAlias = `b${baseTypeIndex}`;
```

And all of `mapBaseTypeFields` that went in to it _ideally_ could go away and lean on that `field.alias` instead.

Granted, it needs updated to be fancier; like right now `field.alias` is instantiated here:

https://github.com/stephenh/joist-ts/blob/d237a730285b422d4b96f4db7ade8e69f90ef94f/packages/orm/src/index.ts#L278

And the `b0` / `b1` / `s0`/ `s1` convention is a) pretty naive, and b) very hard-coded i.e. in https://github.com/stephenh/joist-ts/blob/d237a730285b422d4b96f4db7ade8e69f90ef94f/packages/orm/src/drivers/PostgresDriver.ts#L600

I think probably the right approach is come up with a more purposeful / less "I just made it up at the time" approach to the aliases that aligns across all three of:

* configureMetadata
* QueryBuilder
* addTablePerClassJoinsAndClassTag

...like we probably need aliases to:

* Always have the entity tag in them, so that if an em.find touches two tables that use inheritance, their `b0`s don't stomp on each other
* Split up the `b0` into more like `a_b_0` / `a_s_1` where `a` is the entity-specific prefix, `b/s` is the base table/sub type, and `0/1` identifies "is this the 1st or 2nd or 3rd time this table/entity is used in the query"

...dunno...that is admittedly quite a bit of work, scattered around the place.

PR #531, opened 2023-02-17
feat: Refactor QueryBuilder to use QueryParser

stephenh:


packages/orm/src/QueryParser.ts
```diff
@@ -45,11 +300,14 @@ export function parseEntityFilter(meta: EntityMetadata<any>, filter: any): Parse
  * The ValueFilter is a
  */
 export type ParsedValueFilter<V> =
-  | { kind: "eq"; value: V | null }
```
stephenh, line 48: It ended up being cleaner to break `null` out as its own `is-null` / `not-null` kinds.

PR #534, opened 2023-02-22
feat: Support filtering on base fields

stephenh:
Previously we joined the data in from base tables into `em.find` clauses (so that `em.hydrate` could load the entity with the data from all of the base+child columns), but we couldn't construct a `WHERE` clause entries for columns from the base table.

This is because they use a slightly different alias, i.e. instead of `sp.name` it needs to be `sb_b0.name`, i.e. the small publishers "first base" table.

My first attempt at this tried to use a hard-coded alias of just `b0` in the `meta.allFields` array, but this was too limited b/c it's possible for tables to be added to a query twice.

A better approach is to use an alias _suffix_ in `meta.allFields`, b/c then we can still use the already-unique `alias` that `QueryParser` is maintaining (and will be unique if a single table does happen to be in a query twice), and so we can just twice the known-unique `alias` + the `_b0` to get "the right" alias into the `WHERE` clause.


stephenh:
@0ces with the refactored `QueryBuilder` / `QueryParser`, and a ~cute approach to the field aliasing (using a suffix that is determined in `configureMetadata` at boot), I was able to implement "find by a sub-type but filter on base-type columns". :tada: :-)

Let me know if it works for you!

PR #540, opened 2023-02-27
feat: Support ands/ors in em.find.

stephenh:


stephenh:
Going to merge as a good start and keep iterating on it.

PR #554, opened 2023-03-04
fix: utils import

0ces:
Hey @stephenh, I was trying your latest change for aliases but I'm getting an error when trying to run the codegen.

image

Already tried this change locally and now the codegen passes.


stephenh:
@0ces shoot! Thank you! Yeah, I don't have build setup to rewrite those `src/` imports, so they leak into the code. I'm really surprised the Joist build process didn't catch this.

Thank you!

PR #558, opened 2023-03-06
fix: Forgot to add o2ms to the GraphQL filters.

stephenh:


stephenh:
cc @0ces if you end up needing this

PR #559, opened 2023-03-06
fix: Allow non-GQL find queries to use pruning.

stephenh:


stephenh:
cc @0ces this might also be useful, i.e. to let you use just `find` and get the query/join pruning feature, without having to use `findGql` per my disclaimers earlier this morning.

PR #593, opened 2023-03-29
fix: Fix DELETEs bumping oplocks for UPDATEs.

### LINKED ISSUE #591: Race condition with `SET NULL` constraints
We can issue SQL where `em.flush`:

1. `DELETE`s an entity like `Author`
   - Another entity like `Book` has `books.author_id` with a `SET NULL` on delete action
   - This automatically sets `books.author_id = NULL` for this `Book`
   - The SQL trigger bumps `books.updated_at` for that `Book`
2. Joist does its own `UPDATE books SET author = null WHERE id = ... AND updated_at = last_updated_at`
   - This fails with an oplock failure b/c the trigger bumped the `updated_at` to not equal our `last_updated_at`

Potential solution is that we issue all `UPDATE`s first, so that they won't be invalidated by `ON DELETE` behavior?

stephenh:
Fixes #591.

### DIFF
--- a/packages/orm/src/drivers/PostgresDriver.ts
+++ b/packages/orm/src/drivers/PostgresDriver.ts
@@ -279,39 +279,46 @@ export class PostgresDriver implements Driver {
     const knex = this.getMaybeInTxnKnex(em);
     const now = getNow();
     await this.idAssigner.assignNewIds(knex, todos);
+
+    // Do a 1st pass for INSERTs+UPDATEs so that we avoid DELETEs triggering cascade
+    // behavior that, in conjunction with our trigger-maintained `updated_at`s, causes
+    // our oplocks to fail. See https://github.com/stephenh/joist-ts/issues/591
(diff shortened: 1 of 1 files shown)


packages/orm/src/drivers/PostgresDriver.ts
```diff
@@ -279,39 +279,46 @@ export class PostgresDriver implements Driver {
     const knex = this.getMaybeInTxnKnex(em);
     const now = getNow();
     await this.idAssigner.assignNewIds(knex, todos);
+
+    // Do a 1st pass for INSERTs+UPDATEs so that we avoid DELETEs triggering cascade
+    // behavior that, in conjunction with our trigger-maintained `updated_at`s, causes
+    // our oplocks to fail. See https://github.com/stephenh/joist-ts/issues/591
     for await (const todo of Object.values(todos)) {
-      if (todo) {
```
stephenh, line 283: I removed this `if (todo)` which made the diff uglier...

PR #600, opened 2023-03-31
feat: use column comments instead of joist-config for relation renames

zgavin:


packages/integration-tests/migrations/1580658856631_author.ts
```diff
@@ -190,6 +181,18 @@ export function up(b: MigrationBuilder): void {
...
 
+  createEntityTable(b, "users", {
+    name: { type: "varchar(255)", notNull: true },
+    email: { type: "varchar(255)", notNull: true },
+    // for testing o2o/m2o renames
+    author_id: foreignKey("authors", {
+      notNull: false,
+      unique: true,
+      oneToOneName: "userOneToOne",
+      referenceName: "authorManyToOne",
```
stephenh, line 192: Thought: I have the terms:

* `fieldName: "authorManyToOne"`, and
* `otherFieldName: "userOneToOne"`

Burned in my head as what I think of these; do you mind renaming?

My thought is that `fieldName` is just a generalization of `referenceName`, and other `otherFieldName` is similarly a generalization of "the name on the other side, regardless of whether the other side is `o2o`/`m2o`".

(Seems nice to have the user not really care whether they need to use `oneToOneName` or `oneToManyName` based on the `unique`-ness of the FK.)

zgavin: This way corresponds to what the codegen calls them though

packages/integration-tests/migrations/1580658856631_author.ts
```diff
@@ -198,6 +201,8 @@ export function up(b: MigrationBuilder): void {
     parent_book_review_id: foreignKey("book_reviews", { notNull: false, unique: true }),
     parent_publisher_id: foreignKey("publishers", { notNull: false }),
     parent_author_id: foreignKey("authors", { notNull: false }),
+    // for testing collection renames
+    user_id: foreignKey("users", { notNull: true, collectionName: "createdComments" }),
```
stephenh, line 205: Nit: Maybe make this `notNull: false`? Would reduce the tests that need to create users just to use comments for their boundary cases.

packages/integration-tests/migrations/1580658856631_author.ts
zgavin: I think i'm about a 5/10 so i'll change it

packages/integration-tests/migrations/1580658856631_author.ts
zgavin: Should we really allow anonymous commenting? seems like a slippery slope towards becoming 4chan 😋

stephenh (approved):
Looks great!

docs/docs/modeling/relations.md
```diff
@@ -124,6 +148,12 @@ To use polymorphic references, there are two steps:
    
    Joist with then use the `publisher` name to scan for any other `publisher_`-prefixed foreign keys and automatically pull them in as components of this polymorphic reference.
 
+## Renaming Relations
+
+Joist will make a best effort to infer a logical name for any given relation based on the foreign key's name and the table it points to, but this is not always perfect.  Sometimes another name will be desired or two foreign keys will generate a naming collision.  In such circumstances, it's possible to provide additional data in the database schema to indicate to joist which name to use.  Joist uses `pg-structure`'s [`commentData`](https://www.pg-structure.com/nav.02.api/classes/dbobject.html#commentdata) to look for the properties `fieldName` (for renaming a m2o reference) and `otherFieldName` (for renaming the opposing m2o/m2m/o2o relation) and will prefer using these values over its own inferred name.
```
stephenh, line 153: Looks good!

PR #603, opened 2023-04-03
chore: Show that Book.sequel works.

stephenh:


stephenh:
Closing out since it was just showing how it worked.

PR #619, opened 2023-04-17
feat: Add hasReactiveAsyncProperty to fix reactivity on properties.

stephenh:
Previously, reactive rules and reactive fields were allowed to depend on async properties.

There were tests against this behavior that seemed reactive, in that they showed reactivity to gross mutations to the object graph, i.e. _anything that was presented in the load hint_.

For example, given an async property such as:

class BookReview {
  readonly isPublic2 = hasAsyncProperty("comment", br =>
    return !br.comment.get?.text.includes("Ignore");
  });
}

Any reactive rules/fields that depended on this `isPublic2` would be recalculated if the `BookReview.comment` field ever changed. Because we passed the load hint along as a reactive hint.

However, any _non-graph-mutating_ changes, such as simple changes to primitive fields that were used within the hasAsyncProperty lambda, would not trigger reactivity.

There were two potential approaches to fix this:

1. Keep the "reactives rules/fields can use async properties w/just load hints" but embellish the load hint with the primitive of every possible field the lamba could read.

I.e. a load hint like `{ book: "review" }` would be embellished to be

```
{ book: { "title": {}, review: { "rating", "otherFieldA", "otherFieldB" } }
```

Even if the lambda only accessed `rating`.

2. Instead of "embellishing load hints into reactive hints", break any usage of a reactive rule/field on a load-hint-using hasAsyncProperty, and force the user to rewrite the property with a new hasReactiveAsyncProperty.

The new hasReactiveAsyncProperty is semantically exactly the same as hasAsyncProperty, except that it takes a reactive hint, and hence it's lambda is forced to specify which fields it actually will access.

Given that hasReactiveAsyncProperty has a user-provided reactive hint, there is no need for "embellishment", and we can hook up reactivity now with assurances that even primitive field changes will trigger the appropriate reactivity.

This PR implements the 2nd option.

### DIFF
--- a/packages/integration-tests/src/EntityManager.reactiveRules.test.tsx
+++ b/packages/integration-tests/src/EntityManager.reactiveRules.test.tsx
@@ -3,6 +3,7 @@ import {
   Book,
   BookReview,
   Color,
+  Comment,
   newAuthor,
   newBook,
   newBookReview,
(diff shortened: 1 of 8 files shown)


zgavin (approved):
our poor tortured toy data model 😂

stephenh:
> our poor tortured toy data model

Lol, yes; this was like my 3rd attempt at "how do I least painfully slice this in as new fields/behavior" :-)

PR #637, opened 2023-05-09
feat: auto wrap `tsvector` columns in `to_tsvector`

blambillotte:


packages/integration-tests/src/entities/Author.test.ts
```diff
@@ -226,6 +226,29 @@ describe("Author", () => {
...
+    // When we create an author with a first and last name
+    const a1 = new Author(em, { firstName: "Bob", lastName: "Smith" });
+    await em.flush();
+    await em.refresh();
+
+    // a1.search.get is returning "Bob Smith BS" instead of "'bob':1 'bs':3 'smith':2"
+    // expect(a1.search.get).toEqual("Bob Smith BS");
+
+    // Then we expect the firstName, lastName and initials to be cast as tsvector lexemes
+    expect(a1.__orm.data.search).toEqual("'bob':1 'bs':3 'smith':2");
```
blambillotte, line 241: Not exactly sure the best way to test this - it's an odd scenario where the format is mutated by the database itself so the string in memory in the EM at the time of flush won't match what's returned. 

I needed to both `refresh()` and use `a1.__orm.data.search` vs `a1.search.get` to observe the string of lexemes vs. the input string 🤔

blambillotte:
@stephenh I have a draft going to get joist to just auto wrap `tsvector` type columns in `to_tsvector` - am I on the right track here?

packages/integration-tests/src/entities/Author.test.ts
stephenh: Ah huh! yeah, this is interesting.

It breaks the current assumption that Joist has, both fundamentally that:

- A primitive field getter will generally always return the `__orm.data` value--granted we do have FKs where that is not true

And:

- If application code happened to load `Author` row, with `search` _not_ loaded, Joist would duly return the `__orm.data` value, at which point we've leaked the lexemes into the application code, which was probably expect to still see the non-lexemed text (who knows why, but just in theory)

I dunno, this sounds kinda odd, but naively is there a way to do the lexeme-ification in javascript? Like have the derived property return `lexeme("field a field b field c")`?

That would preserve that ".get" at least consistently returns lexemes, so there is not a surprise to the user about it "sometimes returns x/sometimes returns y".

...also, I wouldn't be surprised if we're UPDATE-ing the author, with the current PR would we re-tsvector the value on update? Because we'd have `__orm.data.search` as `robert:1`, b/c that's what we'd loaded from the database, if we hadn't explicitly re-calced/loaded the `search` property's load hint? I.e. `.get` if unloaded will fallback on returning the `robert:1` value...

Granted, if `search` is not a changed field, it shouldn't be included in the `UPDATE`, but if we're updating 3 Authors, and one has its searched changed, and 2 had they're first name changed, we do a single UPDATE with the superset of their fields.

If we did the lexeme in JS, it would also this re-lexeme-ification problem...

Generally I would have assumed that lexem-ification was something really complicated and produced esoteric binary data, but since we're asserting it against it here in the test, it doesn't look too bad, and is maybe something we could calc in JS?

Would the postgres update be okay with that? If the field type was `ts_vector` in the db, and we set it to an already-lexme-ified value, vs. specifically going through `to_tsvector`?

stephenh: https://www.npmjs.com/package/to-tsvector :exploding_head:

packages/orm/src/serde.ts
```diff
@@ -49,8 +50,15 @@ export class PrimitiveSerde implements FieldSerde {
     data[this.fieldName] = maybeNullToUndefined(row[this.columnName]);
   }
 
-  dbValue(data: any) {
-    return data[this.fieldName];
+  dbValue(data: any, knex?: Knex) {
```
stephenh, line 53: Totally get why you're doing this, but coupling the serde stuff to a specific driver/connection like knex is unfortunate... It'd be doable for now, but something that we might eventually have to bake into our driver api, like `driver.to_tsvector(...)` so that the serdes could defer to driver-specific/non-knex implementations.

Wouldn't block this initial PR though.

blambillotte: yeah maybe we skip doing this work for now - it's not as clean of a solution as I was hoping for. We could give that js lib a try outside joist 🤔

blambillotte: Actually looking closer at the lib it doesn't look like a real implementation: https://github.com/lblblong/to-tsvector/blob/master/src/index.ts

It's adding the word positions, but not doing any dictionary stemming

stephenh: Shoot! :thinking:

stephenh:
@blambillotte wdyt if we used a two-column approach? Like we'd have both `search` (controlled by Joist, just a regular string) and `ts_search`, where `ts_search` was maintained by a trigger that called `to_tsvector`?

And we could have joist ignore the `to_tsvector` column...

I dunno, just thinking that'd let Joist stay pretty naive about "hey this is just a string" on the primary, and avoid some of the weirdness around "sometimes we need to ts_vector-ize it/sometimes we don't".

Also maybe a dumb question, but given how nicely the "vectorize in JS" was going to work out, is there a 3rd party stemming library that outputs/we could adapt its output to what to_tsvector would generate internally? I.e.

https://github.com/NaturalNode/natural

I did a brief google for "what stemming algorithm does tsvector use?" but didn't find anything in ~1 minute of looking and IANAE a search expert.

stephenh:
Er, I suppose if we ignore the column, then we couldn't use it for searches. :-(

packages/orm/src/serde.ts
stephenh: If we can't end up doing JS-side stemming, I think we'll need to update this code to try and recognize an already-stemmed value, and then _not_ wrap it in `to_tsvector`...

It looks like a pretty recognizable format, but still is a little janky.

blambillotte:
> Er, I suppose if we ignore the column, then we couldn't use it for searches. :-(
> 
> Unless maybe we had an `ignore: "domain"` type flag that would not put the field in the domain model, but would let `em.find` use it for where queries... ...and it'd also have to know how to ts_vector the search criteria? Meh...

yeah so we actually need to do a custom query anyway [like so](https://github.com/homebound-team/graphql-service/pull/3361/files#diff-7cf0a8771b8b0dec0a9bb867bbe10469e92cb520c41960920aab7dc5991133e3R12) so joist-ignoring the extra column would be fine right? 

I'd actually be interested in trying out a [generated column](https://www.postgresql.org/docs/current/ddl-generated-columns.html) for this vs a trigger. Normally you have to use a trigger if you have any joins, but in this case our derived field is taking care of that for us 🤔

stephenh:
I guess that's true re custom SQL query; I was hoping we'd teach `em.find`
to just do something magical if the user did `em.find(Author, { search:
"foo bar" })`, like add in all that plainto_tsquery, but perhaps that's not
realistic / something the user is always going to want to futz with
anyway...

Great point about generated columns! Saw those just once and then forgot
about them. Seems like exactly what we want.

Seems like Joist could learn to support those in general, like treat it as
an always-read-only field... Granted, for immediate purposes, we could
still just ignore it.

Issue #413, opened 2022-09-07
Support relations as persisted derived fields


stephenh:

stephenh:
@chr1sjf0x I did a spike of this in #416 ; I got maybe 80% of the "recalc + save to the db" mechanism done, but then realized that these will be need to be worked into `load` hints.

I.e. I made a `Author.favoriteBook: Book` that is calculated from the `Author.books` collection.

When someone does `em.load(Author, "a:1")`, we know "the favorite book is b:1", but we haven't actually loaded the `Book` entity yet, i.e. just like any other FK. So even though the `PersistedDerivedProperty.get` method is available:

```
const a = await em.load(Author, "a:1");
console.log(a.favoriteBook.get);
```

It prints `"b:1"`. So, unlike primitive `PersistedDerivedProperty`s, I think the user will have to load hint the `favoriteBook`:

```
const a = await em.load(Author, "a:1", "favoriteBook");
console.log(a.favoriteBook.get); // <-- now `Book`
```

Which is interesting b/c `favoriteBook` will now essentially have two different loaded states:

1. Is my last-derived-value loaded, i.e. is the value `b:1` loaded (which needs to be loaded anytime you want to call `.get`)
2. Is my recalc-load-hint loaded, i.e. is `books` loaded so I could re-calc the value (which only needs to be loaded when Joist has detected `favoriteBook` should be recalced) 

So I think we'll probably need some sort of new `PersistedAyncRelation` that both a) hides the `.get` until it's loaded, b) is integrated into `MarkLoaded` similar to `ManyToOneReference`, c) the implementation is a blend of `PersistedAsyncField` and `ManyToOneReference` to balance the dual "value loaded" / "recalc loaded" state.

This all seems doable, but has grown outside of my current/expired timebox. :-) I published #416 with the wip and we can pick it up from there at some point.

stephenh:
This was fixed in #639 :tada:

PR #639, opened 2023-05-12
feat: add `PersistedAsyncReference`

0ces:
Attempt to solve #413


local.env
```diff
@@ -0,0 +1 @@
+DATABASE_URL=postgres://joist:local@localhost:5435/joist
```
0ces, line 1: Added this to able to run a jest test directly from the `Run` button in VSCode

packages/orm/src/relations/hasPersistedAsyncRelation.ts
0ces, line 1: This is a mix between `PersistedAsyncProperty` and `ManyToOneReference`, some methods might not be needed (WIP)

stephenh:
This looks like a great start @0ces ! I haven't had a chance to really dig in yet, but AFAICT you're updating all the right places. And, you're right, `hasPersistedAsyncRelation` will be a mix `PersistedAsyncProperty` and `ManyToOneReference`. 

Just scanning the build, it seems like a lot of tests failed...have you figured out why?

I'll have to think about "how can `PersistedAsyncProperty` and `ManyToOneReference`" be combined ... there is probably a lot of code in `ManyToOneReference` that deals with keeping "the other side" of the m2o up to date, i.e. the o2m collection.

I guess all of that same infra makes sense, for `hasPersistedAsyncRelation`, the only difference is whether (with non-persisted async FKs) the user explicitly mutates the reference, or (with persisted async FK) Joist mutates the reference when it knows its changed... :thinking:

0ces:
Hi @stephenh, @blambillotte helped me to look at my problem with the generic type issue and he saw the you ran into this issue before: https://github.com/microsoft/TypeScript/issues/47213
image

image

image

I wonder if you remember how to address this issue 😣

0ces:
> Hi @stephenh, @blambillotte helped me to look at my problem with the generic type issue and he saw that you ran into this issue before: [microsoft/TypeScript#47213](https://github.com/microsoft/TypeScript/issues/47213) image
> 
> image
> 
> image
> 
> I wonder if you remember how to address this issue 😣

I think I got it, if I pass the otherEntity Metadata then TS correctly infers `U` as `Book` and doesn't throw the error.

image

packages/integration-tests/src/EntityManager.reactiveRules.test.tsx
```diff
@@ -178,16 +178,41 @@ describe("EntityManager.reactiveRules", () => {
...
   });
 
+  it.withCtx("updates derived fields that are foreign keys", async ({ em }) => {
+    // Given an entity with an async derived field
+    const a = newAuthor(em);
+    const b = newBook(em, { author: a, reviews: [{ rating: 10 }] });
+    await em.flush();
+
+    const favoriteBook = a.favoriteBook.get;
+    // For some reason .toMatchEntity(b) fails here
```
0ces, line 188: I wasn't able to get `.toMatchEntity` to work, it will keep trying to match Collections to Arrays.

packages/codegen/src/EntityDbMetadata.ts
```diff
@@ -132,6 +132,7 @@ export type ManyToOneField = Field & {
   otherFieldName: string;
   otherEntity: Entity;
   notNull: boolean;
+  derived: "orm" | "async" | false;
```
stephenh, line 135: Suggestion: I think we can drop `"orm"` for now

packages/codegen/src/generateEntityCodegenFile.ts
```diff
@@ -225,6 +226,13 @@ export function generateEntityCodegenFile(config: Config, dbMeta: DbMetadata, me
   const m2o = meta.manyToOnes.map((m2o) => {
     const { fieldName, otherEntity, otherFieldName, notNull } = m2o;
     const maybeOptional = notNull ? "never" : "undefined";
+
+    if (m2o.derived === "async") {
+      return code`
+        abstract readonly ${fieldName}: ${PersistedAsyncRelation}<${entity.name}, ${otherEntity.type}, ${maybeOptional}>;
```
stephenh, line 232: Suggestion: Let's call `PersistedAsyncRelation` -> `PersistedAsyncReference`

packages/codegen/src/generateMetadataFile.ts
```diff
@@ -118,14 +118,15 @@ function generateFields(config: Config, dbMetadata: EntityDbMetadata): Record<st
...
+    const { fieldName, columnName, notNull, otherEntity, otherFieldName, derived, dbType } = m2o;
     const otherTagName = config.entities[otherEntity.name].tag;
     fields[fieldName] = code`
       {
         kind: "m2o",
         fieldName: "${fieldName}",
         fieldIdName: "${fieldName}Id",
-        required: ${notNull},
+        derived: ${!derived ? false : `"${derived}"`},
+        required: ${!derived && notNull},
```
stephenh, line 129: Thought: Could we leave this as `required: ${notNull}`?

packages/integration-tests/src/EntityManager.reactiveRules.test.tsx
```diff
@@ -178,16 +178,41 @@ describe("EntityManager.reactiveRules", () => {
...
+    expect(a.favoriteBook.get).toBe(b);
+
+    // If there is a new favorite book
+    const b2 = newBook(em, { author: a, reviews: [{ rating: 20 }] });
+    await em.flush();
+    // Then the derived field is updated
+    expect(a.favoriteBook.get).toBe(b2);
```
stephenh, line 195: Suggestion: This is checking the entities internal data and not necessarily what is literally in the database, so maybe add an assertion that calls `select("authors")` and asserts against the `favorite_book_id` column.

packages/integration-tests/src/entities/AuthorCodegen.ts
```diff
@@ -108,6 +110,7 @@ export interface AuthorOpts {
   favoriteShape?: FavoriteShape | null;
   mentor?: Author | AuthorId | null;
   currentDraftBook?: Book | BookId | null;
+  favoriteBook?: Book | BookId | null;
```
stephenh, line 113: Suggestion: Let's update `generateEntityCodegenFile` to not output this since users don't need to pass it in.

packages/orm/src/EntityMetadata.ts
```diff
@@ -113,6 +113,7 @@ export type ManyToOneField = {
   otherFieldName: string;
   serde: FieldSerde;
   immutable: boolean;
+  derived: "orm" | "async" | false;
```
stephenh, line 116: Nit: same remove `"orm"` from here

packages/orm/src/relations/hasPersistedAsyncRelation.ts
```diff
@@ -0,0 +1,342 @@
...
+  isSet: boolean;
+
+  /**
+   * Returns the as-of-last-flush previously-calculated entity.
+   *
+   * This is useful if you have to purposefully avoid using the lambda to calc the latest entity,
+   * i.e. if you're in a test and want to watch a calculated entity change from some dummy entity
+   * to the new derived entity.
+   * */
+  fieldValue: U | N;
```
stephenh, line 23: Thought: Maybe we can skip this--this does actually match PersistedAsyncProperty, so lets keep for now.

packages/orm/src/relations/hasPersistedAsyncRelation.ts
```diff
@@ -0,0 +1,342 @@
...
+  get get(): U | N {
+    return this.doGet({ withDeleted: false });
+  }
+
+  get isLoaded(): boolean {
+    return this._isLoaded;
+  }
+
+  set(other: U | N): void {
+    this.setImpl(other);
```
stephenh, line 132: Thought: This should probably just blow up/fail and say it can't be set.

packages/orm/src/relations/hasPersistedAsyncRelation.ts
```diff
@@ -0,0 +1,342 @@
...
+  }
+
+  get isSet(): boolean {
+    return this.current() !== undefined;
+  }
+
+  // Internal method used by OneToManyCollection
+  setImpl(other: U | IdOf<U> | N): void {
```
stephenh, line 140: Musing: Seems like callers doing `o2m.add/remove` should be unsupported; I wonder if eventually we have a `ReadOnlyCollection` / `DerivedOneToManyCollection` that does not have `add` / `remove` methods.

stephenh: What happens if:

1. `Author.favoriteBook = b1`
2. Someone loads `b1.favoritedAuthors.load()`, returns `[a1]`
3. Someone creates `b2` with a higher rating
   * There are no obvious setter/add calls for us to know to update `b1.favoritedAuthors`...
5. `a1.favoriteBook.get` will immediately return `b2` (great)
6. Ideally `b1.favoritedAuthors.get` would now return `[]`

We don't need to solve this for this PR, but the ergonomics of "always up-to-date both sides of relations" is an important feature of joist.

packages/orm/src/relations/hasPersistedAsyncRelation.ts
```diff
@@ -0,0 +1,342 @@
...
+        throw new Error(`Unhandled ${o2m}`);
+      }
+    }
+    setField(this.#entity, this.fieldName, undefined);
+    this.loaded = undefined as any;
+    this._isLoaded = true;
+  }
+
+  maybeRemove(other: U | undefined) {
+    if (other) {
```
stephenh, line 241: Thought: Maybe these go away? ... or maybe we just don't generate `Book.favoritedAuthors` for now until we can put more time into making it super amazing. Or maybe we never need it.

packages/integration-tests/src/EntityManager.reactiveRules.test.tsx
```diff
@@ -178,16 +178,44 @@ describe("EntityManager.reactiveRules", () => {
...
 
+  it.withCtx("updates derived fields that are foreign keys", async ({ em }) => {
+    // Given an entity with an async derived field
+    const a = newAuthor(em);
+    const b = newBook(em, { author: a, reviews: [{ rating: 10 }] });
+    await em.flush();
+
+    // For some reason .toMatchEntity(b) fails here
+    // expect(a.favoriteBook.get).toMatchEntity(b);
+    expect(a.favoriteBook.get).toBe(b);
```
stephenh, line 189: Suggestion: besides asserting against the entity getter, can you use the select helper to assert again the fk in the data base as well?

Oh! I see you added an assertion for that later in the test; can we do a db-based assertion after each em.flush? Just because this is for a derived value.

stephenh (approved):
This is great! Thanks @0ces

packages/integration-tests/src/EntityManager.reactiveRules.test.tsx
0ces: Ok, I moved the `select` assert before the `.get` assert.

0ces:
I'll finish my [ticket](https://app.shortcut.com/homebound-team/story/33711/add-server-side-orderby-and-pagination-for-projects-query-endpoint) first to test everything locally before merging this PR

stephenh:
@0ces / @blambillotte I was chatting with @zgavin and realized that maybe (not for this PR, but in general/someday) it seems like Derived FKs potentially have two different "loaded" states, like:

1) is that I know `author.favoriteBook` is `b:2`, just load that into memory (like a regular m2o.load)
2) is that I want to actually recalc `author.favoriteBook` and so load the entire reactive hint and calc the latest value

I think for now / this PR / your projects dashboard use case, the differentiation of these two probably doesn't matter, but, dunno, seems like eventually it'd be nice to teach `PersistedAsyncRelation` how to differentiate the two.

I haven't really thought of the best API for doing so; although maybe it's as simple as `.load()` always uses the previously-calc-d value, and we'd only load the reactive hint when a) the entity is new, or b) the reactivity mechanism is explicitly re-calc-ing the value.

PR #651, opened 2023-05-19
feat: Add em.findCount.

stephenh:


packages/orm/src/QueryParser.ts
```diff
@@ -357,10 +357,12 @@ export function parseFindQuery(
   } else {
     maybeAddOrderBy(query, meta, alias);
   }
-
-  if (orderBys.length > 0) {
-    Object.assign(query, { orderBys });
+  // Even if they already added orders, add id as the last one to get deterministic output
```
stephenh, line 360: This used to be done in `buildKnexQuery`, but `select count(*)`s shouldn't have it, so moved it here.

PR #655, opened 2023-05-19
fix: Fix gatherEntities looping on cycles.

stephenh:
Granted, we're assuming POJOs won't have cycles (vs. a true already visited cache), but that seems fine for now.


zgavin (approved):

zgavin:
maybe add a test?

stephenh:
Yep, added a test

PR #676, opened 2023-06-05
feat: Add custom type override to config & codegen

brudil:
Following on from the discussion in https://github.com/stephenh/joist-ts/discussions/674

- Adds optional `type` to `FieldConfig`
- Presence overrides `maybeUserType` - I've left `superstruct` to take precidence over `type` if both are specified
- Tried to follow the same naming as superstruct, but clashes with fieldType has lead to clunky naming such as `userFieldTypeType` - happy to change
- Added ip_address to intergration test migration and usage in `packages/integration-tests/src/resolvers/mutations/user/saveUserResolver.test.ts` as it's required. I noted the correct changes in UserCodegen but didn't see any obvious tests around the codegen'd output itself. Let me know if I've missed something :)
- Added a User.test.ts with an `@ts-expect-error` around passing in a primitive
- Attempted to run `yarn format` but it changed a ton of files I haven't touched so I've left those out for now (mostly import reordering).

I'll leave it for another PR, but I noticed the split on `@` within `superstructType` (and now `userFieldTypeType`) breaks if you give namespaced package a la `@company/common`. Can be fixed by joining the string after the first `@`


packages/codegen/src/EntityDbMetadata.ts
```diff
@@ -700,3 +718,9 @@ function superstructType(s: string): Import {
   const [symbol, path] = s.split("@");
   return Import.from(`${pascalCase(symbol)}@${path}`);
 }
+
+function userFieldTypeType(s: string): Import {
+  // Assume it's `foo@...`, turn it into `Foo@...`
+  const [symbol, path] = s.split("@");
+  return Import.from(`${pascalCase(symbol)}@${path}`);
```
stephenh, line 725: I think we could probably skip this `pascalCase` / "assume it's foo -> Foo" step--that was primarily for superstruct, where there is both a runtime const `ipAddress` as well as type system `IpAddress`, and we wanted to have one config field that pointed to both.

For the `fieldType`, I think it'll always just be the type, like as you've got in the test `type: "IpAddress@..."`, so we can expect the user to have already camel-cased it/be pointing to a type.

stephenh (commented):
This looks great! Few small suggestions, but should be quick.

Per the format/import wondering, I'm suspecting that I recently bumped prettier/prettier-organize-imports and didn't reformat the codebase, which is my bad, I'll look in to that. But yeah, great to leave out of your PR.

brudil:
I've (hopefully) addressed all the feedback. Cheers for the speedy review :)

stephenh (approved):
Looks great @brudil , thanks for the PR!

PR #677, opened 2023-06-05
fix: Add support namespaced packages within config imports

brudil:
~~I've noticed this isn't supported downstream in ts-poet. I'll leave this as a draft revisit this once I've had a look at that. This is assuming a double `@@` is the way to go.~~

----
As mentioned in #676, the import logic of `superstructType` and `userFieldTypeType` splits on `@` and uses the first two parts. This causes issues when importing from a namespaced package.

```json
{ "fields": { "address": { "superstruct": "address@@company/common" } } }
```

results in the reconstructed import passed to ts-poet `address@`.

This PR rests all of the following parts and joins them back up with an `@`.

### DIFF
--- a/packages/codegen/src/EntityDbMetadata.ts
+++ b/packages/codegen/src/EntityDbMetadata.ts
@@ -720,6 +720,5 @@ function superstructType(s: string): Import {
 }
 
 function userFieldTypeType(s: string): Import {
-  const [symbol, ...path] = s.split("@");
-  return Import.from(`${symbol}@${path.join('@')}`);
+  return Import.from(s);
 }
--- a/packages/codegen/src/EntityDbMetadata.ts
+++ b/packages/codegen/src/EntityDbMetadata.ts
@@ -715,11 +715,11 @@ function isPgEnum(c: Column): boolean {
 
 function superstructType(s: string): Import {
   // Assume it's `foo@...`, turn it into `Foo@...`
-  const [symbol, path] = s.split("@");
-  return Import.from(`${pascalCase(symbol)}@${path}`);
+  const [symbol, ...path] = s.split("@");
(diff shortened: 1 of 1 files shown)


stephenh:
Oh yeah, that is weird--ts-poet copied a lot of its import syntax from this other code generation project, but yeah in retrospect the `file@module` does clash with the `@module` convention.

My guess is that the modules-starting-with-`@` wasn't very standard back then, when `@` was picked for the import spec format...

Agreed maybe just split instead of `@`...fwiw now that we're not camel-casing things for superstruct, I almost wonder if we could hand the import spec off to ts-poet, and it would handle it internally.

brudil:
The mystery deepens a little bit - after adding a test (https://github.com/stephenh/ts-poet/pull/47) - it looks like ts-poet does support the double `@` correctly. So I'll need to do a little more digging as to why it's breaking the output of the codegen within Joist.

brudil:
Apologies - completely on me. ts-poet is fine, happy to close the linked PR there if that test isn't desired.

Blowing away the typescript build dir and rebuilding fixed it; looks it wasn't running these changes properly.

packages/codegen/src/EntityDbMetadata.ts
```diff
@@ -715,11 +715,10 @@ function isPgEnum(c: Column): boolean {
 
 function superstructType(s: string): Import {
   // Assume it's `foo@...`, turn it into `Foo@...`
-  const [symbol, path] = s.split("@");
-  return Import.from(`${pascalCase(symbol)}@${path}`);
+  const [symbol, ...path] = s.split("@");
```
stephenh, line 718: Ah nice, thanks for fixing the `superstructType` as well!

stephenh (approved):

stephenh:
Great, thanks @brudil !

PR #678, opened 2023-06-06
feat: refactor serde logic for future exposure

brudil:
You're probably sick of me by now :)

I've taken an initial stab at the direction from our discussion around exposing serdes to the config in the future. This PR doesn't expose any API yet, but hopefully gets us closer to bikeshedding on naming and all that stuff.

I haven't tided this up, but wanted to get it up for feedback on this direction.

----

As I see it the main issue with genericising the existing independent constructors was ensuring type-safety across the codegen'd boundary - not all field serde opts are available to all field types - don't want to be attempting to use a `superstruct` option within a Serde class if it's not assigned to a superstruct field etc. We could make the opts optional but this would need handling in each implemention at runtime.

To this end, within `serde.ts` I've split the options across specific field types - extending from `BaseFieldSerdeOpts`.

```ts

interface BaseFieldSerdeOpts {
  fieldName: string;
  columnName: string;
  dbType: string;
  tagName: string;
}

interface EnumFieldSerdeOpts extends BaseFieldSerdeOpts {
  enumObject: any;
}

// and so on
```

To reduce boilerplate, I've refactored FieldSerde to a class that should be extended from by any implementation. This accepts a generic with the field type, defaulting to `BaseFieldSerdeOpts`

```ts
// defaults to BaseFieldSerdeOpts
export class PasswordFieldSerde extends FieldSerde { ...impl }

// has access to `enumObject`
export class EnumFieldSerde extends FieldSerde<EnumFieldSerdeOpts> { ...impl }
```

The opts live on `this.opts` within the instance and are protected. To ensure "using the instance as the Column" still works, the base FieldSerde implements getters for `dbType` and `columnType`.

With future exposure in mind, this means if you attempted to configure the above a `FieldSerde<EnumFieldSerdeOpts>` class on a primitive field, it would have a type error within the codegenerated `metadata.ts`.

Within `generateMetadataFile.ts` we generate the BaseFieldSerdeOpts + any specifc for that field type.

For polymorphic fields, which do not have a `columnType` or `dbType`, rather than introduce a seperate path (not inheriting from BaseFieldSerdeOpts and a seperate PolymorphicFieldSerde class to extend from), I've used `as never` on the aformentioned opts. Perhaps this is a bit iffy. But it doesn't weaken the type-safety, isn't exposed to consumers of FieldSerde and does significantly simplify things.

If this direction sounds good, we should be able to follow this up by exposing the FieldSerde class and opts, adding the option to FieldConfig and, if provided, using it instead of the default within `generateMetadataFile.ts`


packages/codegen/src/generateMetadataFile.ts
```diff
@@ -59,59 +73,64 @@ function generateFields(config: Config, dbMetadata: EntityDbMetadata): Record<st
...
+  dbMetadata.primitives.forEach((field) => {
+    const { fieldName, derived, columnName, columnType, superstruct } = field;
     const serdeType = superstruct
-      ? code`new ${SuperstructSerde}("${fieldName}", "${columnName}", ${superstruct})`
+      ? code`new ${SuperstructSerde}({ ${generateBaseFieldSerdeOpts(dbMetadata, field)} superstruct: ${superstruct} })`
       : columnType === "numeric"
-      ? code`new ${DecimalToNumberSerde}("${fieldName}", "${columnName}")`
-      : code`new ${PrimitiveSerde}("${fieldName}", "${columnName}", "${columnType}")`;
+      ? code`new ${DecimalToNumberSerde}({ ${generateBaseFieldSerdeOpts(dbMetadata, field)} })`
+      : code`new ${PrimitiveSerde}({ ${generateBaseFieldSerdeOpts(dbMetadata, field)} })`;
```
stephenh, line 92: Hey @brudil ! Sorry for the delay, finally got to look at this and think a bit...

These changes stemming from my initial "opt-ize" idea all look good--but after seeing them play out with the existing serdes (particularly the KeySerde, which wants different tag names based on usage), I'm wondering if we could ever really guess all of any given serdes' necessary params/opts (b/c it's awkward enough for the existing ones, let alone future ones).

Which, if we can't do that, maybe we need to try and not "guess all necessary opts", but instead just provide a hook to let the user, at codegen time, include in the codegen output whatever they want.

Like besides the _runtime_ serde impl, let the user provide _codegen_ time customization. Something like:

```json
{
  "entities": {
    "Author": {
      "fields": {
         "ipAddress": { "serdeGen": "generateIpSerde@src/codegen.ts" }
...
```

Where `generateIpSerde` would be called at codegen time, and passed the `dbMetadata` and `field` (which in theory would have all metadata the user could ever want, to include in their generated constructor call), i.e.:

```
function generateIpSerde(dbMetadata: ..., field: ...) {
  return imp(`new ${IpSerde}(...whatever params necessary...);
}
```

This would let users completely control their serde's constructor parameters.

_But_ we probably don't need to start there.

Like maybe instead of jumping to a `serdeGen` config flag, we start with just `serde: IpSerde@src/serde.ts` with the API contract being that your serde constructor accepts just fieldName, columnName, dbType (as regular parameters, not opts-style):

```ts
export class IpAddressSerde implements Serde {
  constructor(fieldName: ..., columnName: ..., dbType: ...) {
}
```

Which is:

a) really simple for the common case of "this serde doesn't need any custom opts",

b) we could back away from my initial opts-ize the serde approach since, after flushing it out a bit, it seems obvious in retrospect that fitting all serde constructors into a single box/set-of-opts is unrealistic, _but_

c) with the `serdeGen` idea in our back pocket, we've got some future proofing for, should users ever come back with "the field name + column name + db type API for my serde is not good enough", we can then implement `serdeGen` to give them exact control over the code that calls their serde constructor.

So, apologies for the approach concern, but I think this would be better/simpler for most custom serdes, and it'd leave the internal serdes generally alone/as-is.

Wdyt? Do you mind trying to implement ^?

(And fwiw given I'm more up-to-speed on the problem/approach now, I shouldn't take as long on the next PR review :pray: --thank you!).

brudil:
Cheers for the review! I get the concern with the current approach. 

The serdeGen concept sounds nice! Is there any existing systems like this in joist? I did notice some mentions of plugins but nothing in the docs. Just curious because we're going to need a way to transpile referenced ts files before executing them etc.

This could be quite powerful in other ways, I’d like to contribute zod support to go alongside superstruct (we’re currently needing to duplicate a lot of our existing zod schemas just for joist) but I’m wondering if this could technically sit outside of joist. As I see superstruct currently goes beyond just the serde impl and also tweaks the setters too. Perhaps codegen serde could have a part to play there?

Happy to go forward with the simplified approach for now, I’ll look to get a PR up tomorrow!

stephenh:
> The serdeGen concept sounds nice! Is there any existing systems like this in joist?
> I did notice some mentions of plugins but nothing in the docs.

Oh yeah, I kinda forgot about the plugins--we use this graphql plugin to scaffold out types for our GraphQL backend, but kinda cheat by just distributing it with Joist directly.

> we're going to need a way to transpile referenced ts files before executing them etc

Good point; lately I execute so much through tsx but that I hand-waved "require a *.ts file", but you're right it'd be a little different.

So, yeah, the `serdeGen` concept would definitely have a few novel aspects to it--but also was kinda assuming/hoping we wouldn't need it just yet, for the first few serde impls?

> I’d like to contribute zod support to go alongside superstruct 

Ah great, would love to have zod support. Fwiw I think if we were choosing from scratch, we'd probably end up choosing zod this time around. Seems like it's one of the leaders in that space these days.

Having you submit support for it would be amazing; I suppose I am torn between doing the GSD approach of just hard-coding support for zod, similar to what we did for superstruct, or taking a more deliberate approach and trying to get zod support via a userland plugin...

The entgo library (a Golang ORM) has implemented several features (i.e. soft deletes) via generic plugin APIs that, on the Joist side, I've been just pragmatically hard-coding directly into the codebase. Not sure if zod support would be the time to revisit/tackle that or not, but it'd be an option.

brudil:
If you’re happy for it to be part of joist for the time being, I’ll get on that as a PR shortly. 

Certainly simplified things, and serdeGen can stay as an “later” idea.

And I suppose if in the future it were desired stuff like zod/superstruct were moved in to user-land, it could be done with minimal changes for users.

stephenh:
> If you’re happy for it to be part of joist for the time being,
> I’ll get on that as a PR shortly.

Great!

> And I suppose if in the future it were desired stuff like zod/superstruct
> were moved in to user-land, it could be done with minimal changes for users.

Yep, that's exactly my thought. I have a few tickets floating around to "user land the soft delete implementation", etc., but I'm not in a big rush to tackle them.

Usually with Joist I back burner things for awhile, until the combination of an obvious implementation + a significant itch/need both coincide/present themselves at the same time, and then knock it out. :-)

brudil:
Superseeded by #698

PR #679, opened 2023-06-06
fix: ignore PersistedAsyncReference on em refresh

0ces:
image
image

### DIFF
--- a/packages/orm/src/EntityManager.ts
+++ b/packages/orm/src/EntityManager.ts
@@ -1,6 +1,8 @@
 import { AsyncLocalStorage } from "async_hooks";
 import DataLoader, { BatchLoadFn, Options } from "dataloader";
 import { Knex } from "knex";
+import { Entity, isEntity } from "./Entity";
+import { Todo, combineJoinRows, createTodos, getTodo } from "./Todo";
 import { constraintNameToValidationError } from "./config";
 import { createOrUpdatePartial } from "./createOrUpdatePartial";
 import { findByUniqueDataLoader } from "./dataloaders/findByUniqueDataLoader";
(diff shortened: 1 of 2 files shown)


packages/orm/src/EntityManager.ts
```diff
@@ -23,34 +22,35 @@ import {
...
-  ValidationRuleResult,
 } from "./index";
-import { Loaded, LoadHint, NestedLoadHint, New, RelationsIn } from "./loadHints";
+import { LoadHint, Loaded, NestedLoadHint, New, RelationsIn } from "./loadHints";
 import { normalizeHint } from "./normalizeHints";
 import { ManyToOneReferenceImpl, OneToOneReferenceImpl, PersistedAsyncReferenceImpl } from "./relations";
 import { JoinRow } from "./relations/ManyToManyCollection";
-import { combineJoinRows, createTodos, getTodo, Todo } from "./Todo";
-import { assertNever, fail, getOrSet, MaybePromise, toArray } from "./utils";
+import { MaybePromise, assertNever, fail, getOrSet, toArray } from "./utils";
```
0ces, line 53: All these import changes were caused by prettier

packages/orm/src/EntityManager.ts
```diff
@@ -1225,14 +1225,21 @@ export class EntityManager<C = unknown> {
...
             // custom load function + any other collections they call.
-            .filter((r) => !(r instanceof CustomCollection || r instanceof CustomReference))
+            .filter(
+              (r) =>
+                !(
+                  r instanceof CustomCollection ||
+                  r instanceof CustomReference ||
+                  r instanceof PersistedAsyncReferenceImpl
+                ),
+            )
```
0ces, line 1242: The comment above also applies to `PersistedAsyncReference` when `em.refresh` is called it doesn't propagate properly and ends up throwing a `... is not loaded`

stephenh: @0ces I think this makes sense...can you add a unit test in `EntityManager.test` (it has some tests for `refresh`) to make sure this doesn't regress?

Issue #684, opened 2023-06-10
jsonb columns don't support top level array


brudil:
Attempting to persist an array to a jsonb column results in `invalid input syntax for type json array`.

I've got a repro with some tests in this branch https://github.com/stephenh/joist-ts/commit/fea62a60107e8f8b335fbb9e4440195fb5097b65

Digging in to this, it looks like node-postgres assumes arrays are native pg arrays: https://github.com/brianc/node-postgres/issues/442

stringifying the array beforehand works, but as far as I can tell, joist passes the data directly.

I've got a basic workaround to validate my suspicions by doing a JSON.stringify within SuperstructSerdes (https://github.com/stephenh/joist-ts/commit/5dce2a3739a840cdf35640af707ffd875a65500e)

Although I'm questioning if this is the right place for it, for example using a type config instead if superstruct would not work. Perhaps if #678 and exposure land, it could be a user-land concern. But with Joist's runtime metadata of column types, it feels like there might be a better place (after serde) to do this before handing off to pg?

stephenh:
Huh! Great find on that node-pg issue.

> using a type config instead if superstruct would not work

Do you mean like if the user is using a `jsonb` column _without_ using the `superstruct` config option? If so, agreed we should handle that, maybe via just a `JsonSerde` that we use for non-superstruct-configured jsonb columns?

Fwiw I think your changes to `SuperstructSerde` make a lot of sense / would be the rest place to do this, and the test case `quotes` / `string[]` is great.

If you want to turn that branch into a PR, that either handles this case for _just_ superstruct-configured fields or both superstruct-configured and superstruct-less (by introducing an additional JsonSerde), that'd be great! Thanks!

brudil:
Will do!

PR #685, opened 2023-06-12
fix: jsonb top level array

brudil:
As discussed in #684, stringifies within SuperstructSerde and a newly introduced JsonSerde.

Will see how this behaves in ~~GHA~~ circleci, but locally I'm getting

```
● EntityManager › findOrCreate resolves dups with different where clauses in a loop

     expect(received).toBe(expected) // Object.is equality

     Expected: "C"
     Received: "B"

       891 |     expect(a1).toEqual(a2);
       892 |     // And the last upsert wins
     > 893 |     expect(a1.lastName).toBe("C");
           |                         ^
       894 |   });
       895 |
       896 |   it("can find and populate with findOrCreate", async () => {

       at Object.toBe (src/EntityManager.test.ts:893:25)
```

which is curious because I'm not sure where my changes are coming in to play here.

### DIFF
--- a/packages/codegen/src/generateMetadataFile.ts
+++ b/packages/codegen/src/generateMetadataFile.ts
@@ -11,6 +11,7 @@ import {
   PolymorphicKeySerde,
   PrimitiveSerde,
   SuperstructSerde,
+  JsonSerde
 } from "./symbols";
 import { q } from "./utils";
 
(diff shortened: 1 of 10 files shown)


brudil:
v20.3.0, it's behaving fine now. `yarn build && make db && yarn test` should be enough to clean slate it right?

I've definitely hit it a few times, usually coincides with another hitting the 5000ms jest time out. I do wonder if it's when resources are constrained, low battery throttling. I'll keep an eye out.

PR #686, opened 2023-06-12
feat: support zodSchema for jsonb columns

brudil:
Getting up an initial version. This was actually a little more invovled than I was thinking it would be. Mainly down to how Zod schemas coerce and transform, meaning that the input and output type can differ, because of this Zod has a `z.input<typeof Schema>` and a `z.output<typeof Schema>` which should be used on the setters and getters respectively.

I've defaulted to `z.infer<>` as the fieldType and just overriden it within the setters and getters, this has kept the complexity down and would be the equlivent of the superstruct support. It differs from superstruct by only expecting the schema referenced in config, and creating the type within zodSchemaType via infer. The pascal/camel case for schema and type is not a pattern I think Zod uses as much, so this simplies usage for users a bit further.

- new FieldConfig option: `zodSchema`
- Supports Zod transformations within the setter, with correct types on get and set
- Includes serialisation for ZodSerde as from #685


packages/codegen/src/EntityDbMetadata.ts
```diff
@@ -1,7 +1,7 @@
 import { camelCase, pascalCase, snakeCase } from "change-case";
 import { Column, EnumType, Index, JSONData, M2MRelation, M2ORelation, O2MRelation, Table } from "pg-structure";
 import { plural, singular } from "pluralize";
-import { Import, imp } from "ts-poet";
+import {Import, imp, code, Code} from "ts-poet";
```
stephenh, line 4: Nit: Can you add spaces around the curly brackets?

packages/codegen/src/EntityDbMetadata.ts
```diff
@@ -731,6 +739,12 @@ function superstructType(s: string): Import {
   return Import.from(`${pascalCase(symbol)}@${path.join("@")}`);
 }
 
+function zodSchemaType(s: string): Code {
+  const schema = Import.from(s);
+  const zod = Import.from(`z@zod`);
+  return code`${zod}.infer<typeof ${schema}>`;
```
stephenh, line 745: I saw your PR description re `z.output` / `z.input`--the dual input/output types are neat.

Just curious, but afaiu isn't `z.infer` just an alias for `z.output`? If so, could we use `z.output` here just for simplicity/consistency with the getter?

packages/codegen/src/generateEntityCodegenFile.ts
```diff
@@ -50,7 +50,7 @@ import {
   newChangesProxy,
   newRequiredRule,
   setField,
-  setOpts,
+  setOpts, Zod,
```
stephenh, line 53: Nit: Should be on a new line

packages/codegen/src/EntityDbMetadata.ts
```diff
@@ -731,6 +739,12 @@ function superstructType(s: string): Import {
   return Import.from(`${pascalCase(symbol)}@${path.join("@")}`);
 }
 
+function zodSchemaType(s: string): Code {
+  const schema = Import.from(s);
+  const zod = Import.from(`z@zod`);
```
stephenh, line 744: Really minor, but could use the `Zod` symbol here.

packages/graphql-codegen/src/graphqlUtils.ts
```diff
@@ -10,7 +10,7 @@ import {
 import { PrimitiveField, PrimitiveTypescriptType } from "joist-codegen";
 import { groupBy } from "joist-utils";
 import prettier, { Options, resolveConfig } from "prettier";
-import { Import } from "ts-poet";
+import {Code, Import} from "ts-poet";
```
stephenh, line 13: Nit: import format

packages/integration-tests/src/EntityManager.test.ts
```diff
@@ -1306,22 +1307,64 @@ describe("EntityManager", () => {
...
+          "received": "undefined",
+          "path": [
+            "street"
+          ],
+          "message": "Required"
+        }
+      ], undefined, 2));
+    });
+
+    it("rejects reading invalid zodSchema values", async () => {
```
stephenh, line 1350: Great tests!

packages/integration-tests/src/entities/types.ts
```diff
@@ -7,6 +8,10 @@ export const address = object({
   street: string(),
 });
 
+export const AddressSchema = z.object({
+  street: z.string(),
```
stephenh, line 12: Doesn't have to be for this PR, but might be nice to have a schema/integration test that explicitly cover a schema with different input & output types.

stephenh (approved):
This looks great!

Reading about the input type vs output type, I'm remembering one of our complaints with superstruct, was both this lack of separate input/output types, but more specifically "two-way serde", which afaict zod is also missing, e.. the ability to go "back" to the input type.

Granted, I think what we want is actually a different use case than "input type/output type", but what we've wanted to do is:

a) have dates represented in the superstruct/zod schema as `Date`s (fine),
b) accept them on the wire as ISO date strings (a coerced input type, so np),
c) store them in the db as strings b/c jsonb doesn't support Dates / other user-land types (again fine, handled by JSON.stringify), but then
d) load the JSON back from the db and get our Dates back (not possible afaiu).

Basically more like a two-way "serialization schema -> json / deserialization json -> schema". Afaict zod doesn't have this?

I've looked at a few other runtime type libraries and afaict most/none of them have it either, which makes me curious why not...are we missing some easy way of doing jsonb ISO date strings back to `Date`s (and not just for `Date`, but generalized to other user land types)? 

I had not found it yet, and was worried Joist would eventually need to grow it's own "json _serde_ library", which of course sounds terrible.

That musing aside, I had a few minor spacing/formatting nits, but otherwise feel free to merge when you're ready. Thanks!

packages/graphql-codegen/src/graphqlUtils.ts
brudil: Is yarn format okay to run now?

packages/codegen/src/EntityDbMetadata.ts
brudil: Yup! I'll change it. Bit silly really, used infer as the default as I didn't want to pass judgement on the other places fieldType is used.

packages/graphql-codegen/src/graphqlUtils.ts
stephenh: Yep, should be; I snuck in a format PR a few days ago. Lmk if not!

packages/codegen/src/EntityDbMetadata.ts
stephenh: Ah yep, totally get using `z.infer` at first b/c yeah I'm anticipating eventually needing to teach more places about the input-vs-output type difference.

I.e. as-is you've fixed the setter, but places like `em.create`, `new Author(em, ...)`, and `newAuthor(em, ...)` are still using the `AuthorOpts`/output type.

But, if we just flipped the `Opts`/`fieldType` over to the input type, I'm sure we'd have an equal/non-zero number of places that would still want the output type.

So, dunno, probably not something to tackle with this PR, but yep I imagine we'll eventually have to track all/most of the `fieldType`-for-jsonb-usages down and teach them about input-vs-output type...

brudil: With those uses you've listed above, I'm erring on using `z.input` as the default.

I'd perhaps propose in a follow up the idea of allowing us to optionally split field type.

```ts
type FieldTypeType = PrimitiveTypescriptType | Import | Code;

type FieldType = fieldTypeType | { input: fieldTypeType, output: fieldTypeType }
```

and then usages wise we can introduce

```
function getInputFieldType(fieldType: FieldType) {
	return 'input' in fieldType ? fieldType.input : fieldType;
}

function getOutputFieldType(fieldType: FieldType) {
	return 'input' in fieldType ? fieldType.input : fieldType;
}
```

This would allow us to define intent at the point of usage in codegen and keep the field logic in EntityDbMetadata.

PR #687, opened 2023-06-13
fix: JsonSerde args within codegen

brudil:
I introduced an issue within #685 passing the wrong amount of args in to JsonSerde in the codegen, resulting in ts issues in any generated `metadata.ts` for any jsonb columns with no `superstruct` config.

There are no usages of non-superstruct'd json within the integration tests as far as I could tell, which would have caught this, so I've added in a json column to `author_stats` which looked like a good place for this, but let me know!

Sorry about that!


stephenh (approved):
Thanks!

PR #695, opened 2023-06-17
fix: ensure exclusive `or` or `and` in ExpressionFilter

brudil:
One of the things I hit early on was not realising each ExpressionFilter needs an `and` or an `or` but not both. TypeScript doesn't catch this with it's union/object handling.

I've added in never to the other side of expressions to ensure TS does catch this:

```ts
/** Filters against complex expressions of filters. */
export type ExpressionFilter = (
  | { and: Array<ExpressionFilter | ColumnCondition | undefined>; or?: never }
  | { or: Array<ExpressionFilter | ColumnCondition | undefined>; and?: never }
) & { pruneIfUndefined?: "any" | "all" };
```

It doesn't result in the best error, but it at least helps point to the fact joist isn't going to support this:
image

### DIFF
--- a/packages/integration-tests/src/entities/AuthorCodegen.ts
+++ b/packages/integration-tests/src/entities/AuthorCodegen.ts
@@ -86,7 +86,7 @@ export interface AuthorFields {
   graduated: { kind: "primitive"; type: Date; unique: false; nullable: undefined };
   wasEverPopular: { kind: "primitive"; type: boolean; unique: false; nullable: undefined };
   address: { kind: "primitive"; type: Address; unique: false; nullable: undefined };
-  businessAddress: { kind: "primitive"; type: z.infer<typeof AddressSchema>; unique: false; nullable: undefined };
+  businessAddress: { kind: "primitive"; type: z.input<typeof AddressSchema>; unique: false; nullable: undefined };
   quotes: { kind: "primitive"; type: Quotes; unique: false; nullable: undefined };
   deletedAt: { kind: "primitive"; type: Date; unique: false; nullable: undefined };
   numberOfPublicReviews: { kind: "primitive"; type: number; unique: false; nullable: undefined };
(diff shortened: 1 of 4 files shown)


packages/integration-tests/src/entities/AuthorCodegen.ts
```diff
@@ -109,7 +109,7 @@ export interface AuthorOpts {
   graduated?: Date | null;
   wasEverPopular?: boolean | null;
   address?: Address | null;
-  businessAddress?: z.infer<typeof AddressSchema> | null;
+  businessAddress?: z.input<typeof AddressSchema> | null;
```
stephenh, line 112: Did these sneak in from a unrelated/prior change? Or maybe they were missing from a prior change, so bringing this file up to date? Just checking.

packages/orm/src/EntityFilter.ts
```diff
@@ -80,6 +80,6 @@ export type ValueFilter<V, N> =
 
 /** Filters against complex expressions of filters. */
 export type ExpressionFilter = (
-  | { and: Array<ExpressionFilter | ColumnCondition | undefined> }
-  | { or: Array<ExpressionFilter | ColumnCondition | undefined> }
+  | { and: Array<ExpressionFilter | ColumnCondition | undefined>; or?: never }
+  | { or: Array<ExpressionFilter | ColumnCondition | undefined>; and?: never }
```
stephenh, line 84: Cool, good idea!

stephenh (approved):

packages/integration-tests/src/entities/AuthorCodegen.ts
brudil: Ah, yup! I this is bringing up the codegen to date - suspect I didn't re-run codegen after changing the zod fieldType.

Happy to move to a separate PR if you would like to keep this one clean.

stephenh: Np, sneaking it in here is just fine. We should eventually get CI setup to detect diffs in the codegen output--we've got that on our internal repos but haven't set it up here yet.

PR #696, opened 2023-06-17
docs: zod + type + em method semantics

brudil:
We've been working with Joist for the last couple of weeks now and speaking to my team wanted to see what they thought is missing or could be improved with the docs while it's all still fresh and new.

- Add `zodSchema` & `type` to FieldConfig under Configuration
- Add `Zod` to the JSONB Fields page alongside Superstruct
- Mention missing operators "Find Queries"
- Add some explain some addtional semantics around find/findOne/load etc; if they throw, what they throw etc
  - Not entirely happy with the formatting of this, so open to ideas

Bit of a sidenote, but I was suprised to see that `load` just throws a generic base Error and not something from Joist, NotFoundError or otherwise.

### DIFF
--- a/docs/docs/features/entity-manager.md
+++ b/docs/docs/features/entity-manager.md
@@ -88,6 +88,11 @@ const em = newEntityManager();
 const a = await em.load(Author, "a:1");
 ```
 
+- Returns
+    - Entity if found
+    - throws `Error` if not
+
+
 ### `#loadAll`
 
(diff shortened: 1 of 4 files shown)


stephenh (approved):
This looks great!

Per `load`, looks like we do have a `NotFoundError` that is thrown by `findOneOrFail`, but we didn't teach `load` to throw that instead--if you wanted to submit a PR for that, I think that's a good idea.

PR #697, opened 2023-06-18
feat: use NotFoundError in load & loadAll

brudil:
I would imagine this is backwards compatiable as NotFoundError is a sub type of Error.


stephenh (approved):
Looks great!

Per backwards compatibility, that is something we'll probably have to start dialing in better. We have a big enough internal codebase that we're very rarely making large breaking changes, but we've also been bullish on making changes as we flush out features/apis.

PR #698, opened 2023-06-18
feat: add basic custom serde support

brudil:
Following on from the previous attempt and discussions in #678.

This implements a new FieldConfig option `serde`. This overrides the class used within metadata.ts. I've left this as only taking precedence over all others if set. (One thing to suggest in the future for dx is the idea of friendly validation of joist-config.json when running codegen).

I've implemented Serde as an abstract class to reduce boilerplate for consumers and the finer nuance (which fields from the constructor need to be public, `column = [this]`.

Tests added around a Users password "hashed" with base64 (perfect security :)) also combining with `type` for the value objects use case.

### DIFF
--- a/packages/integration-tests/src/entities/User.test.ts
+++ b/packages/integration-tests/src/entities/User.test.ts
@@ -1,5 +1,10 @@
 import { User } from "@src/entities";
+import { insertUser } from "@src/entities/inserts";
+import { PasswordValue } from "@src/entities/types";
 import { newEntityManager } from "@src/setupDbTests";
+
+const PASSWORD = "correct.horse.battery.staple";
+const PASSWORD_ENCODED = "Y29ycmVjdC5ob3JzZS5iYXR0ZXJ5LnN0YXBsZQ==";
 
 describe("User", () => {
   it("custom type is exposed", async () => {
@@ -7,5 +12,44 @@ describe("User", () => {
     // @ts-expect-error
     const a1 = em.create(User, { name: "a1", email: "test@test.com", ipAddress: "127.0.0.1" });
     await expect(em.flush()).resolves.toEqual([a1]);
+  });
+
+  it("can interact with serde value", async () => {
+    const em = newEntityManager();
+    const a1 = em.create(User, {
+      name: "a1",
+      email: "test@test.com",
+      password: PasswordValue.fromPlainText(PASSWORD),
+    });
+    await expect(em.flush()).resolves.toEqual([a1]);
+    await expect(a1.password?.encoded).toEqual(PASSWORD_ENCODED);
+  });
+
+  it("loads serde values correctly", async () => {
+    insertUser({
+      id: 1,
+      name: "a1",
+      email: "test@test.com",
+      password: PASSWORD_ENCODED,
+    });
+
+    const em = newEntityManager();
+    const a1 = await em.load(User, "u:1");
+    await expect(a1.password?.encoded).toEqual(PASSWORD_ENCODED);
+  });
+
+  it("can interact with serde value", async () => {
+    insertUser({
+      id: 1,
+      name: "a1",
+      email: "test@test.com",
+      password: PASSWORD_ENCODED,
+    });
+
+    const em = newEntityManager();
+    const a1 = await em.load(User, "u:1");
+    a1.password = PasswordValue.fromPlainText("another");
+    await expect(em.flush()).resolves.toEqual([a1]);
+    expect(a1.password.matches(PASSWORD)).toBe(false);
   });
 });
--- a/packages/integration-tests/src/entities/types.ts
+++ b/packages/integration-tests/src/entities/types.ts
@@ -1,3 +1,4 @@
+import { Serde } from "joist-orm";
 import { Infer, array, object, string } from "superstruct";
 import { z } from "zod";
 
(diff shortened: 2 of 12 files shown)


packages/integration-tests/src/entities/types.ts
```diff
@@ -15,3 +16,34 @@ export const AddressSchema = z.object({
...
+  }
+
+  constructor(public readonly encoded: string) {}
+
+  matches(str: string) {
+    return Buffer.from(str, "utf8").toString("base64") === this.encoded;
+  }
+}
+
+export class PasswordValueSerde extends Serde {
```
stephenh, line 36: This is great; I really like seeing the implementation & your `PasswordValueSerde` impl--seeing them in action, I'm wondering if we could find a way for the user to provide _only_ `fromEncoded` and `toEncoded` (or equivalents).

Like maybe instead of the user providing a full-fledged `Serde` impl, we could have a `CustomSerde` implementation that looked like:

```
export class CustomSerde extends Serde {
  constructor(...the user's encode & decode...) {}
  dbValue(data: any): any {
    return this.toDb(data[...]);
  }
  mapToDb(value: any): any {
    return this.fromDb(value);
  }
  setOnEntity(data: any, row: any): void {
    data[this.fieldName] = this.fromDb(row[this.columnName]);
  }
}
```

As this would a) reduce the boilerplate that user types have to implement, and b) reduce the API coupling of the current `Serde` internal methods (that per my previous disclaimers are "just okay" imo) to the userland serde implementations.

It seems like we could achieve this with two ways:

1. Have `CustomSerde` be an `abstract class` that provided default impls of the existing `dbValue` / `mapToDb` / `setOnEntity` methods, and just asked the user to implement two new simpler methods, like `toDb` & `fromDb`.

2. Keep `CustomSerde` internal/private, and instead ask the user to implement a new "pair of functions" type. Maybe the internal serde is called like `class CustomSerdeAdapter implements Serde` and we export an interface that is just `interface CustomSerde<DomainType, DbType>` with `toDb(dt: DomainType): DbType` and `fromDb(db: DbType): DomainType` methods.

Wdyt about 1 vs. 2? I think I prefer 2...

Granted, the `CustomSerde` `toDb` / `fromDb` methods would make assumptions that the user's serde is only 1 db column <-> 1 domain field, and in theory we should eventually let users do N db columns <-> 1 domain field, i.e. a `Currency` domain type that used two `currency_amount` and `currency_type` db columns for storing a single domain type...

I'm not sure how we'd grow my current `interface CustomSerde` + `toDb` + `fromDb` suggestion into one that also supported this more complicated use case--but that would probably be a ways out anyway. ...and maybe it'd be as simple as adding a `columns: [..., ...]` field to the JSON config, which if present then we'd ask the user to implement a `MultiColumnSerde` that was not as dead simple as the proposed `CustomSerde`+`toDb`+`fromDb`.

Okay, I think that's enough of an idea of how to support multi-column serde "someday" that I'm pretty comfortable doing the option 2 above for now...

Apologies for asking/brainstorming for another rev, but hopefully this will be worth it. Wdyt?

brudil: I'm a fan of two too! I like the flexability it provides being just a simple object.

The fact we've got the users serde impl being wrapped inside of our CustomSerdeAdapter also doesn't lock us in too much which is nice.

I have a impl of these changes, but I'm considering one thing - should the user's serde be aware of nullability? I'm personally erring on no. Null is represented as the absence of a value and essentially outside the scope of serde as serde is for mapping values. I think this is really the only fesible way to go for now at least, as this currently how the `type` config works and creates questions around default values etc. But thought it was worth flagging to ensure we're on the same page re semantics.

brudil:
Pushed an initial implementation of two. I assume the discussions on internal/private are in terms of docs/theoretical as all Serde classes need to be exported as they end being imported in the user's generated metadata.ts.

As mentioned above, this essentially skips the custom serde mapping if the value is undefined.

packages/integration-tests/src/entities/types.ts
stephenh: Yep, agreed that we can keep nullability logic encapsulated in our internal serde, and let the user's serde just detail with non-null values.

packages/integration-tests/src/entities/User.test.ts
```diff
@@ -1,11 +1,55 @@
...
+
+  it("can interact with serde value", async () => {
+    const em = newEntityManager();
+    const a1 = em.create(User, {
+      name: "a1",
+      email: "test@test.com",
+      password: PasswordValue.fromPlainText(PASSWORD),
+    });
+    await expect(em.flush()).resolves.toEqual([a1]);
+    await expect(a1.password?.encoded).toEqual(PASSWORD_ENCODED);
```
stephenh, line 25: These assertions look good, could add another, using our `select` helper to do like:

```
expect((await select("authors"))[0].password).toEqual(PASSWORD_ENCODED);
```

Just to make sure the encoded value goes all the way through to the db.

packages/integration-tests/src/entities/User.test.ts
```diff
@@ -1,11 +1,55 @@
...
+      password: PASSWORD_ENCODED,
+    });
+
+    const em = newEntityManager();
+    const a1 = await em.load(User, "u:1");
+    await expect(a1.password?.encoded).toEqual(PASSWORD_ENCODED);
+  });
+
+  it("can interact with serde value", async () => {
+    insertUser({
```
stephenh, line 42: I think might need awaits for these `insertUser` calls?

stephenh (approved):
Looks great! Thank you!

PR #703, opened 2023-06-21
fix: Fix multiple nested conditions.

### LINKED ISSUE #701: Operator does not exist: integer = integer[]
It is managing the args number but when there is a deep condition (complex condition) it’s summing an extra number to the argsIndex.

```sql
WITH _find ("tag", "arg0", "arg1", "arg2", "arg3") AS (VALUES ($1::int, $2::int, $3::timestamp with time zone, $4::int, $5::int[]), ($6, $7, $8, $9, $10) ) SELECT array_agg(_find.tag) as _tags, "b".* FROM books as b LEFT OUTER JOIN book_advances ba ON b.id = ba.book_id JOIN publishers p ON ba.publisher_id = p.id JOIN _find ON b.deleted_at IS NULL AND b.id = _find.arg0 AND ((ba.created_at > _find.arg1) OR (ba.status_id = _find.arg3 AND p.type_id = ANY(_find.arg4))) GROUP BY "b".id ORDER BY b.title ASC, b.id ASC LIMIT 10000; - operator does not exist: integer = integer[]
```

Where: 

`arg0` => is correct
`arg1` => is correct
`arg2` => should be used for `ba.status_id`
`arg3` => should be used for `p.type_id`
`arg4` => extra arg added for a logic bug on [buildConditions](https://github.com/stephenh/joist-ts/blob/main/packages/orm/src/dataloaders/findDataLoader.ts#L217).

Here is a test case to reproduce:

```ts
fit("test", async () => {
    const [book, bookAdvance, publisher] = aliases(Book, BookAdvance, Publisher);
    resetQueryCount();
    // Querying all books with a paid advance and a big publisher
    // It should create a inner join between publisher and bookAdvance (many to one)
    // And a left join between book and bookAdvance (one to one)
    await em.transaction(async () => {
      const queries = zeroTo(2).map((id) =>
        em.findGql(
          Book,
          { as: book, advances: { as: bookAdvance, publisher }, id: `${id}` },
          {
            conditions: {
              or: [
                {
                  and: [bookAdvance.createdAt.gt(jan1)],
                },
                {
                  and: [bookAdvance.status.eq(AdvanceStatus.Paid), publisher.type.in([PublisherType.Big])],
                },
              ],
            },
          },
        ),
      );
      return queries;
    });
  });
```

stephenh:
Fixes #701.

### DIFF
--- a/packages/orm/src/dataloaders/findDataLoader.ts
+++ b/packages/orm/src/dataloaders/findDataLoader.ts
@@ -216,9 +216,13 @@ function visit(query: ParsedFindQuery, visitor: Visitor): void {
 // Create the a1.firstName=data.firstName AND a2.lastName=data.lastName
 function buildConditions(ef: ParsedExpressionFilter, argsIndex: number = 0): [string, number] {
   const conditions = [] as string[];
+  const originalIndex = argsIndex;
   ef.conditions.forEach((c) => {
     if ("cond" in c) {
       const [op, argsTaken] = makeOp(c.cond, argsIndex);
+      if (c.alias === "unset") {
+        throw new Error("Alias was not bound in em.find");
+      }
       conditions.push(`${c.alias}.${c.column} ${op}`);
       argsIndex += argsTaken;
     } else {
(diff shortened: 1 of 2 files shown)


packages/orm/src/dataloaders/findDataLoader.ts
```diff
@@ -229,7 +233,8 @@ function buildConditions(ef: ParsedExpressionFilter, argsIndex: number = 0): [st
       argsIndex += argsTaken;
     }
   });
-  return [conditions.join(` ${ef.op.toUpperCase()} `), argsIndex];
```
stephenh, line 232: @allangaldinosilva you were exactly right that the bug was in `buildConditions`; this was my mistake, the return value here should be "the number of args taken" and not just "the latest index value".

Thanks for the great reproduction!

PR #705, opened 2023-06-22
fix: move string/null cast to setter, herustic disable on default ""

brudil:
Initial impl following on from our discussion in https://github.com/stephenh/joist-ts/issues/704.

- Moves the string -> undefined check out of setOpts to the setters
- Skips this transformation if the default value is empty string `''`
- I've had to move the transformation in to a `orm` function that is exported and imported by codegen due to a weird TS issue - not sure if it's a bug or something I've just never hit before; `typeof` can create `never` - I always assumed typeof would cast it anyway. If there's a workaround, let me know and I'll update. I guess we could cast as unknown within the setter.


packages/codegen/src/generateEntityCodegenFile.ts
```diff
@@ -96,6 +97,10 @@ export function generateEntityCodegenFile(config: Config, dbMeta: DbMetadata, me
     }
 
     let setter: Code | string;
+    if (p.columnName === "bio") {
```
stephenh, line 100: Nit: snuck in

packages/integration-tests/migrations/1580658856631_author.ts
```diff
@@ -200,6 +200,8 @@ export function up(b: MigrationBuilder): void {
     }),
     ip_address: { type: "varchar(255)", notNull: false },
     password: { type: "varchar(255)", notNull: false },
+
+    bio: { type: "varchar(255)", notNull: true, default: "" },
```
stephenh, line 204: Nice!

stephenh (commented):
This looks pretty great! One small ask but otherwise lgtm

packages/codegen/src/generateEntityCodegenFile.ts
brudil: Drats! Good spot. Wasn't sure if columnDefault was going to be "" or "''".

stephenh (approved):
Looks great!

PR #707, opened 2023-06-23
chore: Remove Makefile in favour of package.json scripts

chr1sjf0x:
Per discussions about all commands using `yarn [command]`


package.json
```diff
@@ -7,7 +7,10 @@
     "migrate": "yarn workspaces foreach run migrate",
     "codegen": "yarn workspaces foreach run codegen",
     "format": "yarn workspaces foreach run format",
-    "test": "yarn workspaces foreach run test"
+    "test": "yarn workspaces foreach run test",
+    "docker": "docker-compose build",
+    "db": "docker compose up db --wait & docker compose exec db ./reset.sh & yarn migrate & yarn codegen",
```
stephenh, line 12: Question: What's the behavior of `&` vs. `&&` which is usually what I type out of habit?

stephenh (approved):
Nice!

package.json
chr1sjf0x: Good eye, the single & means run the command in the background. && gives the serial execution we desire here 👍

stephenh:
Sweet, thnks @chr1sjf0x !

PR #708, opened 2023-06-23
docs: Add `tsvector` full text search example

blambillotte:


blambillotte:
@stephenh Took a stab at an example of full text search. I couldn't quite figure out where to put it, maybe it makes senes to make a new `Examples` top level folder? `Getting Started/Examples` isn't quite right.

stephenh:
@blambillotte this looks great!

In terms of where to put it :thinking: , yeah, seems like maybe we eventually need like a "Recipes" or "Examples" dedicated section.

But for now given we have only one example, it seems overkill to create a new section quite yet.

Wdyt of just putting it into the "Advanced Features" section for now?

stephenh (approved):
:100:

PR #710, opened 2023-06-25
fix: Don't infinite recurse on required self-ref keys.

### LINKED ISSUE #702: Support self-referential FKs in factories
Something like:

```
import { self } from "joist-orm";

export function newAuthor(em: EntityManager, opts: FactoryOpts<Author> = {}): DeepNew<Author> {
  return newTestInstance(em, Author, opts, {
    favoriteAuthor: self,
  });
}

...

const a = newAuthor(em);
expect(a.favoriteAuthor.get).toEqual(a);
```

stephenh:
Skipping an integration test, but locally I made Author.mentor required and verified that it works.

Granted, making this work for a non-pg / non-sequence database is going to suck.

Fixes #702

### DIFF
--- a/packages/orm/src/newTestInstance.ts
+++ b/packages/orm/src/newTestInstance.ts
@@ -60,6 +60,8 @@ export function newTestInstance<T extends Entity>(
   const opts = mergeOpts(testOpts, factoryOpts);
   const use = getOrCreateUseMap(opts);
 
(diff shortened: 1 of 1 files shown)


stephenh:
cc @chr1sjf0x and @roberthdg the factories should handle required self-referential FKs now.

PR #711, opened 2023-06-26
feat: joist-doc

brudil:
Getting up a WIP version of this to see if you're happy to progress with.

## Overview
Implemented as a `joist-doc` package, which run within codegen after file emit, if `"docGen": "markdown"` is set in `joist-config.json`. Is a seperate optional package which would need to be part of the user's dependencies.

`joist-doc` depends on `joist-codegen` and is given the `Config` and `EntityDbMetadata`. Using the `EntityDbMetadata` object, it runs against the known file types, (Entities [todo], EntityCodegen, Enums [todo]).

The files are parsed with babel. Originally this was supposed to ideally use the user's `typescript` but the typescript parser doesn't have the best support for comments within the AST, relying on the complier's linting infrastructure.

From there, each of the file types can traverse the AST finding specific nodes that might need comments. When a commentable node is found, the `CommentStore` is called to optionally provide a comment. If the CommentStore returns a comment, it is added back to the AST to be printed back out and written to the file.

This PR implements a default `MarkdownCommentStore` implementation, which uses `remark` to parse the markdown for each manipulation.

`MarkdownCommentStore` opts for lose-ish heriustics on determining comments, allowing for greater flexability in how users format their markdown files.

The goal was to support both these files:

_Something structured, more detailed, nested like_
```md
# User
## Overview
A User entity is...

## Fields
### email
Something about their email

### Computed Fields
#### isActive
If a login has occured within the last 30 days
```

_And something looser like_

```md
# User
A User entity is...

## email
Something about their email

## isActive
If a login has occured within the last 30 days
```

Thus the implement herustics are as follows:
- for Fields: we find a matching heading (of any level) of the field name
- for Root descriptions (entities): we use the following rules:
  - paragraphs from the top
  - paragraphs from the following the first heading
  - paragraphs from the second heading if it preceeds the first and is called "Overview"

All content is collected until it hits a non-paragraph element. This allows for a neat trick of using a horizontal rule within your doc to stop what is included within the comments, but allowing for more depth within the markdown file.

We obviously might want to tweak these or improve them.

---

There's a little more to go and some code to clean up but wanted to see this direction fits in with what we discussed.


stephenh:
This looks amazing! Yep, definitely looks like a great approach/direction.

In terms of babel, that sgtm; I had seen ts-morph before, i.e. https://ts-morph.com/setup/ast-viewers, which I think could be a candidate for this? I was trying to see if it used babel or not, but I don't see it pulling in any babel/similar dependencies, so my current guess is that it uses the tsc compiler API, so then it probably/I assume has the same limitation you'd found wrt comments.

Fwiw I'm pretty happy to not use the tsc API b/c I think it will also be not-the-fastest.

Granted, anything we write (in JS) will probably need caching; i.e. if `FooCodegen.ts` / `Foo.ts` / `Foo.markdown` haven't changed, then don't pay the cost of re-parsing + re-stitching + re-formatting the AST. Definitely something we can add later, but will probably be required for our internal project to adopt this. 

(When Joist used prettier for formatting, I'd started doing some caching to avoid the cost of formatting every `*.ts` file on every `yarn joist-codegen` b/c it was taking 10s of seconds just to format the entire generated output each time, but then switching to the Deno project's Rust-based dprint formatter made formatting so fast, I was able to just take the caching out.)

> as the integration tests don't care about the actual text within the codegen'd files.

True; fwiw the rationale for checking the generated files into git is to allow admittedly-manual regression testing as PR reviewers can see exactly what changed in the generated files.

Granted, that's not as great as true red/green unit tests, but :shrug: I'd be fine with it to start with, and add fancier/automated assertions later/if needed.

This is really neat!

stephenh:
Wdyt about adding a hint to the comments that "this comment comes from `Author.md`, so don't edit it".

Granted, for `AuthorCodegen.ts` I think it's kinda implicit the user should not edit the comments, but if they're in `Author.ts` and editing the class tsdoc, I could see that being confusing that they really need to go edit the `Author.md`...

brudil:
A source hint sounds like a great idea, I'll implement that for the Entity files.

On perf, this sounds good. Does joist-codegen not emit on non-changed files? Just trying to figure if this is caching "this doesn't need to be touched" or in terms of "this file hasn't changed (hash or something), and we've cached the previous output"

brudil:
@stephenh Out of curiosity how many "doc-able" structures (Entities, enums) do have internally the project? I've kept the current deps light, but it might be worth pulling in `p-limit` or the like as currently every transformation is executed within a Promise.all.

brudil:
Implemented caching with `file-system-cache` (https://www.npmjs.com/package/file-system-cache) which is used by Storybook, persisting to a `.cache` directory. Let me know if you have another preference for this though.

Should get time to tackle enums and entity files themselves tomorrow.

brudil:
Entity files and enum supported added.

Now we've got caching, I've added dprint to get the files back to where they were before joist-doc ran.

I've taken the config directly from ts-poet - it is improved, but still not as if only the comments where added. I'm not sure if this is because I've used `@dprint/formatter` and not the `dprint-node` that `ts-poet` uses - perhaps some changes between versions? Or if those lines are lines that are just new lines from codegen which are lost within babel.

Let me know if you have any thoughts. Otherwise, I think this is pretty much there in terms of functionality.

stephenh:
Sweet! Haven't had a chance to take a look yet, but yeah in terms of `@dprint/formatter` vs. `dprint-node`, afaiu they are basically separate projects for two of the dprint core contributors, where `@dprint/formatter` is "more official" but has a ~kinda-weird imo approach where it downloads plugins + runs them in web assembly, vs. `dprint-node` which is just a stock node native library.

I briefly played with `@dprint/formatter` in ts-poet, but the web assembly approach incurs a mini-compilation on each run, so `dprint-node` ended up being noticeably faster.

Is it possible to keep using `dprint-node`?

brudil:
Sure thing, I'll get that moved over.

brudil:
Moved over just fine.

packages/doc/src/MarkdownCommentStore.ts
```diff
@@ -0,0 +1,148 @@
...
+/**
+ * A Markdown CommentStore with opinionated but loose semantics around
+ * markdown document structure
+ */
+export class MarkdownCommentStore extends CommentStore {
+  private documents = new Map<string, Promise<null | ReturnType<(typeof remark)["parse"]>>>();
+
+  async forField(entity: EntityDbMetadata, fieldName: string, source: FieldSourceType, generated: boolean) {
+    return this.withGenerated(entity.name, generated, await this.findHeadingContent(entity.name, fieldName));
+  }
```
stephenh, line 17: Wdyt of making the `CommentStore` API a bit more coarse-grained, like:

```
interface CommentStore {
  getEntityComments(entity: EntityDbMetadata): EntityComments;
  getEnumComments(enum: EnumData): EnumComments;
|

interface EntityComments {
  classDoc: string;
  fieldDocs: Record<string, string>;
  sourceHash: string;
}
```

Well, I guess we want to get the `sourceHash` before going to the trouble of parsing the whole doc out. So yeah, okay, this fine-grained approach makes sense!

packages/doc/src/MarkdownCommentStore.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  async forEntity(entity: EntityDbMetadata, generated: boolean): Promise<string | void> {
```
stephenh, line 19: Huh, I've not used `| void` as an empty return value for; can we use `| undefined` instead? Just is what I'm more used to.

packages/doc/src/CommentStore.ts
```diff
@@ -0,0 +1,63 @@
...
+export type FieldSourceType = "get" | "set" | "opts";
+
+export type EnumData = { name: string };
+
+/**
+ * The base API for a providing comments to be applied to the various files.
+ *
+ * Focused around Joist concepts (Entities, Fields etc) over the files themselves.
+ */
+export abstract class CommentStore {
```
stephenh, line 16: Given we have no shared/base impls, wdyt of making this an `interface`?

packages/doc/src/CommentStore.ts
```diff
@@ -0,0 +1,63 @@
...
+  constructor(protected config: Config) {}
+
+  /**
+   * Provides a comment for a Entity Field
+   */
+  abstract forField(
+    entity: EntityDbMetadata,
+    fieldName: string,
+    source: FieldSourceType,
+    generated: boolean,
```
stephenh, line 26: Ah I see, maybe call this `isCodegen` or `isCodegenFile`? I was thinking "well everything that we're outputting is generated", but we meant like the base-vs-codegen file.

I was thinking maybe the integration could inject this, instead of the comment store, but I guess the comment store is the one that knows the file name.

Wdyt of maybe always including the "generated from ..." suffix? Like even if we're writing to a codegen file, I could see someone reading the getter comments and seeing the same "ah I go to Foo.md" hint being somewhat useful. Dunno, unless you think it'd be too cluttered and we should taken the opportunity of "this is a codegen file that is obvious the comment is generated from somewhere" to skip the boilerplate.

packages/doc/src/MarkdownCommentStore.ts
```diff
@@ -0,0 +1,148 @@
...
+  private async getDocument(filename: string) {
+    const existing = this.documents.get(filename);
+
+    if (existing !== undefined) return existing;
+
+    const promise = readFile(path.join(this.config.entitiesDirectory, `${filename}.md`), { encoding: "utf-8" })
+      .then((contents) => {
+        return remark.parse(contents);
+      })
+      .catch(() => null);
```
stephenh, line 54: Maybe do a `console.warn` here just so that users can know that we skipped their file due to an error?

packages/doc/src/MarkdownCommentStore.ts
```diff
@@ -0,0 +1,148 @@
...
+    return promise;
+  }
+
+  /**
+   * Finds content for a given heading. The semantics used here are:
+   * - any level heading, which allows for more flexibility in how users want
+   *   format their docs (second level, 3rd under more groupings etc)
+   * - if found, we continue to collect paragraphs until there is another node
+   *   This means horizontal rules can be used to limit the amount ingested.
+   *   Allowing for more in-depth documentation in the markdown file.
```
stephenh, line 67: Great comment, thanks!

packages/integration-tests/src/entities/User.ts
```diff
@@ -1,7 +1,8 @@
-import { UserCodegen } from "./entities";
-
-import { userConfig as config } from "./entities";
-
+import { UserCodegen, userConfig as config } from "./entities";
```
stephenh, line 1: I'm surprised there isn't a new line after this `import`... looks like maybe Babel is dropping all new lines when it does `generate`?

I wonder if there is a way to avoid that b/c someone Jest's `toMatchInlineSnapshot` does not, although maybe the difference is that they use prettier which just puts the lines back, and my guess is that dprint probably removes extra new lines, but isn't as opinionated about adding newlines if necessary... Wonder if that's an option...

packages/integration-tests/src/entities/User.ts
```diff
@@ -1,7 +1,8 @@
...
+/**
+ * Representing a User of the application. Users are the tied to the authentication system and can log in.
+ *
+ * @author Joist */
```
stephenh, line 5: Kinda picky, but could we get the `*/` to line up on the last line, like:

```
/**
 * Foo bar.
 *
 * @author Joist
 */
``` 

Er, in a different file down below, this `*/` was placed on the new line...I guess in the codegen file, it's added a `\n` I guess? Not sure why, but is that something we could do here as well?

packages/integration-tests/src/entities/UserCodegen.ts
```diff
@@ -163,92 +166,94 @@ export abstract class UserCodegen extends BaseEntity<EntityManager> {
...
+  /**
+   * Hashed with ultra secure base64.
+   */
   get password(): PasswordValue | undefined {
     return this.__orm.data["password"];
   }
-
+  /**
+   * Hashed with ultra secure base64.
+   */
```
stephenh, line 229: Wdyt of detecting when the comment has no newlines, and producing a one-liner comment like:

```
/** Hashed with ultra secure base64. */
```

Also perfectly fine to be a follow up/maybe someday.

packages/integration-tests/src/entities/UserCodegen.ts
```diff
@@ -163,92 +166,94 @@ export abstract class UserCodegen extends BaseEntity<EntityManager> {
...
   get ipAddress(): IpAddress | undefined {
     return this.__orm.data["ipAddress"];
   }
-
+  /**
+   * This is an example of a multi
+   * line comment with
+   *
+   * line breaks and also to show that we just use headings, and not the level, allowing for more customisation in users docs.
+   */
```
stephenh, line 217: Do you know if tsdoc wants/expects comments on both the getter and the setter? I would have guessed when like viewing this in vscode, there is only one logical `ipAddress` property, so it would just like pick one, say the getter, and treat that as where comments should go.

Or maybe it like merges the getter+setter comments? Not sure...

packages/doc/src/Cache.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+  }
+
+  save() {
+    this.fsCache.setSync("manifest", JSON.stringify(this.manifest));
+  }
+
+  async set(filePath: string, { sourceHash, commentStoreHash }: HashPair, generated: string) {
+    this.manifest[filePath] = [sourceHash, commentStoreHash];
+    await this.fsCache.set(hashString(filePath), generated);
```
stephenh, line 40: Just curious, but why are we hashing the filePath before using it as a key? Is it to avoid special/weird characters that `fsCache` would not like as its key values?

packages/doc/src/index.ts
```diff
@@ -0,0 +1,92 @@
...
+   *
+   * - calculates source hash
+   * - asks commentStore for its hash
+   * - if nothing has changed and we have a restore, exit early and use that
+   *
+   * - otherwise: parse, run handler, generated, dprint, write, update cache
+   */
+  async run<T>(integration: IntegrationHandler<T>, topic: T) {
+    const filePath = integration.file(topic, this.config);
+    const file = await readFile(filePath, { encoding: "utf-8" });
```
stephenh, line 32: Maybe call this `contents`? I briefly thought `file` was still the file path, i.e. when reading `hashString(file)`.

packages/doc/src/index.ts
```diff
@@ -0,0 +1,92 @@
...
+        await writeFile(filePath, existingCache, { encoding: "utf-8" });
+        return;
+      }
+    }
+
+    const source = parse(file, { sourceType: "module", plugins: ["typescript"] });
+    const result = await integration.handle(source, topic, this.commentStore);
+
+    // This is taken from ts-poet, which joist-codegen doesn't override. It's included here
+    // to try to ensure that joist-doc isn't reformatting needlessly
```
stephenh, line 50: Thanks for the comment! I imagine ts-poet could export these settings or its own `format(...)` method someday/when we get to it.

stephenh (approved):
Apologies for taking a few days to get to this, but looks great @brudil ! I haven't ran it on our internal app yep, but looking forward to trying it out!

packages/doc/src/CommentStore.ts
brudil: The only thing we can do with an interface is define the constructor signature right? Can change if that's a willing trade off.

packages/doc/src/CommentStore.ts
brudil: Hmm, yeah. I get the name issue. We could flip it to isUserland or the like.

CommentStore has been built to not assume file backing; if someone (no judgement on if this is a terrible idea) wanted to build something like `NotionCommentStore`.

I'll remove this from the CommentStore API, and just always incude it within MarkdownCommentStore. We can always add it back in if it does get cluttered.

packages/integration-tests/src/entities/UserCodegen.ts
brudil: Yeah, I did a bit of a deep dive looking in to this before starting. I believe this is a bit messy at the moment:

- set & get are treated independently at a TSDoc level
- They are merged at usage. There's an open bug for this though (https://github.com/microsoft/TypeScript/issues/48801), so I didn't want to model how this works around what is marked as a bug. With the `fieldSourceType` property, it would be possible for CommentStores to tweak this, by deciding to only provide comments for getters.

packages/doc/src/Cache.ts
brudil: Correct! I'll try it without though because this was just an assumption; looks like file-system-cache is doing its own encoding of the filenames so this might not be required.

brudil: Works without, removed!

packages/doc/src/MarkdownCommentStore.ts
brudil: This essentially catches the file not existing cases, which might be quite noisy to log? It's perhaps a badly placed catch as it's really for the readFile, not the remark.parse which I can't imagine throws that often.

packages/doc/src/CommentStore.ts
brudil: Another thought on this one; if we wanted it to be handled by the integration, we could have the CommentStore API return something like

`type CommentResult = { comment: string, source?: string }`

So we can easily build the string of `Generated from {source}` if given.

brudil: Sorry for the spam; one final thought. I wondered if there's a relevant TSDoc tag to use for this. The one that perhaps best fits is `@privateRemarks` or alternatively a custom tag such as `@source`

packages/doc/src/MarkdownCommentStore.ts
brudil: If you're happy I'll leave as is for the reasons mentioned. sourceHash is the only thing we need for the best-case cached scenario, which avoids the parse and walking of the ast.

brudil: It might be possible to follow an API like this, but retain the performance aspects

```ts
interface CommentStore {
  getEntityComments(entity: EntityDbMetadata): LazyTopic<EntityComments>;
  getEnumComments(enum: EnumData): LazyTopic<EnumComments>;
}

interface LazyTopic<T> {
  sourceHash: string;
  load(): Promise<T>
}

interface EntityComments {
  classDoc: string;
  fieldDocs: Record<string, string>;
}
```

brudil: Another thing perhaps worth mentioning - this inverts flow which would have detrimental perf impact for large documents, Currently we request the from the CommentStore fields that are present in the source. The above would need to treat every heading in the markdown file as a possible field/property in the source, so would product a lot of needless comments that are not tied to fields in the source.

I'm err-ing on leaving as is.

packages/integration-tests/src/entities/User.ts
brudil: I had another look at Jest, they used a generate option I hadn't enabled `{ retainLines: true }` but alas this still doesn't do the job. So I'm curious at how they are doing that - perhaps it is prettier.

From what I've seen, the Babel AST doesn't retain whitespace, so this isn't going to work how we want it to.

The option here is to use https://github.com/benjamn/recast which keeps track of the original ast to only print the bits that have changed.

packages/integration-tests/src/entities/User.ts
brudil: This seems to be fixed now.

brudil:
Couple of discussion points left, resolved the others.

packages/doc/src/MarkdownCommentStore.ts
stephenh: Yep, sgtm!

packages/doc/src/CommentStore.ts
stephenh: Ah yeah, that's true; I'd assumed plugins would want the freedom of "just implement this interface", but if they extend the class, you're right it does document the constructor for things like accepting `config`. Otherwise accepting the `config` would have to be like an `init` method or implicit. Sounds good to keep it then!

packages/doc/src/CommentStore.ts
stephenh: Agreed a custom tag like `@source` seems neat!

I like the `{ comment: string; source?: string }` API and then, yeah, let the integration decide whether it injects the source (or maybe just always inject the `@source`, even for comments in `AuthorCodegen.ts`).

packages/doc/src/MarkdownCommentStore.ts
stephenh: Ah I see, agreed the file missing shouldn't log...

I guess you're probably right that `remark.parse` doesn't throw very often/if ever given how lenient markdown is... I was thinking of like parsing ASTs and things, where parse errors are common, but, right the worst thing that happens with Markdown is an misinterpretation of your wip content, and not a true parse error.

Makes sense!

packages/integration-tests/src/entities/User.ts
stephenh: Just a naive question, but per "retainLines retains too much", is that how in the screenshot the `/** Case is preserved...` is on the same line as `name: string`? Or what is the unexpected retainage?

For the `/** Case is preserved`, if we just add our own newline ...or equivalent AST entry...? would that nudge it down to the next line?

packages/integration-tests/src/entities/UserCodegen.ts
stephenh: Hm, does the "merged at usage" apply to tsdoc output (not just internally)?

Kinda wondering if we could just annotate the getter and assume readers are either just looking at the getter, or using vscode / something that merges the getter+setter since typically the getter+setter is exposed as a single logical property.

Does that seem reasonable, or does the doc output you're using really look best with the comment included in both the getter + setter?

packages/doc/src/Cache.ts
stephenh: Sweet!

packages/codegen/src/config.ts
```diff
@@ -46,6 +46,8 @@ export interface Config {
   /** Your application's request-level `Context` type. */
   contextType?: string;
 
+  docGen?: string;
```
stephenh, line 49: Kinda picky/ocd, but maybe just "docs" instead of docGen?

packages/integration-tests/src/entities/Author.ts
```diff
@@ -155,39 +142,34 @@ export class Author extends AuthorCodegen {
...
+  }, (author) => author.publisher.get?.comments.get[0].entity ?? author.comments.get[0].entity);
 
   /** Example of an async property that has a conflicting/overlapping reactive hint with ^. */
-  readonly allPublisherAuthorNames: AsyncProperty<Author, string | undefined> = hasReactiveAsyncProperty(
-    { publisher: { authors: "firstName" } },
-    (author) => author.publisher.get?.authors.get.flatMap((a) => a.firstName).join(),
-  );
+  readonly allPublisherAuthorNames: AsyncProperty<Author, string | undefined> = hasReactiveAsyncProperty({
+    publisher: { authors: "firstName" },
+  }, (author) => author.publisher.get?.authors.get.flatMap((a) => a.firstName).join());
```
stephenh, line 153: Hrm, this reformatting might be problematic, b/c I think it's undoing whatever prettier/dprint would do...

Hm, I wonder if this is just a legitimate difference between prettier & dprint--like dprint is "prettier-ish", but not 100% the same, and we generally keep them separated out, like `*Codegen.ts` is done by dprint (fast b/c we regularly format all files) & `*.ts` is done by prettier (not fast, but joist-codegen only touches them once on file creation, so it doesn't matter).

But, I bet if we more regularly "ran dprint on prettier-controlled files", or "ru nprettier on dprint-controlled files", we'd get these sorts of things...

Makes me wonder if each integration needs to know "is my source file usually formatted by dprint or by prettier" and then just use the appropriate formatter...

Kinda annoying, but given you'd added caching for the doc-gen piece, we should be good to use prettier I think...everyone's very first `yarn joist-codegen` run will re-prettier the entity files to get their `.cache` populated, but that shouldn't be too bad...

Wdyt of like conditional dprint-vs-prettier formatting of the output?

packages/integration-tests/src/entities/UserCodegen.ts
brudil: I'm happy to move to just getters, it's an easy thing to change if there is ever movement on the above issue (which I think essentially refers to how the TS LSP exposes this stuff) and as it's touching code-generated files.

You can see the behavior here, hoving over the usages at the bottom. https://www.typescriptlang.org/play?#code/MYGwhgzhAEAKYCcCmA7ALtA3gKGtX0A9AFTEHTHQCS0A5kmmgJYq3RgC270KnSAdOWKEC9DLw5IAFAEos2AgF8FePARJlVlGhAbNW7LmB59BWkXl3i+s+UpWr1pIdUPc0ACyTQIpoRboGaCYAE1RmAE9bHHtyfDwNFxpOdy8fP3MCK2Cw9CYouRi8ZWVsFCQAdzhEcNl+CSQFcqr4ZHQ60PD8oA

packages/integration-tests/src/entities/User.ts
brudil: Correct. It seems like retainLines is largely there to not disrupt line numbers, stack traces etc.

Babel docs:
> This option exists so that users who cannot use source maps can get vaguely useful error line numbers.

I'm thinking it's worth perhaps biting the bullet here and trying out recast, because if it works as I think it should, it will completely absolves us of needing to do any reformatting on user files. and skips the prettier/dprint stuff.

packages/integration-tests/src/entities/Author.ts
brudil: Mentioned it above, I'll try out recast because I'm thinking if we can get perfect comment insertion without touching the rest of the line, we should be able to avoid all this.

packages/integration-tests/src/entities/UserCodegen.ts
stephenh: Oh, I see, thanks for the link! I thought "merged" was like "appended", but if they're the same, it's deduped, so then, yeah, seems like not a big deal to put them on either getters-only or setters-only or what not... Probably fine to leave as-is if/until we have a reason to do something different.

packages/integration-tests/src/entities/Author.ts
stephenh: Cool! Fwiw a little curious to see how recast handles it, b/c even if it parses/outputs the existing code with 100% accuracy (which will be great), we'll still be inserting net-new/changed code, which will probably? have to go through a formatter? Maybe?

Or maybe if recast keeps everything basically as-is, and only our ~few lines of new comments are what "need formatting", then it won't matter as much whether we use prettier or dprint? Not sure...

brudil:
Hey, just to say I haven't forgotten this. Been busy elsewhere, but hope to find time to get this over the line once I'm back off holiday.

stephenh:
Great to hear from you @brudil ! Definitely understood getting busy; all of Joist has been built in fits & spurts over the years :-), so whenever you can pick this back up again would be great/np.

I am curious, how's your Objection iirc to Joist migration going? Anything running in production yet? Understood switching ORMs is a huge endeavor for a codebase, so definitely np if you're still prototyping/trying to find time/etc., but would enjoy hearing if you did actually get Joist into production. :crossed_fingers: :-)

stephenh:
@brudil ...a few years later... (cue the spongebob meme :sweat_smile: ) I just landed essentially this feature in #1749 -- we haven't turned it on in our internal repo yet, but looking forward to it!

PR #719, opened 2023-07-02
Fixing a typo

davekerber:


stephenh:
:wave: I think we already merged this fix: https://github.com/stephenh/joist-ts/pull/622

PR #721, opened 2023-07-03
fix(docs): Adding documentation about how to check for the presence of relations and get relation ids.

davekerber:
Adding documentation for some things that caught me up in the beginning of using Joist.  Especially coming from a Rails background. 

Looks like some trailing whitespace was automatically deleted too.


stephenh (approved):
👍

PR #724, opened 2023-07-06
fix: Refactor reactivity

### LINKED ISSUE #726: Missed reactivity hints
Right now we detect "field A has changed, so rerun downstream fields X/Y/Z" in the middle of the `em.flush` "pending entities" loop.

However, there are a few issues with this:

1. Entities from an earlier loop but changed by 2nd loop hooks won't recalc
   * a1 is changed, and goes through loop 1--a1.firstName has not changed, so we don't recalc its downstreams'
   * On loop 2, some other entity's hook changes a1.firstName (which is just a primitive)
2. Hooks see an initially unset value for derived values, b/c we assume hooks should "go first" so that we see their mutations, to later recalc off those mutations
   * But this is kind of a false assumption anyway, b/c we need to watch more mutations than just those from "the hooks that go first"

Seems like ideally we would move "fields that need downstreams recalced" out of the flush loop, or at least not as reliant on it.

We could also potentially leverage setters, i.e. this primitive || derived field changed, oh look it has downstreams, to pre-emptively queue reactive fields that need recalced.

Given that `ReactiveField`s are init-time config data, we could also use them as map keys/set entries to know whether we've already recalced them, to avoid loops.

haislip:
Fixes #726


packages/orm/src/EntityManager.ts
```diff
@@ -1593,61 +1584,6 @@ function recalcDerivedFields(todos: Record<string, Todo>) {
...
-  });
-  await Promise.all(p);
-}
-
-/**
- * Walks `reverseHint` for every entity in `entities`.
- *
- * I.e. given `[book1, book2]` and `["author", 'publisher"]`, will return all of the books' authors' publishers.
- */
-async function followReverseHint(entities: Entity[], reverseHint: string[]): Promise<Entity[]> {
```
stephenh, line 1613: Moved this to `revertHints.ts`

packages/orm/src/config.ts
```diff
@@ -145,24 +145,33 @@ export class ConfigApi<T extends Entity, C> {
  * a `ReactiveRule` with fields `["title"]`, path `books`, and rule `ruleFn`.
  */
 interface ReactiveRule {
```
stephenh, line 147: No changes here, just comments

packages/orm/src/reactiveHints.ts
```diff
@@ -150,6 +151,49 @@ export function reverseReactiveHint<T extends Entity>(
   ];
 }
 
+/**
+ * Walks `reverseHint` for every entity in `entities`.
+ *
+ * I.e. given `[book1, book2]` and `["author", 'publisher"]`, will return all of the books' authors' publishers.
+ */
+export async function followReverseHint(entities: Entity[], reverseHint: string[]): Promise<Entity[]> {
```
stephenh, line 159: Just moved from `EntityManager.ts`, no changes

packages/orm/src/EntityManager.ts
```diff
@@ -1065,6 +1083,13 @@ export class EntityManager<C = unknown> {
 
     this._isFlushing = true;
 
+    const touched = this.entities.filter((e) => e.__orm.isTouched);
```
stephenh, line 1086: `em.touch` used to go through `Todo.asyncFields`, but since that went away to decouple reactivity from the em.flush loops, we just handle touching here explicitly.

packages/orm/src/EntityManager.ts
```diff
@@ -1065,52 +1083,62 @@ export class EntityManager<C = unknown> {
...
 
+      // Recreate todos now that we've run hooks and recalculated fields so know
+      // the full set of entities that will be INSERT/UPDATE/DELETE-d in the database.
+      const entitiesToFlush = [...hooksInvoked];
       const entityTodos = createTodos(entitiesToFlush);
 
       if (!skipValidation) {
         try {
           this._isValidating = true;
-          await addReactiveValidations(entityTodos);
```
stephenh, line 1113: Since `Todo.asyncFields` went away + `validates` had no reason to be part of the `Todo` data structure, I inlined `addReactiveValidations` into `validateReactiveRules`.

haislip (commented):
This solution looks great! Thanks!

PR #729, opened 2023-07-13
fix: orderBy on cti base fields 

brudil:
All order by properties which existed in `meta.allFields` were built against the root alias, rather than the base type table the field belonged to.

- Added repro
- Looks like aliasPrefix is the cleanest way to do this, but haven't been around QueryParser too much yet!


brudil:
@stephenh Unsure if I could have introduced the failing test build within 18.10.0 here.

stephenh (approved):
Looks good!

...yeah, that test failure is flakey, which is weird b/c usually we liberally add extra "order by IDs", etc, to ensure determinism. Havent dug into that yet.

PR #731, opened 2023-07-15
chore: Clean up EntityManager's private fields.

stephenh:
Potential alternatives:

1. Current version, which is `#` private fields, exposed via a separate `__api` object
2. Stay with TS private fields, and have `getEmInternalApi` just return `em` directly casted, b/c the field names line up and callable, i.e.:

```
interface InternalApi {
  someField: string;
}

class Foo {
  private someField: string;
}

function getInternalApi(foo: Foo) {
  return foo as InternalApi;
}
```

### DIFF
--- a/packages/integration-tests/src/relations/ManyToManyCollection.test.ts
+++ b/packages/integration-tests/src/relations/ManyToManyCollection.test.ts
@@ -31,7 +31,7 @@ describe("ManyToManyCollection", () => {
     expect(tag.books.get.length).toEqual(1);
     expect(book.tags.get[0]).toStrictEqual(tag);
     expect(tag.books.get[0]).toStrictEqual(book);
-    expect(em.__data.joinRows["books_to_tags"].length).toEqual(1);
+    expect((em as any).__api.joinRows["books_to_tags"].length).toEqual(1);
   });
 
   it("can load a many-to-many with constant queries", async () => {
--- a/packages/orm/src/EntityManager.ts
+++ b/packages/orm/src/EntityManager.ts
@@ -218,6 +218,8 @@ export class EntityManager<C = unknown> {
       this.ctx = emOrCtx!;
       this.driver = driver!;
     }
+
+    // Expose some of our private fields as the EntityManagerInternalApi
     const em = this;
     this.__api = {
       joinRows: this.#joinRows,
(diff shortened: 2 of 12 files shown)


packages/integration-tests/src/relations/ManyToManyCollection.test.ts
```diff
@@ -31,7 +31,7 @@ describe("ManyToManyCollection", () => {
     expect(tag.books.get.length).toEqual(1);
     expect(book.tags.get[0]).toStrictEqual(tag);
     expect(tag.books.get[0]).toStrictEqual(book);
-    expect(em.__data.joinRows["books_to_tags"].length).toEqual(1);
```
stephenh, line 34: This was the only place outside of `orm/src` we were poking at the internal API, and it was just for an assertion, so went ahead and kept it with a cast.

packages/orm/src/EntityManager.ts
```diff
@@ -179,15 +179,14 @@ export class EntityManager<C = unknown> {
   public readonly ctx: C;
   public driver: Driver;
   public currentTxnKnex: Knex | undefined;
-  private _entities: Entity[] = [];
+  #entities: Entity[] = [];
```
stephenh, line 182: Moving the mixed "some private _foo", "some private foo", some __data: { foo }" all over to `#foo`.

packages/orm/src/EntityManager.ts
```diff
@@ -196,17 +195,17 @@ export class EntityManager<C = unknown> {
...
-  private hooks: Record<EntityManagerHook, HookFn[]> = {
-    beforeTransaction: [],
-    afterTransaction: [],
+  #pendingCascadeDeletes: Entity[] = [];
+  #dataloaders: Record<string, LoaderCache> = {};
+  #joinRows: Record<string, JoinRow[]> = {};
+  /** Stores any `source -> downstream` reactions to recalc during `em.flush`. */
+  #rm = new ReactionsManager();
+  #hooks: Record<EntityManagerHook, HookFn[]> = { beforeTransaction: [], afterTransaction: [] };
+  private __api: EntityManagerInternalApi = {
```
stephenh, line 204: This `__api` stays TS private / not-`#` so that our `getEmInternalApi` can grab it.

PR #741, opened 2023-07-29
feat: Rename entity to fullNonReactiveAccess.

stephenh:
To highlight it's non-reactive-ness.


chr1sjf0x (approved):
Looks good, although I did like the suggestion to use `footGun` 😆

PR #748, opened 2023-08-05
feat: Add reactivity for m2m relations

### LINKED ISSUE #746: Add reactivity support for m2m relations

stephenh:
Adds reactivity for m2m relations by:

* Tweaking `reverseReactiveHint` to put m2m collections into the `fields` array
* Refactored a lot logic out of `ManyToManyCollection` into a `JoinRows` manager-ish class that watches for the explicit "add new join row" / "flag old join row for deletion", and queues reactive fields to be recalced
* Updated the `followReverseHint` to expand m2m collections to both "new values" (already happens for free via `.load()`) as well as "old values" (by asking the `JoinRows` manager for any flagged-for-deletion rows for the m2m relation)
  * This is basically how the FK reactivity works as well, whenever a FK changes, we have to crawl through both the new FK value as well as the old FK value, to ensure both are updated
  * Maybe we could generalized this concept of "old value/new value" into the Reference/Collection impls themselves, instead of hard-coding one-off in `followReverseHint`, but :shrug:, I think the current setup is fine, and the decoupling might actually be nice...

Fixes #746


packages/orm/src/EntityManager.ts
```diff
@@ -179,7 +181,7 @@ export class EntityManager<C = unknown> {
    */
   #pendingCascadeDeletes: Entity[] = [];
   #dataloaders: Record<string, LoaderCache> = {};
-  #joinRows: Record<string, JoinRow[]> = {};
+  #joinRows: Record<string, JoinRows> = {};
```
stephenh, line 184: Instead of having everyone just read/write to a `JoinRow[]` with abandon, starting to abstract a `JoinRows` abstraction around it.

packages/orm/src/reactiveHints.ts
```diff
@@ -180,22 +201,24 @@ export async function followReverseHint(entities: Entity[], reverseHint: string[
     // The path might touch either a reference or a collection
     const entitiesOrLists = await Promise.all(
       current.flatMap((c: any) => {
-        async function maybeLoadedPoly(loadPromise: Promise<Entity>) {
```
stephenh, line 183: Just moved this to a function at the bottom

packages/orm/src/relations/ManyToManyCollection.ts
```diff
@@ -118,27 +118,15 @@ export class ManyToManyCollection<T extends Entity, U extends Entity>
     ensureNotDeleted(this.#entity);
 
     if (this.loaded !== undefined) {
-      if (this.loaded.includes(other)) {
-        return;
-      }
```
stephenh, line 123: Just condensing a few lines

packages/orm/src/relations/ManyToManyCollection.ts
```diff
@@ -147,21 +135,7 @@ export class ManyToManyCollection<T extends Entity, U extends Entity>
     ensureNotDeleted(this.#entity, "pending");
 
     if (!percolated) {
-      const joinRows = getOrSet(getEmInternalApi(this.#entity.em).joinRows, this.joinTableName, []);
-      const row = joinRows.find((r) => r[this.columnName] === this.#entity && r[this.otherColumnName] === other);
```
stephenh, line 151: Moved this logic into JoinRows

zgavin (approved):
Looks great, awesome addition

PR #775, opened 2023-09-05
fix: CTI BaseEntity.toJSON skipping parent fields

CalirDeminar:
Observed the below behaviour when serialising CTI entities with toJSON, of skipping fields of the parent class:
joist-cti-logging-issue-demo

On the assumption that this isn't the desired behaviour:
* Added a test case to repro + catch this
* Updated the toJSON function to use metadata.allFields instead of metadata.fields


stephenh (approved):
Looks great @CalirDeminar , thank you!

Fwiw I just want to acknowledge that the current `toJSON` is extremely naive/simple; we use GraphQL to produce all of our API responses, so have not needed Joist itself to do anything fancier, but I anticipate that surely at some point Joist will/should have a more robust `toJSON` that can like control what fields are output, do multiple levels of nested output, etc... we'll see!

PR #777, opened 2023-09-06
feat: Codegen strict-mode

CalirDeminar:
Introduces a new joist-config.json option, `strictMode`, defaulting to `false`, where warnings what would normally just be logged will instead throw. 
An example case being CI/DI running codegen and warning of a non-deferred FK, would probably want to throw rather than continue and knowingly deploy code that's going to fail on INSERT and UPDATE queries.

Non-deferred FKs are the principle target here. I do see there are additional logged warnings around the config containing references to non-existent fields. I've currently left them not using this feature, but that does feel a bit inconsistent with it being "strict". However those do not seem as actively dangerous as a non-deferred FK, so am open to input as to if those should should throw or not.

There's some interesting questions around testing this. I've extracted the feature into it's own tiny function that's easily tested. But I didn't see a decent place or way to fully test it with an integration style test within codegen, and am open to ideas on that (assuming it's even felt as necessary).

### DIFF
--- a/packages/codegen/src/EntityDbMetadata.ts
+++ b/packages/codegen/src/EntityDbMetadata.ts
@@ -27,6 +27,7 @@ import {
   mapSimpleDbTypeToTypescriptType,
   parseOrder,
   tableToEntityName,
+  logWarning,
 } from "./utils";
 
 /** All the entities + enums in our database. */
@@ -508,9 +509,7 @@ function newManyToOneField(config: Config, entity: Entity, r: M2ORelation): Many
   const ignore = isFieldIgnored(config, entity, fieldName, notNull, column.default !== null);
   // Make sure the constraint is deferrable
   if (!r.foreignKey.isDeferred || !r.foreignKey.isDeferrable) {
-    console.log(
-      `WARNING: Foreign key ${r.foreignKey.name} is not DEFERRABLE/INITIALLY DEFERRED, see https://joist-orm.io/docs/getting-started/schema-assumptions#deferred-constraints`,
-    );
+    logWarning(config, `WARNING: Foreign key ${r.foreignKey.name} is not DEFERRABLE/INITIALLY DEFERRED, see https://joist-orm.io/docs/getting-started/schema-assumptions#deferred-constraints`);
   }
   const derived = fkFieldDerived(config, entity, fieldName);
   return { kind: "m2o", fieldName, columnName, otherEntity, otherFieldName, notNull, ignore, derived, dbType };
--- a/packages/codegen/src/config.test.ts
+++ b/packages/codegen/src/config.test.ts
@@ -1,5 +1,6 @@
 import { Config, defaultConfig, FieldConfig, isFieldIgnored } from "./config";
 import { makeEntity } from "./EntityDbMetadata";
+import { logWarning } from "./utils";
 
 describe("config", () => {
   describe("isFieldIgnored", () => {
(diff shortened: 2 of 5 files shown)


packages/codegen/src/EntityDbMetadata.ts
```diff
@@ -508,9 +509,7 @@ function newManyToOneField(config: Config, entity: Entity, r: M2ORelation): Many
   const ignore = isFieldIgnored(config, entity, fieldName, notNull, column.default !== null);
   // Make sure the constraint is deferrable
   if (!r.foreignKey.isDeferred || !r.foreignKey.isDeferrable) {
-    console.log(
-      `WARNING: Foreign key ${r.foreignKey.name} is not DEFERRABLE/INITIALLY DEFERRED, see https://joist-orm.io/docs/getting-started/schema-assumptions#deferred-constraints`,
-    );
+    logWarning(config, `WARNING: Foreign key ${r.foreignKey.name} is not DEFERRABLE/INITIALLY DEFERRED, see https://joist-orm.io/docs/getting-started/schema-assumptions#deferred-constraints`);
```
stephenh, line 512: Ah yeah, good call @CalirDeminar , that it would be bad to let code slip out against a schema with unconfigured FKs.

Fwiw I'm thinking we could probably just _always_ fail this, and not have the config flag?

I think my original reasoning for now blowing up right here, i.e. this line being a `throw new Error`, was for DX purposes, if a user has, say, 20 FKs that need fixed, I wanted joist-codegen to output all 20 of them that needed fixing, instead of "run once, blows up, fix it, run again, find the next FK to fix, fix it, run again, etc".

That said, I think we could achieve both goals, a good DX of "listing all invalid FKs" but still "make sure this is fatal, but doing something like:

* Here on this line, always do a `console.warn`
* On this line, also set a flag somewhere that says `foundMisconfiguredFk = true`
* At the end of `joist-codegen`, check `foundMisconfiguredFk`, and if it's true, throw a error that results in a non-zero return code

And then just always do this, even on a dev's local machine/without the strict config flag.

Granted, maybe we'll still end up with the concept of a strict mode at some point, but I think this approach would achieve the immediate goals. Wdyt?

CalirDeminar: That seems sensible to me. Do you have any particular opinions on if a failing codegen should output any files? 
My initial thought is we probably shouldn't output the files, so it's obvious that something's wrong. I could see on the development side it being very easy to miss the non-zero return code if we did still output the codegenned files

stephenh: Ah good question, I didn't think about that...

My assumption was that we should emit code, so that the developer can keep kind of working while they fix the broken FKs.

But you're right that'd potentially be easy to miss...

Dunno, maybe it's too cute/sneaky, but we could put something in `metadata.ts` that is like:

```
throw new Error("You have misconfigured FKs a/b/c...");
```

To get maybe the best of both worlds: the dev has some WIP output to start compiling/working against, but they can't actually boot anything up.

...granted, not sure if we can do a `throw new Error` as a top-level statement in the `metadata.ts` module. Maybe just use an IIFE? Wdyt?

CalirDeminar: hmm, that's true. Returning something broken "feels" wrong, but then not doing that I guess that could leave someone with a broken build that stops them from build the fix, stuck with the proverbial Chicken + Egg. 
Putting a throw into `metadata.ts` feels a bit like "Poisoning the Well", but if it's labelled clearly I guess should be dev friendly enough and wouldn't block a build.

I don't *think* we should have a problem with throwing right out of the gate at the top level in `metadata.ts`, so that's probably the way to go 👍

PR #782, opened 2023-09-12
feat: Codegen - Delayed throw on non-deferred FK Found

CalirDeminar:
A continuation / rework of: https://github.com/stephenh/joist-ts/pull/777

Summary: 
* Move to a console.warn for the warning around the FK not being deferred
* Crash with a non-zero exit code if any entities are deemed invalid for having an non-deferred FK
    * In this case also insert a throw within `metadata.ts` to ensure the generated code won't run in an unsafe manner despite the codegen throw

### DIFF
--- a/packages/codegen/src/index.ts
+++ b/packages/codegen/src/index.ts
@@ -54,7 +54,7 @@ if (require.main === module) {
 
     const entityTables = db.tables.filter((t) => isEntityTable(config, t)).sortBy("name");
     const entities = entityTables.map((table) => new EntityDbMetadata(config, table, enums));
-    if(entities.some((entity) => entity.invalidDeferredFK)){
+    if (entities.some((entity) => entity.invalidDeferredFK)) {
       safeExit = false;
     }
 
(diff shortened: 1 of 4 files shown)


davekerber:
Just a thought. It is possible to use Joist without deferred constraints, it's just harder.  Throwing an error completely could be a blocker for existing databases that want to use Joist.

packages/codegen/src/index.ts
```diff
@@ -89,6 +93,9 @@ if (require.main === module) {
     warnInvalidEntries(config, dbMetadata);
 
     await generateAndSaveFiles(config, dbMetadata);
+    if (!safeExit){
```
stephenh, line 96: Nit ask, can you run the formatter on this file?

stephenh (approved):
Nice, lgtm @CalirDeminar , thanks!

stephenh:
> It is possible to use Joist without deferred constraints, it's just harder.

Ah right, @davekerber I forgot you'd used Joist for awhile w/o realizing it expected to have deferred constraints; what did you have to do for that to work? Just have multiple em.flushes, so that any dependency was flushed first?

I dunno, my assumption is that the DX of that (running w/o deferred constraints) is annoying enough that I'd prefer to ask users to fix their FKs. I believe there is a script-ish in the docs that helps automate that, also its probably not super-easy to run, like I think it assumes you have pg-structure as the npm library, and it's not just a single-line SQL command, which would be ideal.

Eventually it's also possible to just support DBs w/o deferred constraints, by correctly recognizing the order rows need to be flushed in, but that's "hard-ish" (granted, just topo sort them based on dependencies) and adds some steps/complexity that are nice to avoid vs. "just sent all the writes as a single batch and it'll get figured out".

stephenh:
Looks great, thanks @CalirDeminar !

PR #783, opened 2023-09-12
fix: Allow query by deep poly o2m field

allangsilva:


packages/orm/src/QueryParser.ts
```diff
@@ -250,7 +250,13 @@ export function parseFindQuery(
...
         } else if (field.kind === "o2m") {
           const a = getAlias(field.otherMetadata().tableName);
-          const otherColumn = field.otherMetadata().allFields[field.otherFieldName].serde!.columns[0].columnName;
+          const otherField = field.otherMetadata().allFields[field.otherFieldName];
+          let otherColumn = otherField.serde!.columns[0].columnName;
+          // If the other field is a poly, we need to find the right column
+          if (otherField.kind === "poly" && otherField.components.some((c) => c.otherMetadata() === meta)) {
+            // For a subcomponent that matches field's metadata
+            otherColumn = otherField.components.find((c) => c.otherMetadata() === meta)!.columnName;
+          }
```
stephenh, line 259: Really minor, but could this be:

```suggestion
          if (otherField.kind === "poly") {
            const otherComponent = otherField.components.find((c) => c.otherMetadata() === meta)) ?? fail("No poly component found for ${otherField.name}");
            otherColumn = otherComponent.columnName;
          }
```

allangsilva: Looks better, thanks 👏

stephenh (approved):

PR #786, opened 2023-09-14
fix: skip relation cleanup on cascade delete

zgavin:


packages/orm/src/relations/PolymorphicReference.ts
```diff
@@ -247,9 +252,10 @@ export class PolymorphicReferenceImpl<T extends Entity, U extends Entity, N exte
   }
 
   public toString(): string {
-    return `PolymorphicReference(entity: ${this.entity}, fieldName: ${this.fieldName}, otherType: ${
-      this.currentComponent?.otherMetadata().type
-    }, otherFieldName: ${this.currentComponent?.otherFieldName}, id: ${this.id})`;
+    return `PolymorphicReference(entity: ${this.entity}, fieldName: ${
```
zgavin, line 255: think this is just a format update?

stephenh (approved):
Surprised the fix was this easy; fwiw I'm fine to ship to unblock a bug fix, but would strongly ask/request a follow up test b/c I'm like 80% sure the cascade delete logic will get refactor sometime in the next ~year, and 80% sure I'll regress this behavior with a test to make sure I don't. :-)

PR #791, opened 2023-09-17
feat: Change id to behave like idOrFail.

stephenh:
We've had issues of developers "shooting themselves in the foot" by writing code like:

```ts
config.beforeFlush(book => {
  if (book.id !== someOtherThing.book.id) {
    // run different books
  }
});
```

And not realizing that `id` returns `undefined` for new entities, so `// run different books` logic _won't_ run when `book` and `someOtherThing.book` are different books (`b1` and `b2`), but are both new, and so have undefined ids (`undefined !== undefined` is false).

This foot gun seems to happen the most in hooks, but in general it can happen in any pre-`flush` code that is trying to work with/do comparison on ids of potentially-newly-created entities.

The worst thing about it, is that the code often "works in tests", where the entities all happen to be new, but then will fail in production/or other boundary cases that come up later.

Ironically, Joist's unit-of-work/identity cache is a great solution to this, b/c if you just lean into object identities, this code works all the time:

```ts
config.beforeFlush(book => {
  if (book !== someOtherThing.book.get) {
    // business logic
  }
});
```

However, this often requires an extra load hint to get `someOtherThing.book.get` populated/added to the type, and so developers will shy away from that option, as a premature but ill-fated optimization.

So, this PR flips the behavior of `.id` to blow up if the id is not set.

Specifically:

* We previously had `.id`, which was typed as `string | undefined` and would return `undefined` on new entities
* We previously had `.idOrFail`, which was typed as `string` and would blow up on new entities

We're flipping them, so now:

* We have `.id` typed as `string` and it will blow up on new entities
* We added `.idMaybe` typed as `string | undefined` to be the "call id and return undefined if new" method
* We've removed `.idOrFail` since it's replaced by `.id`

The rationale is that `.idMaybe` can be used for situations when you're aware of & opt-in to the "might return `undefined` behavior", but that otherwise the "`id` that people call without really thinking about it" has dramatically safer/safe-by-default behavior.

### DIFF
--- a/docs/docs/testing/test-utils.md
+++ b/docs/docs/testing/test-utils.md
@@ -36,7 +36,7 @@ This means we can immediately assert against `a.books.get` without needing to lo
 const a1 = newAuthor(em);
 await performPostBook(em);
 // Example of what we _don't_ need to do: reload a1
-await a1_2 = em.load(Author, a1.idOrFail);
+await a1_2 = em.load(Author, a1.id);
 expect(a.books.get.length).toEqual(1);
 ```
 
--- a/packages/codegen/src/generateEntityCodegenFile.ts
+++ b/packages/codegen/src/generateEntityCodegenFile.ts
@@ -361,10 +361,10 @@ export function generateEntityCodegenFile(config: Config, dbMeta: DbMetadata, me
   const factoryMethod = imp(`new${entity.name}@./entities`);
   const EntityManager = imp("t:EntityManager@./entities");
 
-  const idCode =
+  const idMaybeCode =
     config.idType === "untagged-string"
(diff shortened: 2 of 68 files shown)


stephenh:
Fyi @davekerber and @brudil , heads up that this is a breaking change we'll probably land soon (see PR description for the rationale).

In theory codebases shouldn't have a lot of `idOrFail`s (which are getting removed, and can become just `id`; any `id` usages that still want to work w/new entities need to be `idMaybe` now), and if so they are trivial to fix.

I haven't actually tried running on our codebase yep; I am anticipating ~100s of fixes, but that it will be very quick/boilerplate to do...

zgavin (approved):

PR #814, opened 2023-10-03
feat: batch assignNewIds

brudil:
If batching is desired, this might be a solution.

### DIFF
--- a/packages/orm/src/EntityManager.ts
+++ b/packages/orm/src/EntityManager.ts
@@ -1063,9 +1063,9 @@ export class EntityManager<C = unknown> {
     this.#pendingCascadeDeletes.push(entity);
   }
 
-  assignNewIds() {
+  async assignNewIds() {
     let pendingEntities = this.entities.filter((e) => e.isNewEntity && !e.isDeletedEntity && !e.idMaybe);
-    return this.getLoader<string, string>("assign-new-ids", "global", async (entityTestIds) => {
+    await this.getLoader<string, string>("assign-new-ids", "global", async (entityTestIds) => {
       let todos = createTodos(entityTestIds.map(id => pendingEntities.find(e => e.toString() === id)!));
       await this.driver.assignNewIds(this, todos);
       return entityTestIds;
--- a/packages/orm/src/EntityManager.ts
+++ b/packages/orm/src/EntityManager.ts
@@ -1063,10 +1063,13 @@ export class EntityManager<C = unknown> {
     this.#pendingCascadeDeletes.push(entity);
   }
 
-  async assignNewIds() {
+  assignNewIds() {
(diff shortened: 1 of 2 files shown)


packages/orm/src/EntityManager.ts
```diff
@@ -1063,10 +1063,13 @@ export class EntityManager<C = unknown> {
...
 
-  async assignNewIds() {
+  assignNewIds() {
     let pendingEntities = this.entities.filter((e) => e.isNewEntity && !e.isDeletedEntity && !e.idMaybe);
-    let todos = createTodos(pendingEntities);
-    return this.driver.assignNewIds(this, todos);
+    return this.getLoader<Entity, void>("assign-new-ids", "global", async (entities) => {
+      let todos = createTodos(Array.from(entities));
+      await this.driver.assignNewIds(this, todos);
+      return [];
```
stephenh, line 1071: I think the dataloader contract expects this array length to match the 'entities' array length, do maybe we could return some dummy values, or even just return entities.

packages/orm/src/EntityManager.ts
```diff
@@ -1063,10 +1063,13 @@ export class EntityManager<C = unknown> {
...
+    }).loadMany(pendingEntities);
```
stephenh, line 1072: I'm only on mobile so I'm not 100% sure but my guess is we might need to pass primitives like maybe pendingEntities.map(e => e.toString())?

We'd want that to be something stable, which I think toString would be...

stephenh (approved):
Hey @brudil ! Yep, I agree this should be batched. I haven't pulled up the build failure but once this is passing tests I think it's great. Thanks!

packages/orm/src/EntityManager.ts
stephenh: No style guide, but I think we usually do '|| fail(...)' to avoid the bangs.

Mapping back to entities makes sense. Fwiw maybe do a lookup in a map to avoid the scans that could be a N^2.

I think we have a groupBy utility method for making the list into a map.

stephenh: Oh yeah, instead of passing IDs plus look it up in a map, we can probably pass a custom hash function to the getLoader function ... I don't have a link handy, but some of our other getLoader calls do that, like in the findDataLoader iirc

brudil: I think due to the nature of the em, a cache key isn't actually required. Each entity is only going to exist once and internally Dataloader uses a Map<>. Happy to use `cacheKeyFn` though if desired

stephenh: Ah sure, that makes sense!

stephenh (approved):
Looks great, thanks!

PR #829, opened 2023-10-13
fix: Fix AsyncMethods w/no params looking like AsyncProperties.

stephenh:


packages/integration-tests/src/relations/hasAsyncMethod.test.ts
```diff
@@ -49,4 +49,24 @@ describe("hasAsyncMethod", () => {
...
+    await insertAuthor({ first_name: "a1" });
+    await insertBook({ author_id: 1, title: "programming in action" });
+    // Given a stock author
+    const em = newEntityManager();
+    const a1 = await em.load(Author, "a:1");
+    // When we access an async method
+    expect(await a1.booksTitles.load()).toBe("programming in action");
+  });
+
+  it("without params can be accessed via get", async () => {
```
chr1sjf0x, line 63: question: is this test name right?
```suggestion
  it("with params can be accessed via get", async () => {
```

chr1sjf0x (approved):

packages/integration-tests/src/relations/hasAsyncMethod.test.ts
stephenh: I think it's right? We're calling `.get()` without any params.

PR #837, opened 2023-10-24
fix: complex filters on cti of parent fields do not filter to correct table

brudil:
Getting up some cases here to show the problem - can move them somewhere more appropriate once the fix is in - it's the last two.

Essentially it looks like complex filtering don't take in to account that they might be filtering on the parent table.

See how a complex condition (using either the Publisher or SmallPublisher as the alias) against `name`, which exists on the publisher table, has the filter condition's alias set to `sp` - i.e the child table.

```js
{
      selects: [ 'sp.*', 'sp_b0.*', 'sp.id as id' ],
      tables: [
        { alias: 'sp', table: 'small_publishers', join: 'primary' },
        {
          alias: 'sp_b0',
          table: 'publishers',
          join: 'outer',
          col1: 'sp.id',
          col2: 'sp_b0.id',
          distinct: false
        }
      ],
      conditions: [],
      orderBys: [ { alias: 'sp', column: 'id', order: 'ASC' } ],
      complexConditions: [
        {
          op: 'and',
          conditions: [
            {
              alias: 'sp',
              column: 'name',
              dbType: 'character varying',
              cond: { kind: 'eq', value: 'p1' }
            }
          ]
        }
      ]
    }
```

Haven't fiddled around with the QueryParser yet, so I'm having a look, @stephenh any pointers lmk!

### DIFF
--- a/packages/integration-tests/src/EntityManager.ctiQueries.test.ts
+++ b/packages/integration-tests/src/EntityManager.ctiQueries.test.ts
@@ -1,0 +1,98 @@
+import {newEntityManager} from "@src/setupDbTests";
+import {Publisher, SmallPublisher} from "@src/entities";
+import {alias} from "joist-orm";
+
+describe('EntityManager cti queries', () => {
+
+  it('finds against child with simple parent filter', async () => {
+    const em = newEntityManager();
+
+    em.create(SmallPublisher, {
+      name: 'p1',
+      city: 'location'
+    });
+
+    await em.flush();
+
+    const res = await em.find(SmallPublisher, { name: 'p1' })
+
+    expect(res.length).toBe(1);
+  });
+
+  it('finds against parent with simple parent filter', async () => {
+    const em = newEntityManager();
+
+    em.create(SmallPublisher, {
+      name: 'p1',
+      city: 'location'
+    });
+
+    await em.flush();
+
+    const res = await em.find(Publisher, { name: 'p1' })
+
+    expect(res.length).toBe(1);
+  });
+
+  it('finds against child with simple child filter', async () => {
+    const em = newEntityManager();
+
+    em.create(SmallPublisher, {
+      name: 'p1',
+      city: 'location'
+    });
+
+    await em.flush();
+
+    const res = await em.find(SmallPublisher, { city: 'location' })
+
+    expect(res.length).toBe(1);
+  });
+
+  it('finds against child with simple child & parent filter', async () => {
+    const em = newEntityManager();
+
+    em.create(SmallPublisher, {
+      name: 'p1',
+      city: 'location'
+    });
+
+    await em.flush();
+
+    const res = await em.find(SmallPublisher, { name: 'p1', city: 'location' })
+
+    expect(res.length).toBe(1);
+  });
+
+  it('finds against child with complex child & parent filter, alias child', async () => {
+    const em = newEntityManager();
+
+    em.create(SmallPublisher, {
+      name: 'p1',
+      city: 'location'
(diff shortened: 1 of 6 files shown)


stephenh:
Hey @brudil ! Okay, I think we might get lucky and this one won't be too bad...

* In an entity's metadata, there is both `fields` and `allFields`, where `allFields` also has fields from the subtype
* Fields within `allFields` also have an `aliasSuffix`, that is like `_b0` that is this ~statically known hint/way of knowing what concrete alias to use for CTI/base fields
* The `complexConditions` / `const p = alias(Publisher)` management is a little wonky right now, b/c the `alias` creates aliases ahead of time, and then later when we're talking the join tree, we bind we declared alias variable to its actual abbreviation
* This is done by the `setAlias` method that looks like:

```
  const mgmt: AliasMgmt = {
    tableName: meta.tableName,
    setAlias(newAlias: string) {
      conditions.forEach((c) => (c.alias = newAlias));
    },
  };
```

Basically what needs to happen is that that `forEach` needs to be: 

```
      conditions.forEach((c) => (c.alias = `${newAlias}${aliasSuffix}`));
```

* The wrinkle is that we don't have ready access to the `aliasSuffix` for the current column, so I think this array of conditions:

```
  const conditions: ColumnCondition[] = [];
```

Probably needs to be an array of `{ condition: ColumnCondition; field: fieldICameFrom }[]`

The `new PrimitiveAliasImpl` could take the `field` as a cstr arg, so that we could do:

```
  private addCondition(value: ParsedValueFilter<V>): ColumnCondition {
    const cond: ColumnCondition = {
      alias: "unset",
      column: this.column.columnName,
      dbType: this.column.dbType,
      cond: mapToDb(this.column, value),
    }; 
// change here
    this.conditions.push({ field: this.field, condition: cond });
    return cond;
  }
```

And that should get at least your current test case to work.

When I reference "we're lucky", what will be harder outside of this 1st query is things like:

* Filtering on sub-table fields (pretty sure they're not even in the types now--so even getting the `FullPublisher` type created will need some thought)
* Filtering on CTIs when they're not the base `em.find` target, like if someone does `em.find(Author, { publisher: ...column in a base/sub table that is not the author.publisher_id FK target... })` that will likely require teaching `em.find` to add in the base/child table as another outer join

But we can tackle those later/as-needed, and I think the bullet points above should hopefully unblock your current use case without too much work. Let me know how it goes!

stephenh:
Hey @brudil , since it'd been a few days, and I needed to scratch an "just hack on something Joist related for fun" itch :-) I went ahead and tried out/pushed up my suggested approach...

It fixes one of your two failing tests--the one of using a child alias, and then filtering against a base column.

I'm not entirely sure how to fix the last failing test--using an `em.find(Child, ...)` but a "mismatched" alias of `alias(Parent)` ...

Actually a little bit of thinking, and I've got that passing too!

Want to take a look and :+1: ? Happy to have you add more tests for things we're still missing.

Thanks!

stephenh:
Going to go ahead and merge, but @brudil happy to see more tests/PRs of things that still don't work!

brudil:
Hey! Sorry, I was away last week. Didn't mean to just abandon it :) Cheers for this we'll give it a whirl!

stephenh:
Cool! Fwiw the latest release of Joist has some initial support for `number`-typed `id` fields ... I'm having some bumps landing that in our internal app, so if you see the same on your side, maybe stay back a few versions (i.e. this 1.114.1 w/these changes should be fine).

Hopefully should have those bumps smoothed out in the next few days.

PR #859, opened 2023-11-09
fix(isPolyHelper): allow pasing in `unknown` types to discern whether or not it's a valid type

TylerR909:
Fixes https://github.com/stephenh/joist-ts/issues/858

Expands Poly-helpers like `isCommentParent({})` to allow non-Entities to be passed in (which already worked, but TS wasn't too happy about it)


stephenh (approved):
:+1: nice!

PR #868, opened 2023-11-13
feat: Allow accessing ids in derived fields.

### LINKED ISSUE #808: Support async derived fields that want to use id on create
Something like:

```ts
  /** This string represents the values that will be indexed for search */
  readonly search: PersistedAsyncProperty<Project, string> = hasPersistedAsyncProperty(
    "search",
    {
      buildAddress: ["city", "street1", "street2", "postalCode"],
    },
    (project) => {
      const {
        buildAddress: { city, street1, street2, postalCode },
      } = withLoaded(project);
      return [project.idMaybe, street1, street2, city, postalCode].compact().join(" ");
    },
  );
```

Should be able to use the `project.id` right away, instead of missing it until the 2nd time the `Project` is updated.

stephenh:
If the id is not yet assigned, we'll save the relation until later, and pre-emptively assign the ids.

Fixes #808


packages/orm/src/relations/AbstractPropertyImpl.ts
```diff
@@ -0,0 +1,12 @@
+/** A base class for relations & non-relation (i.e. `PersistedAsyncProperty) properties. */
+export abstract class AbstractPropertyImpl<T> {
+  #entity: T;
+
+  constructor(entity: T) {
```
zgavin, line 5: question: what does using a symbol based field give us? could we just do 

```suggestion
  constructor(public readonly entity: T) {
```

zgavin (approved):

packages/orm/src/relations/AbstractPropertyImpl.ts
stephenh: Yeah, I'm like 70% sure that `public readonly entity` becomes as owned field, so shows up in `Object.keys` / etc., but that `#entity + getter` w/the getter defined on the class, won't show up.

Yeah:

```
> class Foo {
... a = 1
... #b = 2
... get b() { return this.#b; }
... }
undefined
> f = new Foo();
Foo { a: 1 }
> Object.keys(f);
[ 'a' ]
>
```

At some point, this was important to avoid Jest and/or JSON.stringify (probably Jest) crawling into "all the things" when doing diffs of failures/etc.

Not entirely sure how necessary it still is...

PR #869, opened 2023-11-13
fix: Fix base classes getting rules from sub classes.

stephenh:
Lots of `getAllMetas` should only be `getBaseAndSelfMetas`, i.e. we were pulling in validation rules that should have been subclass-only into base classes.


blambillotte (approved):
thanks for the fix!

stephenh:
FYI @brudil this was a pretty bad bug in our CTI support--I've added a regression to keep it from happening again, but would be surprised if you hadn't ran in to it. Apologies if so!

PR #894, opened 2023-11-23
fix: Include field name in reference errors.

stephenh:


packages/tests/integration/src/relations/ManyToOneReference.test.ts
```diff
@@ -28,7 +27,7 @@ describe("ManyToOneReference", () => {
     await insertAuthor({ first_name: "f" });
     const em = newEntityManager();
     const author = await em.load(Author, "1");
-    expect(() => author.publisher.id).toThrow("Reference is unset");
+    expect(() => author.publisher.id).toThrow("Reference publisher is unset");
```
TylerR909, line 30: Can you log the other side of this relationship too? Like `Author reference to publisher is unset`

stephenh: How about `Reference Author:1.publisher is unset`? Just `toString()`ing the entity will get type + the id, which seems useful.

PR #902, opened 2023-11-30
fix: fixed ordering of beforeDelete and cleanupOnEntityDeleted so relations will always remain until all beforeDeletes are called

zgavin:


stephenh (approved):
Sweet, love the test

PR #911, opened 2023-12-12
fix: Really check if relations are loaded for CustomCollection and CustomReference

chr1sjf0x:
Removing some more instances of cached `_isLoaded: boolean` fields becoming stale due to graph changes and actually checking instead.


stephenh (approved):
Nice!

PR #913, opened 2023-12-15
feat: Add config.setDefault API

### LINKED ISSUE #909: Allow Dynamic Defaults
In auditing hooks, we often have `beforeCreate`s and even `beforeUpdate`s that are just applying field defaults. We should have a better way of expressing these than "just a hook".

Maybe a single API:

```ts
config.setDefaults({
  firstName: (a) => ...,

  lastName: (a) => ...,
```

Or separate methods:

```ts
config.setDefault("firstName", a => ...);
```

* The `setDefaults` seems nicer for very small snippets, but the `setDefault` seems more scalable
* The `setDefault` also seems like it could facilitate dependencies between fields

Thoughts:

* When does this hook fire? Seems like it should be the very first operation, before `beforeCreate`

stephenh:
Provides first-class support for setting defaults, without relying on "just more hooks".

Fixes #909

### DIFF
--- a/packages/orm/src/index.ts
+++ b/packages/orm/src/index.ts
@@ -24,6 +24,7 @@ export { AliasAssigner } from "./AliasAssigner";
 export * from "./Aliases";
 export { BaseEntity } from "./BaseEntity";
 export { Entity, EntityOrmField, IdType, isEntity } from "./Entity";
+export * from "./EntityFields";
 export * from "./EntityFilter";
 export * from "./EntityGraphQLFilter";
 export * from "./EntityManager";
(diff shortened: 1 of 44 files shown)


packages/orm/src/index.ts
```diff
@@ -155,92 +156,105 @@ export function setOpts<T extends Entity>(
...
-  // If `values` is a string (i.e. the id), this instance is being hydrated from a database row, so skip all this,
-  // because `hydrate` manually calls `serde.setOnEntity`.
-  // If `values` is undefined, we're being called by `createPartial` that will do its own opt handling.
-  if (values === undefined || typeof values === "string") {
-    return;
-  }
   const { calledFromConstructor, partial } = opts || {};
-  const meta = getMetadata(entity);
-
-  Object.entries(values as {}).forEach(([key, _value]) => {
```
stephenh, line 167: This red looks like a lot of changes, but is just indentation by one level

PR #918, opened 2023-12-19
feat: Add AllEnumDetails to graphql codegen

blambillotte:
Adds a combined `AllEnumDetails` type to the `enums.graphl` codegen to make it easier to create a query resolver that can return the "details" of any enum

```graphql
type AllEnumDetails {
  enumFoo: [FooDetail!]! 
  enumBar: [BarDetail!]! 
}
```


stephenh:
@blambillotte I think the PR title check is catching that the commit message doesn't have `feat:` in it (when github "squad and merge" runs, it doesn't actually use the PR title afaiu.)

PR #919, opened 2023-12-21
feat: support baseClass polymorphics

0ces:
This PR solves an issue where polymorphic relationships from the baseClass cannot be used/assigned


packages/codegen/src/generateEntityCodegenFile.ts
```diff
@@ -594,7 +594,11 @@ function fieldHasDefaultValue(field: PrimitiveField | EnumField): boolean {
   );
 }
 
-function generatePolymorphicTypes(meta: EntityDbMetadata) {
+function generatePolymorphicTypes(meta: EntityDbMetadata, baseEntity?: EntityDbMetadata) {
+  // If this entity has a base class we will filter out any polymorphic fields that are from the base class
+  if (meta.baseClassName && baseEntity) {
+    meta.polymorphics = meta.polymorphics.filter((p) => !baseEntity.polymorphics.includes(p));
+  }
```
stephenh, line 601: @0ces just from reading the comment, it sounds like we were like including these fields in the output twice?

Like if `BasePublisher` had a poly field, the `SubPublisher` would also output it, and then that would break somehow? 

Would love to see some test coverage--looks like you were working on that, but then reverted?

0ces: Yeah, I set that in case in the `joist-config.json` we set something like:
```json
{
...
	"SubPub

Comment (0ces) [reply packages/codegen/src/generateEntityCodegenFile.ts:601]: Yeah, I set that in case in the `joist-config.json` we set something like this:

```json
{
	...
	"BasePublisher": { "relations": { "parent": { "polymorphic": true } }, "tag": "bp" },
	"SubPublisher": { "relations": { "parent": { "polymorphic": true } }, "tag": "sp" },
	...
}
```

It would just take the base poly, but it might be worth to just keep the child polymorphic, now that I think about it, it might be useful to be able to overwrite the polys from the base (?), I'll remove this and update the logic I added at the PolymorphicReference to select the correct one

Yes, I reverted the tests as there were more changes than I expected, I'll find another example instead of "ReviewComment", I was testing with the some BP-GQL-Server entities and it worked as expected but I'll test this here better

I'll convert this PR into a draft for now and will ask for a review when this gets into a better shape

stephenh: Hm, "overwrite polys from the base", do you mean something like:

* `BasePublisher.parent = Foo | Bar` (because of `base_publishers.parent_foo_id` and `base_publishers.parent_bar_id`), but
* `SubPublisher.parent = Foo | Bar | Zaz` (because of a new `sub_publishers.parent_zaz_id` that is a FK only added in the sub publisher table?

I could see that working, albeit it sounds kinda complicated. :-) But doable. What's the use case we're trying to use this over in graphql-service?

0ces: Yes, I don't see an example of overwriting the polymorphic relationship, but in the graphql-service we created a sub-entity for ToDos and the sub-entity's parent polymorphic feature wasn't working

packages/orm/src/relations/PolymorphicReference.ts
```diff
@@ -72,7 +72,10 @@ export class PolymorphicReferenceImpl<T extends Entity, U extends Entity, N exte
     private fieldName: keyof T & string,
   ) {
     super(entity);
-    this.field = getMetadata(entity).fields[this.fieldName] as PolymorphicField;
+    // If this polymorphic field is a subclass of another entity
+    const meta = getMetadata(entity);
+    const baseMeta = meta.baseTypes[0];
+    this.field = (baseMeta ?? meta).fields[this.fieldName] as PolymorphicField;
```
stephenh, line 78: Suggestion: It seems like `baseMeta ?? meta` will _always_ look at the base, when the field could still be on the subclass?

Fwiw I think changing the original `.fields[this.fieldName]` to just `.allFields[this.fieldName]` will probably work? `allFields` should already be setup have both base+sub fields merged together.

packages/tests/integration/schema/internalComment.graphql
```diff
@@ -0,0 +1,26 @@
+extend type Mutation {
+  saveInternalComment(input: SaveInternalCommentInput!): SaveInternalCommentResult!
+}
+
+type InternalComment {
```
stephenh, line 5: Suggestion: Fwiw since we don't really need it, can you delete all of the GraphQL scaffolding from this new entity? I've only recently started removing it (unnecessary GQL scaffolding in the tests), but it just ends up being boilerplate that we don't really need for Joist's integration tests.

packages/codegen/src/generateEntityCodegenFile.ts
stephenh, line 601: @0ces I'm not sure this new code is really doing anything, because:

a) `meta.polymorphics` should not have any base-class fields in it, i.e. the code here:

```
    this.polymorphics = polymorphicRelations(config, table).map((rc) =>
      newPolymorphicField(config, table, this.entity, rc),
    );
```

When running for, say, AdminUser, is not going to find any poly fields from the User table.

b) When I comment this out locally, and rerun `yarn codegen`, there are no code changes.

My guess is that the primary bug fix is the `fields` -> `allFields` change in `PolymorphicReference.ts`, and this change in `generateEntityCodegenFile` is not necessary?

packages/tests/integration/src/relations/PolymorphicReference.test.ts
```diff
@@ -132,4 +132,22 @@ describe("PolymorphicReference", () => {
...
     expect(isCommentParent(undefined)).toBe(false);
   });
+
+  it("can use base class parent", async () => {
+    const em = newEntityManager();
+    // Given an admin user (which extends user)
+    // User has a polymorphic reference to a favorite publisher
+    const adminUser = em.createPartial(AdminUser, { name: "a1", email: "test@test.com", role: "admin" });
+    // And a small publisher
+    const smallPublisher = em.createPartial(SmallPublisher, { name: "small", city: "city" });
```
stephenh, line 142: Suggestion: Can we use the factories here:

```
    // Given an admin user (which extends user)
    // User has a polymorphic reference to a favorite publisher
    const adminUser = newAdminUser(em);
    // And a small publisher
    const smallPublisher = newSmallPublisher(em);
```

packages/codegen/src/generateEntityCodegenFile.ts
0ces: You're right, this change isn't needed, just the `allFields` does it.

stephenh:
Looks great, thanks @0ces !

PR #931, opened 2024-01-05
fix: IncrementalCollectionOp on soft-deletable

0ces:


packages/orm/src/index.ts
```diff
@@ -140,6 +140,8 @@ export function setOpts<T extends Entity>(
           const allowDelete = !field.otherMetadata().fields["delete"];
           const allowRemove = !field.otherMetadata().fields["remove"];
 
+          const isSoftDeletable = !!field.otherMetadata().allFields["deletedAt"];
```
stephenh, line 143: We support configurable deleted at column names, so maybe use:

```
  const { entity, createdAt, updatedAt, deletedAt } = meta;
```

To get the field name. Might have to do like `getBaseMeta(field.otherMetadata()).deletedAt`

If that's defined, it will tell you: a) that soft delete is enabled on this entity, and b) what the field name is to use later in the `.set` call.

stephenh (approved):
Looks good as a stop-gap until we get softOrHardDelete behavior built into em.delete.

PR #945, opened 2024-01-16
fix: Fix m2m over reactivity

stephenh:
If we had:

* Books <- m2m -> Tags
* Author.tagsOfAllBooks watching `books: "tags"`
* And `a1 / b1.tags = [t1]` was an existing b1/tag
* When we added `b2.tags.add(1)` this correctly recalc'd `b2 -> a2`, `a2.tagsOfAllBooks`
* But it also walked from `t1.books.authors` and found `a1` and `a2`

This was unnecessary, because b1.tags hadn't changed.

### DIFF
--- a/packages/orm/src/ReactionsManager.ts
+++ b/packages/orm/src/ReactionsManager.ts
@@ -54,6 +54,8 @@ export class ReactionsManager {
         //   dirty, otherwise it will be left out of any INSERTs/UPDATEs.
         this.getPending(rf).todo.add(entity);
         this.getDirtyFields(getMetadata(rf.cstr)).add(rf.name);
+        // Keep for future debugging...
+        // console.log(`${entity}.${fieldName} changed, queuing ${rf.name} walk from ${entity} -> ${rf.path}`);
       }
     }
   }
@@ -134,10 +136,13 @@ export class ReactionsManager {
           }
           for (const doing of todo) pending.done.add(doing);
           // Walk back from the source to any downstream fields
-          return (await followReverseHint(todo, rf.path))
+          const relations = (await followReverseHint(todo, rf.path))
             .filter((entity) => !entity.isDeletedEntity)
             .filter((e) => e instanceof rf.cstr)
             .map((entity) => (entity as any)[rf.name]);
+          // Keep for future debugging...
+          // console.log(`Recalc-ing ${todo} -> ${rf.path} -> ${rf.cstr.name}.${rf.name} for found ${relations.length}`);
+          return relations;
         }),
       );
       // Multiple reactions could have pointed back to the same reactive field, so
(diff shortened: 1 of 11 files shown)


packages/orm/src/reactiveHints.ts
```diff
@@ -120,25 +120,40 @@ export function reverseReactiveHint<T extends Entity>(
             },
           );
         }
-        case "m2m":
+        case "m2m": {
```
stephenh, line 123: Reversals of m2m is different enough now (it already had some special casing on the old line 134) that I broke it out into its own case

packages/orm/src/relations/ManyToManyCollection.ts
```diff
@@ -228,23 +228,28 @@ export class ManyToManyCollection<T extends Entity, U extends Entity>
...
       m2m.remove(this.entity);
     });
-    this.loaded = [];
+    this.#loaded = [];
   }
 
   private maybeApplyAddedAndRemovedBeforeLoaded(): void {
-    if (this.loaded) {
-      // this.loaded.unshift(...this.addedBeforeLoaded);
-      // this.addedBeforeLoaded = [];
```
stephenh, line 237: I'd noticed this before, but was not sure why these lines were commented out.

They caused my initial repro attempts to fail, but due to this bug, instead of the reactivity bug--so went ahead and fixed this while here.

packages/orm/src/relations/ManyToManyCollection.ts
```diff
@@ -228,23 +228,28 @@ export class ManyToManyCollection<T extends Entity, U extends Entity>
...
-      this.removedBeforeLoaded?.forEach((other) => {
-        remove(this.loaded!, other);
-        getEmInternalApi(this.entity.em).joinRows(this).addRemove(this, this.entity, other);
+    if (this.#loaded) {
+      this.#addedBeforeLoaded?.forEach((e) => {
+        if (!this.#loaded?.includes(e)) {
```
stephenh, line 237: Fwiw I don't love the "for each added, scan loaded.includes", it would technically be an N^2 on large collections...

packages/tests/integration/src/relations/ManyToManyCollection.test.ts
```diff
@@ -231,8 +242,7 @@ describe("ManyToManyCollection", () => {
 
     book.tags.add(tag);
     expect(tag.books.get).toContain(book);
-    expect((await book.tags.load()).length).toBe(1);
```
stephenh, line 234: For some reason this test was asserting against the prior buggy behavior; after the add, `2` is the right length here. :shrug:

packages/orm/src/ReactionsManager.ts
```diff
@@ -54,6 +54,8 @@ export class ReactionsManager {
         //   dirty, otherwise it will be left out of any INSERTs/UPDATEs.
         this.getPending(rf).todo.add(entity);
         this.getDirtyFields(getMetadata(rf.cstr)).add(rf.name);
+        // Keep for future debugging...
+        // console.log(`${entity}.${fieldName} changed, queuing ${rf.name} walk from ${entity} -> ${rf.path}`);
```
stephenh, line 58: We should eventually have a way for users to turn this on at runtime...

PR #957, opened 2024-02-12
fix: nin any -> all operator in `buildKnexQuery`

brudil:
Fixes the `nin` condition within `buildKnexQuery` to use `all()` rather than `any()` which for `nin` entirely permissive.

This is inline with findDataLoader's `makeOp`: https://github.com/stephenh/joist-ts/blob/8960c68ae54cbe0f9d5907ad4aac5520d8383fa5/packages/orm/src/dataloaders/findDataLoader.ts#L271


stephenh:
Awesome, thank you!

PR #977, opened 2024-03-21
fix: Fix not loading o2os in createOrUpdatePartial.

stephenh:
Calling `setOpts(...)` with a o2o requires the o2o to be loaded, much like setting o2m collections.

However we weren't checking for o2os in the `createOrUpdatePartial` code that partitions out opts from "primitives / relations".

### DIFF
--- a/packages/orm/src/QueryParser.ts
+++ b/packages/orm/src/QueryParser.ts
@@ -711,7 +711,10 @@ export type ParsedValueFilter<V> =
   | { kind: "between"; value: [V, V] };
 
 /**
- * Parses the many/hodgepdoge (ergonomic!) patterns of value filters into a `ParsedValueFilter`.
+ * Parses the many/hodgepdoge (ergonomic!) patterns of value filters into a `ParsedValueFilter[]`.
+ *
+ * Note we return an array because filter might be a `ValueGraphQLFilter` that is allowed to have
+ * multiple conditions, i.e. `{ lt: 10, gt: 5 }`.
  */
 export function parseValueFilter<V>(filter: ValueFilter<V, any>): ParsedValueFilter<V>[] {
   if (filter === null) {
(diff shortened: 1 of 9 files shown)


packages/orm/src/QueryParser.ts
```diff
@@ -711,7 +711,10 @@ export type ParsedValueFilter<V> =
   | { kind: "between"; value: [V, V] };
 
 /**
- * Parses the many/hodgepdoge (ergonomic!) patterns of value filters into a `ParsedValueFilter`.
+ * Parses the many/hodgepdoge (ergonomic!) patterns of value filters into a `ParsedValueFilter[]`.
+ *
+ * Note we return an array because filter might be a `ValueGraphQLFilter` that is allowed to have
+ * multiple conditions, i.e. `{ lt: 10, gt: 5 }`.
```
stephenh, line 717: Unrelated comment update

packages/orm/src/createOrUpdatePartial.ts
```diff
@@ -122,37 +122,41 @@ export async function createOrUpdatePartial<T extends Entity>(
...
-      collectionsToLoad.push(field.fieldName);
+      relationsToLoad.push(field.fieldName);
 
-      const entities = !value
-        ? []
-        : (value as Array<any>).map(async (value) => {
-            if (!value || isEntity(value)) {
+      const entities = await Promise.all(
+        toArray(value).map(async (value: any) => {
+          if (!value || isEntity(value)) {
```
stephenh, line 129: This diff looks large, but really just removed the `!value ? [] : ...cast...` with `toArray` and everything else here is indentation change.

packages/orm/src/createOrUpdatePartial.ts
```diff
@@ -122,37 +122,41 @@ export async function createOrUpdatePartial<T extends Entity>(
...
+            if (removeMarker === true) entity.remove = true;
+            if (opMarker) entity.op = opMarker;
+            return entity;
+          }
+        }),
+      );
+      return [
+        name,
+        // Unwrap & keep null for o2o fields
+        field.kind === "o2o" ? (value === null ? null : entities[0]) : entities,
```
stephenh, line 158: This is new, to unwrap o2os out of the collection, since I wanted to mostly keep the ^ generic handling.

...that said, all of the ^ generic handling is around delete/remove/op markers, which o2os don't support/don't need, so maybe I should have just had totally separate `else` block anyway...

PR #993, opened 2024-03-26
fix: Fix em.load getting wrong STI type.

stephenh:


packages/orm/src/dataloaders/loadDataLoader.ts
```diff
@@ -27,7 +27,7 @@ export function loadDataLoader<T extends Entity>(
         conditions: [{ kind: "column", alias, column: "id", dbType: meta.idDbType, cond: { kind: "in", value: keys } }],
       },
       orderBys: [{ alias, column: "id", order: "ASC" }],
-    };
+    } satisfies ParsedFindQuery;
     addTablePerClassJoinsAndClassTag(query, meta, alias, true);
```
stephenh, line 31: This `addTablePerCTI` was not checking "wait, this is a an STI subtype", so added a `__class` column to the select query.

This `__class` used by the CTI infra to know "did this come from table 1 / table 2 / table 3", but on the STI side, we just use "what was the type"?

Which is great, but we accidentally added the `__class` here for an STI load, and so the CTI load logic kicked in, and fell into the "else load the default subtype" clause instead of looking at the `type_id` value.

This PR fixes `addTablePerCTI...` to make sure the type is using CTI inheritance.

packages/tests/integration/joist-config.json
```diff
@@ -70,7 +70,6 @@
         "parentOldTask": { "stiType": "TaskOld" },
         "publishers": { "stiType": "TaskOld" },
         "specialNewAuthor": { "stiType": "TaskNew" },
-        "parentOldTask": { "stiType": "TaskOld" },
```
stephenh, line 73: Turns out this was in here twice, this line is a duplicate.

PR #994, opened 2024-03-27
fix: Fix subtypes not running reactive rules.

stephenh:


packages/orm/src/EntityManager.ts
```diff
@@ -1639,14 +1640,16 @@ async function validateReactiveRules(
   const p1 = Object.values(todos).flatMap((todo) => {
     const entities = [...todo.inserts, ...todo.updates, ...todo.deletes];
     // Find each statically-declared reactive rule for the given entity type
-    const rules = getBaseAndSelfMetas(todo.metadata).flatMap((m) => m.config.__data.reactiveRules);
```
stephenh, line 1642: This `todo.metadata` was always the base type, i.e. we would group Task+TaskOld+TaskNew all into the same `todo`, which is good to issue just 1 SQL call.

But then we were only checking the "base and self" meta's list of reactive rules to run.

Now that each reactive rule knows the subtype its for, we could probably just push all of the rules onto the base meta, and then filter out ones for subtypes...

An optimization for another PR someday

packages/tests/integration/src/EntityManager.reactiveRules.test.tsx
```diff
@@ -210,60 +210,65 @@ describe("EntityManager.reactiveRules", () => {
   });
 
   it.withCtx("creates the right reactive rules", async () => {
-    const cstr = expect.any(Function);
     const fn = expect.any(Function);
-    expect(getMetadata(Author).config.__data.reactiveRules).toEqual([
+    expect(getReactiveRules(Author)).toMatchObject([
       // Author's firstName/book.title validation rule
-      { cstr, name: sm(/Author.ts:\d+/), fields: ["firstName"], path: [], fn },
+      { cstr: "Author", name: sm(/Author.ts:\d+/), fields: ["firstName"], path: [], fn },
```
stephenh, line 216: This is just updating all of the assertions to use this `getReactiveRules` helper methods that maps the constructors over to `cstr.name` because otherwise the jest output, as usual, is terribly huge/large/dumb.

PR #1002, opened 2024-03-30
fix: Preserve changed/new state for beforeCommit w/RQFs.

stephenh:
With the introduction of ReactiveQueryFields, the `em.flush` method basically needs it's "own dirty tracking", such that after we've done the 1st "micro-flush" (new term) to INSERT/UPDATE the entities, and then detect an RQF needs recalced we need to:

* Reset our "internal dirty state" on all of the just-flushed changes, but

* Preseve the "external dirty state" until the very end of em.flush, i.e. after `beforeCommit` hooks run.

Such that, after calling the RQF.recalc we can tell:

* What, if anything, has actually changed in this 2nd micro-flush loop, vs. the 1st micro-flush loop (i.e. in our internal dirty state), but

* External observation, i.e. hooks like `beforeCommit`, that are not called until *all micro-flushes have finished*, can tell what the flush-wide changed/state was (i.e. the external dirty state).

And ideally we'd keep track of both without a lot of duplicative bookkeeping when RQFs just aren't in use.

We achieve this by adding an `#orm.flushedData` hash, that is nearly always `undefined`/ignored, but when we're specifically in an RQF series of micro-flushes, we enable it to record only the "diffs from the last micro-flush".

I also added some wider cleanup to `EntityOrmField`, i.e. making it's `isNew` and `deleted` flags private, and updating code appropriately.

### DIFF
--- a/packages/orm/src/Entity.ts
+++ b/packages/orm/src/Entity.ts
@@ -112,11 +112,17 @@ export class EntityOrmField {
   }
 
   /** This is our internal "is pending flush", which is aware of RQF micro-flush loops. */
-  get pendingOperation(): "insert" | "update" | "delete" | "none" {
+  get pendingOperation(): "insert" | "update" | "delete" | "none" | "created-then-deleted" {
     if (this.#deleted) {
-      // Skip entities that were created and them immediately `em.delete`-d; granted this is
+      // Mark entities that were created and them immediately `em.delete`-d; granted this is
       // probably rare, but we shouldn't run hooks or have the driver try and delete these.
-      return this.#deleted === "pending" && !this.#new ? "delete" : "none";
+      // Note that we return them, instead of skipping entirely, so that fixupCreatedThenDeleted
+      // can later be called.
+      return this.#deleted === "pending" && this.#new
+        ? "created-then-deleted"
+        : this.#deleted === "pending"
+          ? "delete"
+          : "none";
     } else if (this.#new === "pending") {
       // Per ^, if `#new === flushed`, we fall through to check if micro-flush UPDATEs are required
       return "insert";
@@ -146,6 +152,7 @@ export class EntityOrmField {
   }
 
   fixupCreatedThenDeleted(): void {
+    // Ideally we could do this via `resetAfterFlushed`?
     this.#deleted = "deleted";
   }
 
--- a/packages/orm/src/Entity.ts
+++ b/packages/orm/src/Entity.ts
@@ -20,8 +20,6 @@ export interface Entity {
   readonly isNewEntity: boolean;
   readonly isDeletedEntity: boolean;
   readonly isDirtyEntity: boolean;
-  readonly isPendingFlush: boolean;
-  readonly isPendingDelete: boolean;
   set(opts: Partial<OptsOf<this>>): void;
   setPartial(values: PartialOrNull<OptsOf<this>>): void;
   /**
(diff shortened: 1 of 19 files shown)


packages/orm/src/Entity.ts
```diff
@@ -47,44 +45,113 @@ export class EntityOrmField {
...
+    this.wasNew ||= !!this.#new;
+    this.#new = undefined;
+    if (this.#deleted === "pending" || this.#deleted === "flushed") {
+      this.#deleted = "deleted";
+    }
+  }
+
+  /** Our public-facing `isNewEntity`. */
+  get isNewEntity(): boolean {
+    return !!this.#new;
```
zgavin, line 101: nit: kinda prefer the more explicit `!== undefined` over `!!`

packages/orm/src/Entity.ts
```diff
@@ -47,44 +45,113 @@ export class EntityOrmField {
...
+  }
+
+  /** Our public-facing `isDeletedEntity`. */
+  get isDeletedEntity(): boolean {
+    return !!this.#deleted;
```
zgavin, line 106: nit: ditto

packages/orm/src/Entity.ts
```diff
@@ -47,44 +45,113 @@ export class EntityOrmField {
...
+  }
+
+  /** Our public-facing `isDirtyEntity`. */
+  get isDirtyEntity(): boolean {
+    return Object.keys(this.originalData).length > 0;
+  }
+
+  /** This is our internal "is pending flush", which is aware of RQF micro-flush loops. */
+  get pendingOperation(): "insert" | "update" | "delete" | "none" {
```
zgavin, line 115: nit: maybe extract these values out into an enum?

packages/orm/src/Entity.ts
```diff
@@ -47,44 +45,113 @@ export class EntityOrmField {
...
+  }
+
+  /** Called when an `em.load` tries to find the entity and it's just gone from the db. */
+  markDeleteBecauseNotFound(): void {
+    this.#deleted = "deleted";
+  }
+
+  /** Called by `em.delete`, returns true if this is new information. */
+  markDeleted(): boolean {
+    if (!this.#deleted) {
```
zgavin, line 141: nit: again prefer the more explicit `=== undefined`

packages/orm/src/EntityManager.ts
```diff
@@ -1055,11 +1055,8 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW> {
    */
   delete(entity: Entity): void {
     // Early return if already deleted.
-    const orm = getOrmField(entity);
-    if (orm.deleted) {
-      return;
-    }
-    orm.deleted = "pending";
+    const changed = getOrmField(entity).markDeleted();
```
zgavin, line 1062: nit: maybe `alreadyMarkedDeleted` instead of `changed`?

packages/orm/src/Entity.ts
```diff
@@ -47,44 +45,113 @@ export class EntityOrmField {
...
+      return this.isTouched || Object.keys(this.originalData).length > 0 ? "update" : "none";
     }
   }
+
+  get changedFields(): string[] {
+    return this.flushedData ? Object.keys(this.flushedData) : Object.keys(this.originalData);
+  }
+
+  /** Called when an `em.load` tries to find the entity and it's just gone from the db. */
+  markDeleteBecauseNotFound(): void {
```
zgavin, line 135: nit: feel like the tense should match, ie `markDeleted`

zgavin (approved):

PR #1008, opened 2024-04-02
feat: Add ReactiveGetters.

### LINKED ISSUE #1007: Add hasReactiveGetter for getters that know their source
We'd like to have getters that can be used in reactive hints, but are always loaded b/c they use columns that are limited to the entity itself.

```ts
  get isBilled(): boolean {
    return this.isApproved || this.isPaid;
  }
```

Becomes:

```ts
   readonly isBilled: ReactiveGetter<Bill, boolean> = hasReactiveGetter("status", b => b.status === Approved || b.status === Paid);
```

stephenh:
These provide a way to have dumb/simple/always-get-able getters, but still denote/advertise the underlying primitive fields that are being used, so that the reactivity infra can setup dependency tracking appropriately.

Fixes #1007

### DIFF
--- a/packages/orm/src/reactiveHints.ts
+++ b/packages/orm/src/reactiveHints.ts
@@ -17,6 +17,7 @@ import {
   Reference,
 } from "./relations";
 import { LoadedOneToOneReference } from "./relations/OneToOneReference";
+import { ReactiveGetterImpl } from "./relations/ReactiveGetter";
 import { AsyncPropertyImpl } from "./relations/hasAsyncProperty";
 import { fail, mergeNormalizedHints } from "./utils";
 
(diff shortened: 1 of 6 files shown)


zgavin (approved):
Kinda surprised the typing just works and you didn't have to update `Reactable`?

stephenh:
@zgavin ah yeah, I just had `ShallowReactiveHint` use `FieldsOf` directly instead of the `Reactable`, so that got it to be "only immediate fields (primitives, enums, and m2os)" and not the other relations.

zgavin:
@stephenh yeah but im suprised that works in a `ReactiveHint` since whatever your `ReactiveGetter`  is named that field wouldn't conform to `Reactable<T extends Entity> = FieldsOf<T> & Loadable<T> & SuffixedFieldsOf<T> & SuffixedLoadable<T>;`

PR #1025, opened 2024-04-09
feat: esm codegen

brudil:
This is a bit of a WIP.

So far taken the approach of a `esm: true` flag within joist-config.json.

I've added a `joist-tests-esm-misc` albiet without tests, but with the option enabled + ts-config.json moduleResoution "Node16" - this should act as a decent way of ensuring regressions in this area - 'testing' via "tsc --noEmit"

~~@stephenh Hitting a couple of decisions, just enabling the ts-poet option in saveFiles has no effect.~~

I think this is for a couple reasons:
- `export from` statements, which the `imp` (I don't think supports) but we extentions on these all the same for this mode to work
- A fair bit of joist-codegen does not use `imp` but hardcodes as strings

Take for example `generateEntityFile`.

```ts
export function generateEntityFile(meta: EntityDbMetadata): Code {
  const entityName = meta.entity.name;
  const codegenClass = imp(`${entityName}Codegen@./entities`);

  return code`
    import { ${meta.entity.configConst.symbol} as config } from "./entities";

    export class ${entityName} extends ${codegenClass} {}

    // remove once you have actual rules/hooks
    config.placeholder();
  `;
}
```

I'm happy to push this through using the current strings with some lil additional checks template tags, and then perhaps if everything should/could be using imp (and perhaps a new export thing in ts-poet) we can go that in a seperate PR?

- [x] `imp` from `ts-poet` usages outputting correctly
- [x] currently hardcoded imports & exports managed via string manipulation
- [x] codegen.XCodegen.ts
- [x] index.ts
- [x] Entity file
- [x] factories/*.ts
- [x] entities.ts


brudil:
@stephenh actually there's places I think impl should be working and it's not doing anything, from what I can see in ts-poet, saveFiles options should be merged deeply merged with any options in CodegenFile, but something isn't being pulled through somewhere.

brudil:
I'm a bit closer now - didn't realise the ts-poet func needed an extention and then it can act on it (stripping or transforming) so I've updated these, are we're getting somewhere.

Now only issues are with the factories and the entities themselves - these are not having their .ts coverted when the option is enabled, which is odd because as far as I can tell, it's being saved by the same `saveFiles` function and they are actually using the same `Import` `Entity.type`

brudil:
As mad as it sounds, it's based on if the overwrite flag is enabled - will look in to if this is some ts-poet bug.

brudil:
Sweet, ts-poet fix/update has done it.

@stephenh this is ready for review.

- This does not change any "stringly done" importing/exporting but just adds in `.js` to the end where needed.
- For the `imp` usage, I've added `.ts` to these, to active the ts-poet option
- `esm: true` in joist-config.json will enable it, otherwise we pass in `false` so ts-poet strips these out again

For testing/regressions

- I've created a `joist-tests-esm-misc` package
  - This sets "type": "module" in package.json, and "moduleResolution": "Node16" in tsconfig.json, which causes tsc errors if imports are not .js
  - We can use this as regression test suite, by using "tsc --noEmit" in the "test" script instead of jest
  - I'm in two minds about if we should not gitignore src/*entities in this directory
    - This will require codegen to be reun before tests (is this already a requirement/part of the testing pipeline?)
    - But it will ensure we're testing against all files the codegen produces and not just those that have overwrite enabled.

packages/codegen/src/EntityDbMetadata.ts
```diff
@@ -455,9 +455,9 @@ function newEnumField(config: Config, entity: Entity, r: M2ORelation, enums: Enu
   const columnType = (column.type.shortName || column.type.name) as DatabaseColumnType;
   const fieldName = enumFieldName(column.name);
   const enumName = tableToEntityName(config, r.targetTable);
-  const enumType = imp(`${enumName}@./entities`);
-  const enumDetailType = imp(`${plural(enumName)}@./entities`);
-  const enumDetailsType = imp(`${enumName}Details@./entities`);
+  const enumType = imp(`${enumName}@./entities.ts`);
```
stephenh, line 458: I was wondering "why are all of the existing test projects not having `.ts` added to their output?" and realized/forgot that ts-poet has the `importExtensions=false` to strip extensions--that was handy! :-)

stephenh (approved):
This is amazing! Really awesome that we can have a type=module test project in the repo as well. Thank you @brudil !

...wow, also realizing this means I can probably try getting our app running vitest! :tada: 

(...we use `toMatchEntity` ton, so I think making that work in non-jest environments is probably step 1...)

brudil:
Just to note, we've hit some issues with this with paths like `Notifications.types.ts` but PR up in ts-poet here for a fix https://github.com/stephenh/ts-poet/pull/57 due to broken regex.

PR #1029, opened 2024-04-10
feat: ensure codegen output meets `verbatimModuleSyntax` expectations

brudil:
For those using esm with non-tsc transpiler, such as SWC ensuring that types are specifically imported by `import type` is important as there no inference of if the import should be in the final output or not.

TypeScript 5.0 introduced a new option to be stricter against this sort of thing: `verbatimModuleSyntax` when enabled, tsc will error at any type being imported as a value.

ts-poet has great functionality for this using the `t:*` prefix on imports. This PR essentially just updates the symbols.ts file to utilise this on every symbol import of a type.

This PR has enabled verbatimModuleSyntax within the new esm testing dir from #1025 to validate this and prevent future regressions and we're passing now!

This will create some noise to codegen files even with `esm` not enabled in joist-config.json as this is a import-level change, but I've left that in the last commit outside of the implement commits prior.


stephenh:
Ah nice!

We can probably go ahead and merge this; my only thought is that ts-poet doesn't currently know about the "put `type` inside the brackets" capability (which either didn't exist when I added the `t:` syntax to ts-poet, or more likely I just didn't know about it :-)), like:

```
import { type Foo, Bar } from "../"
```

Which seems like cleaner output, compared to the all-`type`s dedicated import line.

But it'll be easy enough to follow this up with a ts-poet bump should it ever get smarter about that.

@brudil lmk if you'd like to try and fix/change that `t:` / `type Foo` output in ts-poet, otherwise I'll plan on poking at it sometime over the next week or so-ish.

brudil:
fwiw; I don't think the schema setup in the esm dir are enough to flush everything out - I've missed some running this in our app. I'll follow this up shortly. Apologies.

stephenh:
> I don't think the schema setup in the esm dir are enough to flush everything out

Yep, definitely not surprised; there is a reason the main `tests/integration` schema is so convoluted :-) 

But also happy to have that `esm-misc` schema be only as complicated as necessary to flush out the ESM boundary conditions.

Issue #1031, opened 2024-04-10
esm, circular imports, await


brudil:
During our transition to esm, I hit a curious issue

<img width="714" alt="image" src="https://github.com/joist-orm/joist-orm/assets/382352/3b992ee7-94b2-44f7-a063-e8a52619529b">

I got this after moving a different package of ours to esm too. This module is imported within the entity (sync function that's used in a getter).

Thinking there had been more rabbit holes than an bunny after acupuncture, I hastily moved the package back to cjs and continued with the rest of the work. It has been in the back of my mind though - what could have been causing this?

To cut a boring story short - that package was importing Facebook's @lexical/headless. Our package moving to esm, obviously then starts utilising their esm builds. And that's when I saw something interesting.

The imported file:
```js
/**
 * Copyright (c) Meta Platforms, Inc. and affiliates.
 *
 * This source code is licensed under the MIT license found in the
 * LICENSE file in the root directory of this source tree.
 */
const mod = await (process.env.NODE_ENV === 'development' ?
  import('./LexicalHeadless.dev.mjs')
  : import('./LexicalHeadless.prod.mjs'));
export const createHeadlessEditor = mod.createHeadlessEditor;
```

That's a top-level await in the wild! This is when things started clicking around the nature of the error.

Enough that it's reproducable with

```ts
import {
  ContentEditorCodegen,
  contentEditorConfig as config,
} from './codegen/ContentEditorCodegen.js';

// Behold, Await, drama!
await new Promise<void>((r) => { r() });

export class ContentEditor extends ContentEditorCodegen {}

config.placeholder();
```

Now I know we're probably the first users of Joist in a ESM context, so I'm more than happy to continue hacking on this and playing around with this. We can workaround this issue in otherway in the meantime, so it's by no means a priority.

I think this will become a issue that will eventually need solving as more applications - and packages - start moving over.

Something interesting though - I've resolved this issue by changing metadata.ts to manually import ContentEditor (entity file with the promise)

```ts
// removed from big import from entities.ts

import { ContentEditor } from '../ContentEditor.js';
import { contentEditorConfig } from '../codegen/ContentEditorCodegen.js';
```

----

Questions for @stephenh if you've got time, I know there's the entities.ts file which explains it's purpose in fixing issues around circular importing. Any issues or gotchas should I look out for which have arrisen before re-circulars?

I'll see how this goes might it might be as simple as importing things slightly more directly within metadata.ts in esm mode.

[...]

PR #1032, opened 2024-04-10
fix: directly import entities from metadata.ts, support top level await

brudil:
Bit of trial and error here but looks as if I've figured out a fix for #1031 but lmk if I'm missing something big here!

1. This directly imports entities (their codefile directly rather than via entities.ts) within metadata.ts
2. This does not touch imports to entities elsewhere (such as in *Codegen.ts file, which does trigger circ issues)
3. This is enabled by default, but happy to update this PR to gate this under the esm option
4. Currently not tested as top level await is esm only so perhaps need to expand the esm dir in the future
  - Applied this patch in our app and we're running once again, tests passing

There's codegen consequences of this but I haven't committed them until feedback and thoughts on 3.

### DIFF
--- a/packages/codegen/src/EntityDbMetadata.ts
+++ b/packages/codegen/src/EntityDbMetadata.ts
@@ -45,6 +45,8 @@ export type Entity = {
   name: string;
   /** The symbol pointing to the entity itself. */
   type: Import;
+  /** The symbol pointing to the entity itself, directly rather than via entities.ts */
+  typeForMetadataFile: Import;
   /** The name of the entity's runtime metadata const. */
   metaName: string;
   /** The symbol pointing to the entity's runtime metadata const. */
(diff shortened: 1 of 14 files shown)


stephenh:
@brudil  I'm curious to understand root cause more, but specifically per `3` yep happy to have this change just apply to everyone, as I assume it won't break CJS setups, so might as well be the same between the two.

brudil:
@stephenh Pushed the codegen. Will research further on the black magic that this is, but if the tests pass, happy for this to go in whenever.

stephenh:
@brudil cool, will merge! I'm a little nervous that, given how annoying this circular issue is in general / even before ESM, that we might hit some boundary case in our app when I roll this out to it, but :crossed_fingers: we'll see! Hopefully it just works.

brudil:
I'll be curious to hear about how it works against your app!

stephenh:
@brudil just bumped our app and all is good! :sweat_smile:

PR #1042, opened 2024-04-16
feat: Trigger hooks on m2m relation changes

Grandlouis:


stephenh:
@Grandlouis splitting up changes into two "fields" and "relations" keys is too much of a breaking change.

We should either figure out how to get both fields and relations into changes, or for now introduce a 2nd temporary field like entity.changedRelations, that we can combine into entry.changes later

Grandlouis:
> @Grandlouis splitting up changes into two "fields" and "relations" keys is too much of a breaking change.
> 
> We should either figure out how to get both fields and relations into changes, or for now introduce a 2nd temporary field like entity.changedRelations, that we can combine into entry.changes later

Uhh i didn't saw your comment until now, just got it working, let me see how to include it in `fields` instead then 🤔

Grandlouis:
@stephenh Can you take a new at it please ? 🙏

packages/tests/integration/src/relations/ManyToManyCollection.test.ts
```diff
@@ -565,4 +565,17 @@ describe("ManyToManyCollection", () => {
     expect((user as any).comments).not.toBeDefined();
     expect(user.likedComments).toBeDefined();
   });
+
+  it("can identify changes on the many-to-many", async () => {
```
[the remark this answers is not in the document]
Grandlouis: Aha! good idea, i'll try the other end of the relation as well.

Here is the config of the Author -> https://github.com/joist-orm/joist-orm/pull/1042/files#diff-8a5247c2e2dcce30fabb9d4870c1fc4b1d15ecab0664765ccf8e305cf06286e2R266

packages/orm/src/changes.ts
```diff
@@ -31,7 +38,9 @@ export interface ManyToOneFieldStatus<T extends Entity> extends FieldStatus<IdOf
  *    i.e. `Publisher.changes` is typed as `Changes<Publisher, keyof Publisher | keyof SmallPub | keyof LargePub>`
  * @type R An optional list of restrictions, i.e for `Reacted` for to provide `changes` to its subset of fields.
  */
-export type Changes<T extends Entity, K = keyof FieldsOf<T>, R = K> = { fields: K[] } & {
+export type Changes<T extends Entity, K = keyof (FieldsOf<T> & RelationsOf<T>), R = K> = {
+  fields: NonNullable<K>[];
```
stephenh, line 42: Is the `NonNullable` required here? When would `K` be null/undefined?

Grandlouis: its because the `RelationsOf<T>` relies on the `OptsOf<T>` which can be `undefined`, so adding the non-nullable as a safety way to tell other casting types this fields won't include undefined

packages/orm/src/changes.ts
```diff
@@ -58,14 +67,30 @@ export function newChangesProxy<T extends Entity>(entity: T): Changes<T> {
...
         ) as (keyof OptsOf<T>)[];
+
+        // scan the join rows to get if the relation has any rows
+        const m2mFieldsChanged: (keyof RelationsOf<T>)[] = [];
+        const emApi = getEmInternalApi(entity.em);
+        // we will report changes only on the many to many relations for now
+        const m2mFields = Object.values(getMetadata(entity).allFields).filter((f) => f.kind === "m2m");
+        for (const field of m2mFields) {
+          const m2m = entity[field.fieldName as keyof T] as ManyToManyCollection<any, any>;
+          const joinRow: any = emApi.joinRows(m2m);
```
stephenh, line 86: The `: any` is not a great hack; also I think just checking `joinRow.rows.length > 0` is going to over-report changes, like it's legal for `joinRow.rows` to be populated, and not actually have any changes queued up.

Can you remove this `any` and isntead do something like:

```
if (emAPi.joinRows(m2m).hasChanges) {
  . ..
}
```

Where `hasChanges` is a method/property in `JoinRows` that does a copy/paste/reuse of `toTodo` which specifically scans for added/deletes m2m rows.

packages/orm/src/config.ts
```diff
@@ -97,6 +97,11 @@ export class ConfigApi<T extends Entity, C> {
     this.__data.cascadeDeleteFields.push(relation);
   }
 
+  touchOnChange(relation: keyof RelationsIn<T> & LoadHint<T>): void {
```
stephenh, line 100: Do we need the `& LoadHint<T>`?

packages/orm/src/changes.ts
stephenh: Weird, I really can't explain why that's necessary, but :shrug: I guess we can ignore it for now.

packages/tests/integration/src/entities/Book.ts
```diff
@@ -78,4 +78,13 @@ config.addRule("tags", (b) => {
   return b.tags.get.length === 3 ? "Cannot have exactly three tags" : undefined;
 });
 
+// Example of a trigger for a many to many field
+config.touchOnChange("tags");
+config.beforeFlush(async (book) => {
```
stephenh, line 83: Nit can drop the `async`

packages/tests/integration/src/relations/ManyToManyCollection.test.ts
```diff
@@ -565,4 +565,19 @@ describe("ManyToManyCollection", () => {
...
+    const em = newEntityManager();
+    const book = newBook(em, { title: "To be changed by hook" });
+    const t1 = newTag(em, { name: "t1" });
+    await em.flush();
```
stephenh, line 573: Things will sometimes only work because entities are already loaded in memory; instead of using factories to create the book and tag, and can you use the `insert...` methods, and then have the test just do an em.load for both entities (but no populates of the `book.tags` or `tag.books` collections) and watch the same logic work?

Also we should have a test that for m2m removals. I.e. use the same `insert...` methods to start out with an m2m row between a book & tag, and then remove it.

packages/orm/src/config.ts
Grandlouis: can be removed 👍

packages/orm/src/changes.ts
Grandlouis: Adding 👍

packages/tests/integration/src/relations/ManyToManyCollection.test.ts
Grandlouis: I got it, i just added couple of unit test to cover those scenarios, i was missing the removal trigger xd but is working now

packages/orm/src/changes.ts
Grandlouis: typescript gymnastics? 😕 i really don't know why it cast undefined as well with the other keys

stephenh (approved):

PR #1045, opened 2024-04-19
feat: Add config.afterMetadata hook.

stephenh:


zgavin (approved):
tests are for people who write code that doesn't work the first time

Issue #1044, opened 2024-04-18
touchOnRelationChange follow ups


stephenh:
* Add a `state: pending | flushed` to `JoinRow` so that it can wait and be reset until after `afterComplete`
* Make it work for `o2m` and `o2o`
* Add a `ChangableFields` type that combines the OptsOf/RelationsIn

PR #1046, opened 2024-04-19
feat: m2m change detection follow up

Grandlouis:
Trying to satisfy this ticket -> https://github.com/joist-orm/joist-orm/issues/1044

~Update: I was only able to delay the "reset"/"update" of the m2m rows until `afterCommit` so all hooks get access to the m2m relation changes on the `changes.fields` proxy~

Update 2: Now the `JoinRow` has an `op` fields that indicates the state it has on the flush cycle, whether it is pending, flushed or completed. We use this flags to indicate if there are changes on the m2m relations.

Support for o2m, o2o and old value still pending 😅


packages/orm/src/EntityManager.ts
```diff
@@ -1200,6 +1200,7 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW> {
 
     try {
       const allFlushedEntities: Set<Entity> = new Set();
+      const allFlushedJoinRowsResetFns: Set<() => void> = new Set();
```
stephenh, line 1203: I think we can avoid this `new Set` and use:

```
      let joinRowTodos = combineJoinRows(this.#joinRows);
```

To loop over, later in the file.

packages/orm/src/EntityManager.ts
```diff
@@ -1260,6 +1262,8 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW> {
           if (e.isNewEntity && !e.isDeletedEntity) this.#entityIndex.set(e.idTagged, e);
           getInstanceData(e).resetAfterFlushed();
         }
+        // Update the joinRows refs to reflect the new state
+        allFlushedJoinRowsResetFns.forEach((resetFn) => resetFn());
```
stephenh, line 1266: I'd prefer something like:

```
for (joinRow in joinRowTodos) {
  joinRow.resetAfterFlushed();
}
```

Where `JoinRow.resetAfterFlushed` changed did something like `this.op: "inserted"` or `this.op = "deleted"`, similar to how `InstanceData.resetAfterFlushed` works.

And then ideally `JoinRow.hasChanges` changes from:

```
  get hasChanges() {
    const todos = this.toTodo();
    return !!todos;
  }
```

To more like `this.op = pending || flushed`

I.e. `JoinRow.op` is a literal of pending/inserted/deleted (or whatever ~2-3 values make sense to handle both "pending insert" + "pending delete" + "flushed").

packages/orm/src/drivers/Driver.ts
```diff
@@ -25,5 +25,5 @@ export interface Driver {
 
   flushEntities(em: EntityManager, todos: Record<string, Todo>): Promise<void>;
 
-  flushJoinTables(em: EntityManager, joinRows: Record<string, JoinRowTodo>): Promise<void>;
+  flushJoinTables(em: EntityManager, joinRows: Record<string, JoinRowTodo>): Promise<{ resetAfterFlushed?: () => void }>;
```
stephenh, line 28: I think with the `JoinRow.resetAfterFlushed` we should be able to leave this as-is

packages/orm/src/drivers/PostgresDriver.ts
```diff
@@ -152,7 +157,7 @@ export class PostgresDriver implements Driver {
         });
         const { rows } = await knex.raw(sql, bindings);
         for (let i = 0; i < rows.length; i++) {
-          newRows[i].id = rows[i].id;
+          resetAfterFlushedIds.push({ rows: newRows, index: i, id: rows[i].id });
```
stephenh, line 160: I'd like to keep this as-is, i.e. still immediately assign the `.id` and maybe do like `newRows[i].op = "flushed"`, to indicate it's no longer "op = pending", and instead has been moved to flushed-in-the-db.

packages/orm/src/drivers/PostgresDriver.ts
```diff
@@ -184,8 +189,14 @@ export class PostgresDriver implements Driver {
             await knex(joinTableName).del().whereIn([m2m.columnName, m2m.otherColumnName], data);
           }
         }
+
+        deletedRows.forEach((_r, index) => resetAfterFlushedIds.push({ rows: deletedRows, index, id: undefined }));
```
stephenh, line 193: Same ideally this does like `deletedRows[i].op = "flushed"`.

stephenh (commented):
Hey @Grandlouis ; thanks for the PR! I'd like to try using a `JoinRow.op` flag instead and see how that works; it will be similar to how `InstanceData.op` works, which I think will be nice for clarity, and also probably help us eventually implementing the "use changes to get both old value + new value".

stephenh (approved):
This is great @Grandlouis , thank you!

PR #1072, opened 2024-05-04
fix: Fix id alias columns typed as numbers.

stephenh:
Previously the `id` column for like:

```
const a = alias(Author);
em.find(..., where: { a.id... });
```

The `id` column was typed as `number`, like the underlying database, instead of string to match our tagged ids.

This was because of a mistake in the entity codegen generation of the `Fields` interface, just using the wrong `idType`.

### DIFF
--- a/packages/orm/src/Aliases.ts
+++ b/packages/orm/src/Aliases.ts
@@ -27,13 +27,15 @@ export function aliases<T extends readonly MaybeAbstractEntityConstructor<any>[]
 }
 
 export type Alias<T extends Entity> = {
-  [P in keyof FieldsOf<T>]: FieldsOf<T>[P] extends { kind: "primitive" | "enum"; type: infer V; nullable: infer N }
-    ? PrimitiveAlias<V, N extends undefined ? null : never>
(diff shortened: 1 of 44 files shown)


packages/orm/src/Aliases.ts
```diff
@@ -27,13 +27,15 @@ export function aliases<T extends readonly MaybeAbstractEntityConstructor<any>[]
...
 
 export type Alias<T extends Entity> = {
-  [P in keyof FieldsOf<T>]: FieldsOf<T>[P] extends { kind: "primitive" | "enum"; type: infer V; nullable: infer N }
-    ? PrimitiveAlias<V, N extends undefined ? null : never>
-    : FieldsOf<T>[P] extends { kind: "m2o"; type: infer U }
-      ? EntityAlias<U>
-      : FieldsOf<T>[P] extends { kind: "poly"; type: infer U extends Entity }
-        ? PolyReferenceAlias<U>
-        : never;
+  [P in keyof FieldsOf<T>]: P extends "id"
```
stephenh, line 30: Put a special case here for `"id`" to get it flipped from a `PrimiviteAlias<string>` to `EntityAlias<T>`, which seems more accurate.

PR #1080, opened 2024-05-09
fix: Add tests for previous reactive hint fix, with more fixes.

stephenh:
This adds a test to reproduce the prior "walk through a different subtype that does not have the relation" error, as well as applies the fix to m2o and m2m handling as well.

There is also a related fix that runtime filtering of the type filter, i.e. `publisher@Publisher` should pass for subtypes of `Publisher` like either a `LargePublisher` or `SmallPublisher`, and previously we were only exact-matching on the type name.

### DIFF
--- a/packages/orm/src/reactiveHints.ts
+++ b/packages/orm/src/reactiveHints.ts
@@ -1,6 +1,16 @@
 import { Entity } from "./Entity";
 import { FieldsOf, MaybeAbstractEntityConstructor, RelationsOf, getEmInternalApi } from "./EntityManager";
-import { EntityMetadata, getBaseAndSelfMetas, getMetadata } from "./EntityMetadata";
+import {
+  EntityMetadata,
+  ManyToManyField,
+  ManyToOneField,
+  OneToManyField,
+  OneToOneField,
+  PolymorphicFieldComponent,
+  getBaseAndSelfMetas,
+  getMetadata,
+  getSubMetas,
+} from "./EntityMetadata";
 import { Changes, FieldStatus, ManyToOneFieldStatus } from "./changes";
 import { isChangeableField } from "./fields";
 import { getProperties } from "./getProperties";
@@ -146,9 +156,10 @@ export function reverseReactiveHint<T extends Entity>(
           if (!isReadOnly) {
             fields.push(field.fieldName);
           }
+          const otherFieldName = maybeAddTypeFilterSuffix(meta, field);
           return reverseReactiveHint(rootType, field.otherMetadata().cstr, subHint, undefined, false).map(
             ({ entity, fields, path }) => {
-              return { entity, fields, path: [...path, field.otherFieldName] };
+              return { entity, fields, path: [...path, otherFieldName] };
             },
           );
         }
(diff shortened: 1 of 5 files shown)


packages/orm/src/reactiveHints.ts
```diff
@@ -161,21 +172,18 @@ export function reverseReactiveHint<T extends Entity>(
...
               },
             );
           });
         }
         case "m2m": {
-          const otherField =
-            field.otherMetadata().allFields[field.otherFieldName] ??
-            fail(`No field ${field.otherMetadata().type}.${field.otherFieldName}`);
-          const otherFieldName =
-            otherField.kind === "poly" ? `${field.otherFieldName}@${meta.type}` : field.otherFieldName;
```
stephenh, line 174: I extract this to `maybeAddTypeFilterSuffix` so it could be used across the relation `case`s

packages/tests/integration/src/reactiveHints.test.ts
```diff
@@ -1,116 +1,151 @@
...
+import { ReactiveTarget, convertToLoadHint } from "joist-orm/build/reactiveHints";
 
 const am = getMetadata(Author);
 
 describe("reactiveHints", () => {
   it("can do immediate primitive field names", () => {
-    expect(reverseReactiveHint(Author, Author, "firstName")).toEqual([
-      { entity: Author, fields: ["firstName"], path: [] },
-    ]);
+    expect(reverse(Author, Author, "firstName")).toEqual([{ entity: "Author", fields: ["firstName"], path: [] }]);
```
stephenh, line 18: I added a helper method to turn `entity` into a string name, to stop blowing up Jest diffs on the entity constructors.

packages/orm/src/reactiveHints.ts
```diff
@@ -264,11 +288,11 @@ export async function followReverseHint(entities: Entity[], reverseHint: string[
   // And "walk backwards" through the reverse hint
   while (paths.length) {
     const path = paths.shift()!;
-    const [fieldName, viaPolyType] = path.split("@");
+    const [fieldName, viaType] = path.split("@");
```
stephenh, line 291: Renamed to just `viaType` since it's used for both polys & CTE/inheritance now

packages/orm/src/reactiveHints.ts
```diff
@@ -374,14 +398,34 @@ export interface ReactiveTarget {
   path: string[];
 }
 
-async function maybeLoadedPoly(loadPromise: Promise<Entity>, viaPolyType: string | undefined) {
-  if (viaPolyType) {
-    const loaded: Entity = await loadPromise;
-    return loaded && getMetadata(loaded).type === viaPolyType ? loaded : undefined;
+function maybeApplyTypeFilter(loadPromise: Promise<Entity | Entity[]>, viaType: string | undefined) {
```
stephenh, line 401: Took out the async/await

packages/orm/src/reactiveHints.ts
```diff
@@ -374,14 +398,34 @@ export interface ReactiveTarget {
...
 }
 
+/** Handle `viaType` filtering with subtype awareness. */
+function isTypeOrSubType(entity: Entity, typeName: string): boolean {
+  const meta = getMetadata(entity);
+  // Easy check for the name is the same
+  if (meta.type === typeName) return true;
+  // Otherwise see if the entity is a subtype of the typeName, i.e. if our poly/type
+  // filter is `@Publisher`, and we're a `SmallPublisher`, that's valid to traverse.
+  for (const other of getSubMetas(meta)) {
```
stephenh, line 423: Now we allow subtypes of the filter to pass

PR #1100, opened 2024-05-26
feat: Support schemas with non-deferred FKs.

### LINKED ISSUE #1097: Support schemas that don't use deferred FKs
* Probably topo sort by required FKs
* Sort the order in `metadata.ts` somewhere
* If a schema is using deferred FKs, don't sort / don't change any behavior
* If cycles prevent insertion order, use dummy ids, and backfill in a 2nd clean-up flush
* This probably places into support of non-`SEQUENCE` based databases

stephenh:
Todo:

* Add a `nonDeferredForeignKeys: ignore | warn | error"
  * Warn is the default, creates a SQL file
  * Error just stops, current behavior
  * Ignore, will use fixups
* Detect not-null cycles, error out
  * Make them deferred (if postgres)
  * Make one nullable (any db, can fixup)

Fixes #1097


docs/docs/getting-started/tour.md
```diff
@@ -33,7 +33,7 @@ export class AuthorCodegen {
   // ...all the boilerplate fields & m2o/o2m/m2m relations generated for you...
   readonly books: Collection<Author, Book> = hasOne(...);
   get firstName(): string { ... }
-  set firstName(): string { ... }
+  set firstName(value: string): string { ... }
```
zgavin, line 36: 😱

packages/codegen/src/index.ts
```diff
@@ -39,10 +48,61 @@ async function main() {
...
+      console.log(" - Execute the generated alter-foreign-keys.sql file (one-time fix)");
+      console.log(" - Set 'nonDeferredFks: ignore' in joist-config.json");
+      console.log("");
+
+      console.log("See https://joist-orm.io/docs/getting-started/schema-assumptions#deferred-constraints");
+      console.log("");
+
+      await writeAlterTables(nonDeferredFks);
+
+      if (setting === "error") hasError ??= true;
```
zgavin, line 84: nit: just `hasError = true`?

packages/codegen/src/sortForeignKeys.ts
```diff
@@ -0,0 +1,155 @@
...
+  const nullableCycles: string[] = [];
+  const notNullCylces: string[] = [];
+
+  // If we see `books.author_id`, mark `author.inDegree++`
+  entities.forEach((entity) => {
+    entity.nonDeferredFks
+      .filter((m2o) => m2o.notNull)
+      .forEach((m2o) => {
+        const added = addDependency(graph, entity, m2o);
+        if (Array.isArray(added)) notNullCylces.push(added.join(" -> "));
```
zgavin, line 59: nit: spelling "Cylces"

zgavin (approved):
I think i've grokked this, good to open up support for more setups. 

Why would you not have deferred FKs though 🤔

stephenh:
> Why would you not have deferred FKs though

@zgavin totally agreed; my hope is that users will use the `alter-foreign-keys.sql` to convert over, my assumption is just that anyone not using Joist today a) probably doesn't even know deferred FKs are a thing / b) does not have them enabled for every FK in their system.

So just want to lower the bar/entry point, at least for PG setups.

PR #1104, opened 2024-06-04
fix(fields.ts): fail not defined

TylerR909:
Line 21 is throwing this error:

```sh
    ReferenceError: fail is not defined
```

Instead of the correct `Missing serde for myFieldName` error.

It looks like an import was added last week for `import { maybeRequireTemporal } from "./utils";` where there didn't used to be any imports from `./utils`, so maybe it used to be getting this for free or from the environment somehow?


stephenh (approved):
Ah yeah, jest "helpfully" adds it's `fail` to the global type scope, so this can sometimes fall through and not get caught. Thanks for the fix!

PR #1112, opened 2024-06-13
fix: Fix newTestInstance not using entities in arrays.

stephenh:


packages/tests/integration/src/EntityManager.factories.test.ts
```diff
@@ -121,6 +122,39 @@ describe("EntityManager.factories", () => {
...
+  it("will use existing entities within the opts literal", async () => {
+    const em = newEntityManager();
+    // Given two authors (to turn off the obvious default)
+    const [, a2] = [newAuthor(em), newAuthor(em)];
+    // Given we want to make a Book
+    const b = newTestInstance(
+      em,
+      Book,
+      // and we refer to one of the authors in the opts literal
+      { comments: [{ parent: a2 }] },
```
stephenh, line 134: This test already passed, as-is

packages/tests/integration/src/EntityManager.factories.test.ts
```diff
@@ -121,6 +122,39 @@ describe("EntityManager.factories", () => {
...
+  it("will use existing entities within the opts literal array", async () => {
+    const em = newEntityManager();
+    // Given two comments (to turn off the obvious default)
+    const [, c2] = [newComment(em), newComment(em)];
+    // Given we want to make a Book
+    const b = newTestInstance(
+      em,
+      Book,
+      // and we refer to one of the comments in the opts literal
+      { comments: [c2] },
```
stephenh, line 151: This one did not, because we spread + `Object.values(c2)` which skipped over putting `c2` itself into the use map.

PR #1115, opened 2024-06-15
feat: bytea support

brudil:
Lightweight support for `bytea`.

Knex returns Buffers which are a superset of Uint8Array. joist-codegen types the primitive as Uint8Array - but this should be configurable in userland with the `type` or/and CustomSerde options, if they wish to opt in to the Buffer typing. Alternatively an instanceof check can be used.

I haven't touched too much outside of `mapSimpleDbTypeToTypescriptType` for this. Followed the bigint tests within Author.

Haven't really touched filtering here - not even sure how sensible that would be to do anyway.


packages/tests/integration/migrations/1580658856631_author.ts
```diff
@@ -139,6 +139,8 @@ export function up(b: MigrationBuilder): void {
     tags_of_all_books: { type: "varchar", notNull: false },
     // for testing full-text-search fields that want to use `id`
     search: { type: "text", notNull: false },
+    // for testing full-text-search fields that want to use `id`
```
stephenh, line 142: Looks like a stale comment?

stephenh (approved):
Looks great!

packages/tests/integration/migrations/1580658856631_author.ts
brudil: Good spot! fixed in last commit.

PR #1117, opened 2024-06-19
feat: Only flush changed tables

stephenh:


stephenh:
@brudil FYI given you're using CTI, if you're using the out-of-the-box `flush_database` function, this will probably be a breaking change for you...

We found a 8% speedup on our test suite by not `DELETE`-ing "literally every single table in the schema", but using this trick to watch for which sequences were ticked.

It works pretty great, but given that CTI sub-tables don't get their own sequences, we changed `createSubTable` to make the CTI "subtable id that is a FK to the base id" to also `ON DELETE CASCADE`.

This lets hew new `flush_database` get away with only deleting the rows from the base table, and then deferring to pg to clear out the subtables.

Hopefully this is okay! I.e. besides `flush_database`, it kinda makes sense to have the `ON DELETE CASCADE` into subtables anyway, I think.

PR #1119, opened 2024-06-21
feat: Optionally allow unsaved entities to be returned from a resolver

chr1sjf0x:
For some circumstances, like a mutation which supports a `dryRun` option, it would be nice if resolvers could return unsaved entities, without the test helpers failing.

This PR introduces an optional flag to `run` methods to facilitate this.

### DIFF
--- a/packages/test-utils/src/run.ts
+++ b/packages/test-utils/src/run.ts
@@ -1,4 +1,4 @@
-import { Entity, EntityManager, isEntity } from "joist-orm";
+import { Entity, EntityManager, isDefined, isEntity } from "joist-orm";
 import { fail } from "joist-utils";
 import { Context } from "./context";
 
@@ -6,11 +6,16 @@ type MaybePromise<T> = T | Promise<T>;
 
 export type ContextFn<C> = (ctx: C) => MaybePromise<C>;
 
+type RunOptions = {
+  expectUnsavedEntities?: boolean;
+};
+
 /** Runs the `fn` in a dedicated / non-test Unit of Work . */
 export async function run<C extends Context, T>(
   ctx: C,
   fn: (ctx: C) => MaybePromise<T>,
   contextFn: ContextFn<C> = newContext,
+  opts?: RunOptions,
 ): Promise<T> {
   const { em } = ctx;
   // Ensure any test data we've setup is flushed
@@ -18,14 +23,14 @@ export async function run<C extends Context, T>(
   const result = await fn(await contextFn(ctx));
   // We expect `fn` (i.e. a resolver) to do its own UoW management, so don't flush.
   await em.refresh({ deepLoad: true });
-  return mapResultToOriginalEm(em, result);
+  return mapResultToOriginalEm(em, result, opts);
 }
 
 /** Creates a `run` with a custom `newContext` method. */
 export function makeRun<C extends Context>(
   contextFn: ContextFn<C>,
-): <T>(ctx: C, fn: (ctx: C) => MaybePromise<T>) => Promise<T> {
-  return (ctx, fn) => run(ctx, fn, contextFn);
+): <T>(ctx: C, fn: (ctx: C) => MaybePromise<T>, opts?: RunOptions) => Promise<T> {
+  return (ctx, fn, opts) => run(ctx, fn, contextFn, opts);
 }
 
 /** Runs the `fn` in a dedicated / non-test Unit of Work for each value in `values */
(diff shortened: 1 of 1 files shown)


packages/test-utils/src/run.ts
```diff
@@ -1,31 +1,36 @@
...
+import { Entity, EntityManager, isDefined, isEntity } from "joist-orm";
 import { fail } from "joist-utils";
 import { Context } from "./context";
 
 type MaybePromise<T> = T | Promise<T>;
 
 export type ContextFn<C> = (ctx: C) => MaybePromise<C>;
 
+type RunOptions = {
+  expectUnsavedEntities?: boolean;
```
stephenh, line 10: Thought: maybe `skipNewEntities`?

I'm also kinda wondering if, if this is for testing dry runs, should we just calling this like `dryRun: true` or `reloadEntities: false`, and have it skip the load "translate loaded entities" entirely?

packages/test-utils/src/run.ts
```diff
@@ -1,31 +1,36 @@
-import { Entity, EntityManager, isEntity } from "joist-orm";
+import { Entity, EntityManager, isDefined, isEntity } from "joist-orm";
 import { fail } from "joist-utils";
 import { Context } from "./context";
 
 type MaybePromise<T> = T | Promise<T>;
 
 export type ContextFn<C> = (ctx: C) => MaybePromise<C>;
 
+type RunOptions = {
```
zgavin, line 9: chore:  this should be exported

packages/test-utils/src/run.ts
```diff
@@ -72,24 +77,31 @@ function gatherEntities(result: any): Entity[] {
...
-            e.idTaggedMaybe ?? fail("Joist 'run' returned an unsaved entity; are you missing an em.flush?"),
-          ),
-      )
+      .filter((e) => {
+        if (e.idTaggedMaybe === undefined) {
+          if (opts?.expectUnsavedEntities === true) return false;
+          fail("Joist 'run' returned an unsaved entity; are you missing an em.flush?");
+        }
+        return !em.findExistingInstance(e.idTaggedMaybe);
+      })
```
zgavin, line 91: issue: this code would actually fail in `load` if `assignNewIds` had been called during the test run, but no flush had been called.  The previous implementation had the same fault, but since it never intended to support this behavior it wasn't an issue.  You should use `isNewEntity` instead.

nit: this implementation is getting a little convoluted, it seems more readable to reformat it.  Incorporating @stephenh's naming suggestion:



```typescript
const { dryRun = false } = opts ?? {};
if (!dryRun && newEmEntities.some((e) => e.isNewEntity) )
  fail("Joist 'run' returned an unsaved entity; are you missing an em.flush?");

await Promise.all(
    newEmEntities
      .filter((e) => !e.isNewEntity)
      .filter((e) => !em.findExistingInstance(e.idTagged))
      .map((e) => em.load(e.idTagged))
);
```

packages/test-utils/src/run.ts
```diff
@@ -72,24 +77,31 @@ function gatherEntities(result: any): Entity[] {
...
+    newEmEntities
+      // Exclude entities which have not been flushed, so that they don't fail the lookup
+      .filter((e) => isDefined(e.idTaggedMaybe))
+      .map((e) => [e.id, em.findExistingInstance(e.idTagged) as Entity]),
+  );
   function doMap(value: any): any {
     if (isEntity(value)) {
-      return cache[value.id];
+      // If the value has no ID, simply return it as is
+      return value.idMaybe ? cache[value.id] : value;
```
zgavin, line 104: issue: this seems problematic as it's effectively breaking the `em` isolation that is the entire point of `run`.  This should really be a clone from the old `em` to the new.   Also as pointed out previously, it would be possible for an entity to have an id at this point but not be in the cache (if the previous issue was fixed).

packages/test-utils/src/run.ts
```diff
@@ -72,24 +77,31 @@ function gatherEntities(result: any): Entity[] {
...
+        return !em.findExistingInstance(e.idTaggedMaybe);
+      })
       .map((e) => em.load(e.idTagged)),
   );
   // generate a cache of id -> entity in original em
-  const cache = Object.fromEntries(newEmEntities.map((e) => [e.id, em.findExistingInstance(e.idTagged) as Entity]));
+  const cache = Object.fromEntries(
+    newEmEntities
+      // Exclude entities which have not been flushed, so that they don't fail the lookup
+      .filter((e) => isDefined(e.idTaggedMaybe))
```
zgavin, line 98: chore: this should also be `!e.isNewEntity`

packages/test-utils/src/run.ts
zgavin: I also would prefer a less verbose name.  `dryRun` seems more natural to me, imo.

zgavin:
@stephenh @chr1sjf0x if we really want to support dry runs, maybe what we should  go all the way and make it an option on `em.flush` /  `em.transaction` or the `em` itself.  We could then automatically rollback the transaction before committing it and then keep a list of entities that weren't actually inserted for `mapToOriginalEm`

stephenh:
@chr1sjf0x cc @zgavin thinking about the "testing a resolver in dry-run mode" use case, I think that makes sense, but wonder if we could support it w/o having `mapToOriginalEm` end up returning the user "some instances from em1 / some instances from em2", which I think would be just kind of fundamentally confusing...

I wonder if we could either:

* Provide a `RunOptions`-like param that was like `observeCtx: (ctx) => ...` that, if set, would be called with the inner context right before the `return`, to let the test do any assertions it wants, albeit/granted it would have to find them somehow in the inner context
* Provide a `RunOptions`-like param that just completely disables/skips `mapToOriginalEm`, and returns the values as-is. Granted, the test won't be able to use instance/=== comparison between the returned entities, and its originally setup entities, but :shrug: that seems like what the test is specifically asking to do anyway.

@chr1sjf0x would either of those approaches facilitate your tests/assertions in a sufficiently-ergonomic manner?

chr1sjf0x:
@stephenh / @zgavin - I think disabling the `mapToOriginalEm` feature for such cases feels like the most straight forward approach.

I do like @zgavin's idea of an option on `em.transaction` for automatic rollback on dry runs, that would cleanup my current implementation. But I think that would still require the option to disable `mapToOriginalEm` in the test.

PR #1143, opened 2024-07-15
feat: Add recursive collections

stephenh:
* Any self-referential FK nows gets `parentsRecursive` and `childrenRecursive` relations that use pg recursive CTEs to load N levels up/down in a single SQL call
* Adds a new `ReadOnlyCollection` & `RecursiveChildrenCollectionImpl` / `RecursiveParentsCollectionImpl` relations--these relations don't actually hold any data, but instead just a) kick off the CTE (via corresponding data loaders, so the lookups are still auto-batched) calls, and then b) otherwise just sit on top of the existing o2m/m2o infra/relations, and their `.get` methods just recurse up/down through the o2m/m2o until terminating
* Because of this approach (going through the stock o2m/m2o relations), the recursive relations continue to see any WIP changes to the entity graph :tada:


packages/codegen/src/generateEntityCodegenFile.ts
```diff
@@ -85,14 +88,13 @@ type Relation = { kind: "abstract"; line: Code } | { kind: "concrete"; fieldName
...
 /** Creates the base class with the boilerplate annotations. */
 export function generateEntityCodegenFile(config: Config, dbMeta: DbMetadata, meta: EntityDbMetadata): Code {
+  const { entitiesByName: metasByName } = dbMeta;
   const { entity, tagName } = meta;
   const entityName = entity.name;
 
   // Avoid using `do` as a variable name b/c it's a reserved keyword
   const varName = keywords.includes(tagName) ? uncapitalize(entityName) : tagName;
 
-  const metasByName = groupBy(dbMeta.entities, (e) => e.name);
```
stephenh, line 94: Unrelated, but `DbMetadata` has this `groupBy` built-in now, so moved to using that.

packages/codegen/src/inheritance.ts
```diff
@@ -11,7 +11,7 @@ import { fail } from "./utils";
 export function applyInheritanceUpdates(config: Config, db: DbMetadata): void {
   const { entities, entitiesByName } = db;
   setClassTableInheritance(entities, entitiesByName);
-  expandSingleTableInheritance(config, entities);
+  expandSingleTableInheritance(config, entitiesByName, entities);
```
stephenh, line 14: Unrelated fix up, the "expand STI" logic wasn't adding the `TaskOld` / `TaskNew` inferred/discovered metas to `entitiesByName`, so we pass that along to get updated.

packages/orm/src/QueryParser.ts
```diff
@@ -86,6 +86,8 @@ export interface ParsedFindQuery {
   condition?: ParsedExpressionFilter;
   /** Any optional orders to add before the default 'order by id'. */
   orderBys: ParsedOrderBy[];
+  /** Optional CTE to prefix to the query, i.e. for recursive relations. */
+  cte?: { sql: string; bindings: readonly any[] };
```
stephenh, line 90: This is kind of "bag on the side", but adding a "blob of CTE sql" to `ParsedFindQuery`...

The idea is that we should be able to use the CTE basically as a fancy `where` clause, and then otherwise reuse the "how to conditionally join & loader CTI tables" that `ParsedFindQuery` can already represent.

packages/orm/src/drivers/PostgresDriver.ts
```diff
@@ -52,13 +52,15 @@ export class PostgresDriver implements Driver {
     parsed: ParsedFindQuery,
     settings: { limit?: number; offset?: number },
   ): Promise<any[]> {
-    const knex = this.getMaybeInTxnKnex(em);
-    return buildKnexQuery(knex, parsed, { limit: em.entityLimit, ...settings });
+    const { sql, bindings } = buildRawQuery(parsed, { limit: em.entityLimit, ...settings });
+    return this.executeQuery(em, sql, bindings);
```
stephenh, line 56: I couldn't find an easy way to prefix our existing `buildKnexQuery` with "a blob of CTE sql", because: a) you can either use knex.raw or knex.from but not both together (afaict) and b) their CTE support wants to be fluent DSL like `knex.withRecursive(etc => ...)` and I just wanted to hand it the raw sql.

So, I went ahead and moved this over to a `buildRawQuery` that creates our SQL strings directly, without going through knex builders.

Thanks to the AST-ification of `ParsedFindQuery`, this was actually surprisingly easy to do; I'd been meaning to do it for awhile anyway, to decouple us from knex, and make it easier to explore alternative drivers.

packages/orm/src/drivers/buildKnexQuery.ts
```diff
@@ -73,13 +75,8 @@ export function buildKnexQuery(
...
 
-  if (limit) {
-    query.limit(limit);
-  }
-
-  if (offset) {
-    query.offset(offset);
-  }
+  if (limit) query.limit(limit);
+  if (offset) query.offset(offset);
```
stephenh, line 79: Unrelated changes while reading the file

packages/orm/src/drivers/buildRawQuery.ts
```diff
@@ -0,0 +1,161 @@
...
+/**
+ * Transforms `ParsedFindQuery` into a raw SQL string.
+ *
+ * In theory this should be implemented within each Driver, but the logic will be largely
+ * the same for different dbs.
+ */
+export function buildRawQuery(
+  parsed: ParsedFindQuery,
+  settings: { limit?: number; offset?: number },
+): { sql: string; bindings: readonly any[] } {
```
stephenh, line 18: This is a copy/paste of `buildKnexQuery`, with the knex calls rewritten to string building.

packages/orm/src/drivers/buildRawQuery.ts
```diff
@@ -0,0 +1,161 @@
...
+  if (offset) {
+    sql += ` OFFSET ?`;
+    bindings.push(offset);
+  }
+
+  return { sql, bindings };
+}
+
+/** Returns a tuple of `["cond AND (cond OR cond)", bindings]`. */
+function buildWhereClause(exp: ParsedExpressionFilter, topLevel = false): [string, any[]] | undefined {
```
stephenh, line 93: These two `buildWhereClause` / `buildRawCondition`s are literally unchanged I think; could probably move to like a `buildUtils` and share between the two...

packages/orm/src/keys.ts
```diff
@@ -77,9 +77,8 @@ export function keyToTaggedId(meta: HasTagName, dbValue: string | number): Tagge
 
 /** Fails if any keys are tagged; used by internal functions b/c we still allow most direct API input to be untagged. */
 export function assertIdsAreTagged(keys: readonly string[]): void {
-  const invalidKeys = keys.filter((k) => k.indexOf(tagDelimiter) === -1);
```
stephenh, line 80: i want this to be as cheap as possible, so rewrote to remove the lambda

packages/orm/src/loadHints.ts
```diff
@@ -38,16 +40,18 @@ export type MarkLoaded<T extends Entity, P, UH = {}> =
       ? LoadedReference<T, Loaded<U, UH>, N>
       : P extends Collection<MaybeBaseType, infer U>
         ? LoadedCollection<T, Loaded<U, UH>>
-        : P extends AsyncProperty<MaybeBaseType, infer V>
-          ? // prettier-ignore
-            [V] extends [(infer U extends Entity) | undefined]
+        : P extends ReadOnlyCollection<MaybeBaseType, infer U>
+          ? LoadedReadOnlyCollection<T, Loaded<U, UH>>
```
stephenh, line 44: The indentation makes it look like a bigger change, but just added `ReadOnlyCollection` checking ... at some point, it would be neat to collapse this huge ternary into something cleaner, i.e. if the `P extends ReadOnlyCollection` could become something like:

```
P extends Loadable<..., infer L> ? L
```

Where each Loadable type could have a generic that pointed to its Loaded counterpart...

packages/orm/src/relations/ManyToOneReference.ts
```diff
@@ -116,7 +116,8 @@ export class ManyToOneReferenceImpl<T extends Entity, U extends Entity, N extend
   }
 
   get isLoaded(): boolean {
-    return this._isLoaded;
+    // Even if `.load()` has not been called, `.get` will detect
+    return this._isLoaded || this.current() === undefined || !!this.maybeFindEntity();
```
stephenh, line 120: I'm kind of surprised it didn't already do this, but if we called `m2o.get` on a technically-unloaded m2o, and it would the entity in the EM anyway, it would go ahead and return it.

We were already relying on this as basically the way to preload m2os (which is the mechanism the recursive dataloaders use), but the `isLoaded` wasn't updated to handle the "I didn't have `.load()` called but turns out I am actually loaded" scenario.

packages/orm/src/relations/ManyToOneReference.ts
```diff
@@ -133,11 +134,16 @@ export class ManyToOneReferenceImpl<T extends Entity, U extends Entity, N extend
     // This should only be callable in the type system if we've already resolved this to an instance,
     // but, just in case we somehow got here in an unloaded state, check to see if we're already in the UoW
     if (!this._isLoaded) {
-      const existing = this.maybeFindEntity();
```
stephenh, line 136: Before this failed anytime `existing` was `undefined`, but really if `current` is `undefined`, i.e. the m2o is unset, then its expected to `maybeFindEntity` to also not return anything.

packages/orm/src/relations/ManyToOneReference.ts
```diff
@@ -341,7 +347,7 @@ export class ManyToOneReferenceImpl<T extends Entity, U extends Entity, N extend
   }
 
   public toString(): string {
-    return `ManyToOneReference(entity: ${this.entity}, fieldName: ${this.fieldName}, otherType: ${this.otherMeta.type}, otherFieldName: ${this.otherFieldName}, id: ${this.id})`;
+    return `ManyToOneReference(entity: ${this.entity}, fieldName: ${this.fieldName}, otherType: ${this.otherMeta.type}, otherFieldName: ${this.otherFieldName}, id: ${this.idMaybe})`;
```
stephenh, line 350: Changed to `idMaybe` so that `toString()` doesn't blow up on validly-unset m2os.

PR #1152, opened 2024-07-19
feat: support `NULL` values on `IN` operator

### LINKED ISSUE #1148: Support rewriting `in: [null]` to `IS NULL`

allangsilva:
Fixes #1148

### DIFF
--- a/packages/orm/src/Aliases.ts
+++ b/packages/orm/src/Aliases.ts
@@ -10,7 +10,7 @@ import {
   getMetadata,
 } from "./EntityMetadata";
 import { ColumnCondition, ParsedValueFilter, RawCondition, makeLike, mapToDb, skipCondition } from "./QueryParser";
-import { ExpressionFilter, getConstructorFromTaggedId, maybeResolveReferenceToId } from "./index";
+import { ExpressionCondition, ExpressionFilter, getConstructorFromTaggedId, maybeResolveReferenceToId } from "./index";
 import { Column } from "./serde";
 import { fail } from "./utils";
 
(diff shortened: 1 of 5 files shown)


packages/orm/src/Aliases.ts
```diff
@@ -360,13 +356,10 @@ class PolyReferenceAlias<T extends Entity> {
 
   private addCondition(comp: PolymorphicFieldComponent, value: ParsedValueFilter<T | TaggedId>): ColumnCondition {
```
allangsilva, line 357: @stephenh I couldn't get it working for complex conditions since this method requires a `ColumnCondition`, should we have a `addConditions` that would return a `ParsedExpressionFilter` when needed multiple conditions to handle a operator value (e.g. `in` with null values) ?

allangsilva: This is the test I'm skipping for now https://github.com/joist-orm/joist-orm/pull/1152/files#diff-e3d5ac49e36bbf97dd949765585ed0fb87430ca1c4caa83d682fea55fcb556deR461

stephenh: Ah yeah ... that was a lot trickier than I anticipated. :thinking: 

Fwiw @allangaldinosilva I hacked on this for an hour or so today (I thought it would be a quicker fix and it was not!) and I technically have it working...

I ended up leaving `mapToDb` as-is, and doing the `in: [1, null]` handling outside of it.

I added this `ConditionBuilder`, originally thinking that was going to help me handle the aliases, but actually it did not :rofl: , but I left it in anyway, because I think its a nice clean-up to what we had their before.

Dunno, take a look and lmk if you have thoughts...

It does seem like there are at least two other cases that need to be updated for the `in: [1, null]`, one is the poly handling code (around `QueryParser.ts:259`) and the other is `EntityAliasImpl.in` (around `Aliases.ts:312`). If you wanted to address those in this PR, that'd probably be great, but we could also leave those to follow-ups.

In particular poly handling is always a little tricky :thinking:

packages/orm/src/Aliases.ts
```diff
@@ -39,28 +39,28 @@ export type Alias<T extends Entity> = {
...
-  gte(value: V | undefined | PrimitiveAlias<V, any>): ColumnCondition | RawCondition;
-  lt(value: V | undefined | PrimitiveAlias<V, any>): ColumnCondition | RawCondition;
-  lte(value: V | undefined | PrimitiveAlias<V, any>): ColumnCondition | RawCondition;
-  like(value: V | undefined): ColumnCondition;
-  ilike(value: V | undefined): ColumnCondition;
-  search(value: V | undefined): ColumnCondition;
-  between(v1: V | undefined, v2: V | undefined): ColumnCondition;
+  eq(value: V | N | undefined | PrimitiveAlias<V, any>): ExpressionCondition;
+  ne(value: V | N | undefined | PrimitiveAlias<V, any>): ExpressionCondition;
+  in(values: (V | null)[] | undefined): ExpressionCondition;
```
stephenh, line 44: Changing this to return an `ExpressionCondition` is what lets `in` conditionally return a `{ or: ... }` when it notices the `null` value...

packages/orm/src/Aliases.ts
stephenh: Sounds good! I'll hit merge. :rocket:

PR #1153, opened 2024-07-19
fix: Lightly optimize followReverseHint.

stephenh:
This was not actually a hot spot, but still good to avoid lambdas when iterating over 10k of entities.

### DIFF
--- a/packages/orm/src/reactiveHints.ts
+++ b/packages/orm/src/reactiveHints.ts
@@ -32,7 +32,7 @@ import {
 import { LoadedOneToOneReference } from "./relations/OneToOneReference";
 import { ReactiveGetterImpl } from "./relations/ReactiveGetter";
 import { AsyncPropertyImpl } from "./relations/hasAsyncProperty";
-import { fail, mergeNormalizedHints } from "./utils";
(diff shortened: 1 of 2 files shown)


packages/orm/src/reactiveHints.ts
```diff
@@ -318,44 +318,42 @@ function maybeAddTypeFilterSuffix(
...
+      if (isManyToMany) {
+        const m2m = c[fieldName] as ManyToManyCollection<any, any>;
+        const joinRows = getEmInternalApi(m2m.entity.em).joinRows(m2m);
+        // Return a tuple of [currentRows, removedRows]
+        promises.push(joinRows.removedFor(m2m, c));
+      }
+    }
+    const nextLevel = await Promise.all(promises);
+    // Make a new set so that we dedupe as we go
+    current = flatAndUnique(nextLevel);
```
stephenh, line 354: One of the actual changes here is that we `new Set`/dedupe entities as we crawl through the hint, instead of just at the end, i.e. for `template.items.options.foo`, the items will be ~1000 some items, but but only ~40 unique options across them, so if we unique as we go, we only call `foo.load()` 40 times instead of ~1000.

PR #1162, opened 2024-07-28
fix: simple pg connection string ssl handling

### LINKED ISSUE #1161: newPgConnectionConfig does not handle ssl settings specified in connection string
`newPgConnectionConfig` currently does its own parsing of the DATABASE_URL here:

https://github.com/joist-orm/joist-orm/blob/004dad18dbeb69430fe66481ec39cc5f2354519c/packages/utils/src/connection.ts#L37-L48

This does not pass through any SSL options specified in the connection string (e.g. `sslmode=require`) and means it is impossible to use with any databases that require this, notably Vercel Postgres.

It would be good to at least handle `ssl=true` to mitigate this.

It seems to me given `createKnex` and the `Client` constructor from the `pg` module both already allow a connection string to be passed directly and handle SSL options correctly, it would be better just to pass the connection string directly in these cases?

The exception is `Pool` which requires an object.

I have made a very simple patch but by all means disregard if there is a better alternative.

acgray:
Simple approach to resolve #1161

### DIFF
--- a/packages/utils/src/connection.test.ts
+++ b/packages/utils/src/connection.test.ts
@@ -9,6 +9,7 @@ describe("connection", () => {
       password: "local",
       port: 5432,
       user: "joist",
+      ssl: false,
     });
   });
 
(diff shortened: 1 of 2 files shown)


stephenh (approved):
This is great @acgray , thank you!

PR #1169, opened 2024-08-02
feat: Native support for regex operators on find

chr1sjf0x:
Direct support for [PSQL regular expression](https://www.postgresql.org/docs/current/functions-matching.html#FUNCTIONS-POSIX-REGEXP) where clauses.

### DIFF
--- a/packages/orm/src/EntityFilter.ts
+++ b/packages/orm/src/EntityFilter.ts
@@ -74,6 +74,10 @@ export type ValueFilter<V, N> =
   | { nlike: V | undefined }
   | { ilike: V | undefined }
   | { nilike: V | undefined }
+  | { reg?: V | undefined }
+  | { ireg?: V | undefined }
+  | { nreg?: V | undefined }
+  | { nireg?: V | undefined }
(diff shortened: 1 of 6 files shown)


packages/orm/src/EntityFilter.ts
```diff
@@ -74,6 +74,10 @@ export type ValueFilter<V, N> =
   | { nlike: V | undefined }
   | { ilike: V | undefined }
   | { nilike: V | undefined }
+  | { reg?: V | undefined }
```
stephenh, line 77: @chr1sjf0x nice! Could prefer for these to be `regex/iregex/nregex/niregex`, just given the ubiquity of "regex" as a term, but otherwise looks great!

chr1sjf0x: Change made 👍 

Ready for merge?

stephenh (approved):

packages/orm/src/EntityFilter.ts
stephenh: Yep, feel free to merge when the build passes

PR #1188, opened 2024-08-20
feat: add tstzrange field support

blambillotte:
Treat `tstzrange` timezone range fields as strings so the implementor can handle the [upper and lower bound](https://www.postgresql.org/docs/current/rangetypes.html#RANGETYPES-IO) syntax how they want. 

Potentially, joist could parse these values and return a typed object? Something like:
```ts
type TstzRange {
  lowerBound: Date;
  lowerBoundType: RangeBoundaryType.Inclusive ;
  upperBound: Date;
  upperBoundType: RangeBoundaryType.Exclusive; 
}
```

Or there are other examples out there like from this pacakge: https://github.com/WhoopInc/node-pg-range
```ts
  range = {
    begin: Thu Mar 26 2014 17:57:02 GMT-0400,
    end: Thu Mar 27 2014 17:57:02 GMT-0400,
    bounds: '[)'
  }
```


blambillotte:
@stephenh what do you think about just treating range fields as strings as an initial approach?

stephenh:
@blambillotte I think `string` makes sense for now / as a default--seems like it would be hard to pick a default/winning range impl that everyone would like atm.

That said, this could be a good time (not this PR but this use case) to refine joist's "user land type" / "custom serde" feature, i.e. we can already say "the User.password field should go through this custom serde type": 

https://github.com/joist-orm/joist-orm/pull/698/files#diff-b865f445f30d3b851378389c4b10d00196f46b6b6e4e76409a73f8083a6634a4R47

But currently: a) it has to be configured per-field, and we should also have a default db-wide setting (i.e. always maybe timestamptzrange as ...), and b) currently it doesn't get to control deserialization of the type from postgres, but that's probably an advanced feature anyway.

So, yeah, maybe just strings for this PR--can you make a dummy tstzrange field in one of our `integration-tests` entities/tables? Just to get some trivial test coverage of it.

But otherwise sgtm!

stephenh (approved):

PR #1197, opened 2024-09-04
chore: update node test versions

blimmer:
This PR updates the versions of Node you're testing against. See inline comments for details.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  node: circleci/node@5.2.0
+  node: circleci/node@6.1.0
 
 workflows:
   version: 2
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -10,7 +10,10 @@ workflows:
       - build:
           matrix:
             parameters:
-              node-version: ["18.12.0", "20.8.1", "21.1.0", "22.0.0"]
+              node-version:
+                - "18" # EOL 30 Apr 2025
+                - "20" # EOL 30 Apr 2026
+                - "22" # EOL 30 Apr 2027
```
blimmer, line 16: This orb uses `nvm` under the hood and will pull the most recent in the major range:

Screenshot 2024-09-04 at 17 44 06

.circleci/config.yml
```diff
@@ -49,7 +52,7 @@ jobs:
           condition:
             and:
               - equal: [main, << pipeline.git.branch >>]
-              - equal: ["21.1.0", << parameters.node-version >>]
+              - equal: ["22", << parameters.node-version >>]
```
blimmer, line 55: 22 hits LTS in October, so this is probably safe. Either way, we shouldn't use 21 since it's not supported anymore.

blimmer (commented):
You'll also need to update your branch protection rules to the new job names.

stephenh (approved):

PR #1208, opened 2024-09-11
fix: support single quotes in scanEntityFiles

### LINKED ISSUE #1202: Auto-detect setDefault assumes double quotes
Fun one working out why this wasn't working over in single quote land :)

Happy to PR. Wondering if it's worth inverting the flow here to do a single global regex matchAll to find usages and then match them to the DbMetadata or just adding an additional .includes for single quotes.

brudil:
Turns out it was already rewritten to do the inverted regex flow - I was looking at the inital PR which added it, not main.

- Add adds the single quote to the regex. Fixes #1202
- Changes one of the integration entity setDefault calls to use single quotes w/ prettier ignore comment - I assume this should suffice as a decent enough check.


stephenh:
Thanks!

Issue #1211, opened 2024-09-11
Add oracle mode / dry-run support


stephenh:
We're likely going to use Joist in "oracle mode", which entities are mutated (and potentially `em.flush` called) to "see what happens", but then _avoid_ committing the data to the database.

There are a few options:

1. Just add a `em.flush({ dryRun: true })` flag and that is it
   * Pro: Makes a lot of sense, we've been meaning to add at some point
   * Con: Each caller to `flush` has to know "we're oracling"
2. Add a `em.mode = "read-only" | "mutable-dry-run" | "writable"`
   * `read-only` => entities are 100% immutable, `mutable-dry-run` => this oracle mode, entities can be mutated & `em.flush` called to run hooks, validation rules, reactive fields, but nothing is committed, `writeable` => the default behavior we have today
   * Pro: A single LOC can flip EM into oracle mode and not worry about rando em.flushes committing things
   * Con: Seems like overkill for simpler usages that could use just the `em.flush(dryRun)` opt
3. Add both

PR #1216, opened 2024-09-13
feat: Add EntityManager.mode.

stephenh:
Allows putting the EM into either read-only (no mutations allowed) or in-memory-writes (mutations allowed but no writes).

The in-memory-writes in particular is useful for building "oracle" functionality where you want to mutate entities to "see what would happen", but ensure the WIP data is never committed to the db, even by an errant, over-eager `em.flush`.

### DIFF
--- a/packages/tests/integration/src/setupTestEnv.ts
+++ b/packages/tests/integration/src/setupTestEnv.ts
@@ -2,5 +2,7 @@ import { GetEnvVars } from "env-cmd";
 
 export default async function globalSetup() {
   process.env.TZ = "UTC";
+  // process.env.DEBUG = "knex:query,bindings";
+  // process.env.DEBUG = "knex:*";
   Object.entries(await GetEnvVars()).forEach(([key, value]) => (process.env[key] = value));
 }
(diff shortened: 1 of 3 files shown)


stephenh:
Fixes #1211

packages/tests/integration/src/setupTestEnv.ts
```diff
@@ -2,5 +2,7 @@ import { GetEnvVars } from "env-cmd";
 
 export default async function globalSetup() {
   process.env.TZ = "UTC";
+  // process.env.DEBUG = "knex:query,bindings";
+  // process.env.DEBUG = "knex:*";
```
chr1sjf0x, line 6: question: Was this meant to be left in?

chr1sjf0x (approved):
praise: Nice this was so straight forward 👍

packages/tests/integration/src/setupTestEnv.ts
stephenh: Yeah, I hadn't used it before, and wanted to leave it in as a reminder to myself + quick "just uncomment this".

Eventually thinking Joist could support the same sort of thing, i.e. what is now some of our `setFieldLogging(true)` type flags could instead (or also) be like `DEBUG=joist:fields` type settings. :thinking: 

I think the `DEBUG=...` approach is fairly idiomatic within the node ecosystem, or at least was at some point, it seems as little more OG, but not sure what the new alternative is atm.

PR #1217, opened 2024-09-15
feat: Add cross-entity default dependencies.

stephenh:
Historically we've had issues with cross-hook `beforeCreate` dependencies, i.e. waiting for `Author.beforeCreate` to set a field that `Book.beforeCreate` wanted to read, to set its own book fields.

At least for setting defaults, if all callers use `config.setDefault("theirField", ...)`, this issue is now fully resolved, as we will use the `setDefault` field-level hints to track dependencies (i.e. that `Book.authorNickNames` depends on `Author.nickNames`, both of which are async default fields), and run the upstream fields first.

We had already been doing this for sync vs. async defaults, i.e. we'd always run sync defaults first, but this extends the inter-dependency tracking to async defaults as well.


stephenh:
The easiest way for me to test this, is to mash deploy and test it on our internal app, then do fixups, so doing that now...

PR #1218, opened 2024-09-15
fix: Missing kqDot within distinct/orderBy

brudil:
After renaming a entity to something abreviated to a psql keyword, I hit the following;

```
SELECT DISTINCT "to".*, to.id FROM tender_offer AS "to" LEFT OUTER JOIN other_table AS ot ON "to".id = ot.some_id WHERE ot.other_id = $1 ORDER BY "to".id ASC LIMIT $2 - syntax error at or near "to""
```

Notice the second `to` being unquoted within the select. I suspect this is due to the locations identified within the PR not using kqDot.


stephenh (approved):
Looks great! Thanks!

PR #1220, opened 2024-09-16
fix: continue instead of return on config warnings

brudil:
Updates `warnInvalidConfigEntries` to move to the next itteration if the entity within the config is not found, instead of returning early. This allows all found issues to be printed in one execution, vs needing to manually fix-run ad nasuem.

edit; thoughts on a --strict flag or something to return a non-zero exit code if any of these warnings are printed? We would find something like this very helpful for CI

### DIFF
--- a/packages/codegen/src/config.ts
+++ b/packages/codegen/src/config.ts
@@ -139,7 +139,7 @@ export function warnInvalidConfigEntries(config: Config, db: DbMetadata): void {
     const entities = entitiesByName[entityName];
     if (!entities) {
       console.log(`WARNING: Found config for non-existent entity ${entityName}`);
-      return;
+      continue;
     }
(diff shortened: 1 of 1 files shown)


stephenh:
I think a `--strict` flag makes sense; maybe we could have our own mini-`Logger` with a `logger.warn` and then at the end of the execution, ask "hey logger did you do any warns"?.

Usually I'd pass around a ctx/logger as a param, but this is just a CLI, so could probably do a `export const logger = ...` and have the misc call sites just import it. :shrug: Would be a lot simpler.

If you'd like to work on that, or another approach that achieved the same goal of driving the exit code = 1, that'd be great @brudil ! Thanks!

stephenh (approved):

PR #1227, opened 2024-09-18
feat: sync lens

### LINKED ISSUE #277: Add synchronous version of lens traversal
I.e. so that:

```typescript
const tasksByDefinitionId = target.stages.get
  .flatMap((s) => s.units.get)
  .flatMap((u) => u.activities.get)
  .flatMap((a) => a.tasks.get)
  .filter((t) => t.taskDefinition.isSet)
  .mapToObject((t) => [t.taskDefinition.idOrFail, t]);
```

Could be:

```typescript
const tasksByDefinitionId = target
  .get((p) => p.stages.units.activities.tasks)
  .filter((t) => t.taskDefinition.isSet)
  .mapToObject((t) => [t.taskDefinition.idOrFail, t]);
```

brudil:
This implements a `get(fn: GetLens<this>)` on entities. Fixes #277

A new `GetLens` mapped type, first infers on a GetLike, then removes those that are LoadLike via key mapping. This allows for the types to only allow you to sync get what is loaded.

The runtime implementation uses the existing getLens directly, which seems to be working well in tests. As it takes a `Lens` and not a `GetLens` I've brute forced this with an `as never` for now, but happy to rework if desired.

Hit `Type of property 'fullNonReactiveAccess' circularly references itself in mapped type` so added an Omit to the `this` within the GetLens generic.


stephenh (approved):
Whoa! :exploding_head: @zgavin check this out!

@brudil we have been musing about "man, a sync version of .load would sure be nice" for like, way too long, but never made it to the top of the list.

I love how simple the implementation is as well; great job! Thank you!

PR #1228, opened 2024-09-18
feat: add codegen logger + exit code w/ strict flag

brudil:
Following on from thoughts in #1220, this implements a `logger` with `warn` and `error` methods.

Previous console.log calls starting with either `ERROR:` or `WARNING:` have been updated to the new methods.

Internally we keep track of the highest called log level, which is recalled at executions end and is used to determine an exitCode if any.

Either
- log level >= error;
- log level >= warn & `--strict` flag used

will result in exitCode = 1


brudil:
Looks like `joist-tests-immediate-foreign-keys` is failing this, I'll defer on what is best to do here (an ignore flag for testing perhaps?, cluge in the package.json codegen script?) Although a breaking change, I think the above logic is good for consumers of Joist.

```
[joist-tests-immediate-foreign-keys]: ERROR: Found a schema cycle of not-null foreign keys:
[joist-tests-immediate-foreign-keys]: T3Book -> T3Author
[joist-tests-immediate-foreign-keys]: T4Book -> T4Author
[joist-tests-immediate-foreign-keys]: 
[joist-tests-immediate-foreign-keys]: These cycles can cause fatal em.flush & flush_test_database errors.
[joist-tests-immediate-foreign-keys]: 
[joist-tests-immediate-foreign-keys]: Please make one of the FKs involved in the cycle nullable.
```

packages/codegen/src/index.ts
```diff
@@ -138,7 +140,16 @@ if (require.main === module) {
   if (Object.fromEntries === undefined) {
     throw new Error("Joist requires Node v12.4.0+");
   }
-  main().catch((err) => {
+  main().then(() => {
+    if (
+        // strict mode + warnings or greater
+        (process.argv.includes("--strict") && loggerMaxWarningLevelHit >= LOG_LEVELS.warn) ||
```
stephenh, line 146: Nice!

Per the immediate-fks thing, right, I'm kinda purposefully/lazily using that codegen failure as a poor-mans test harness, vs. making a truly isolated unit test that faked everything out and could trigger/cover the same behavior (which is definitely the most kosher approach, long-term)...

Dunno, maybe we could add a `--never-fail` / `--ignore-errors` / `--always-exit-code-zero` type check here, that is really only intended for that one immediate-fks test `package.json` to use... I think that'd work? And not to too bad...

stephenh (approved):

packages/codegen/src/index.ts
brudil: Pushed with --always-exit-code-zero as i think it's the clearest.

PR #1229, opened 2024-09-18
chore: add format check to gha pr

brudil:
Adds a GHA workflow for PRs to run prettier w/ `--check` flag. This will ensure PRs have run `yarn format` before being merged.

I'm definitely to blame for forgetting it a couple of times :), which then means when it is run it's sprawling and often unrelated to the PR changes itself.

Purely pushed the workflow so far, so this should fail.


packages/tests/esm-misc/src/entities/codegen/AuthorCodegen.ts
```diff
@@ -204,13 +204,13 @@ export abstract class AuthorCodegen extends BaseEntity<EntityManager, string> im
   }
 
   get books(): Collection<Author, Book> {
-    return this.__data.relations.books ??= hasMany(
+    return (this.__data.relations.books ??= hasMany(
```
stephenh, line 207: You might already be on this, but can we setup `.prettierignore`/etcs for any of the codegen files? Just since it's ~expected the dprint formatter we use is "prettier-ish" and so will always have some drift...

brudil: That's a shout. On codegen, I've just added #1231 if you're open to something like that.

brudil: each test package has different prettierignores, so I've raised it to the root and ensured that all **/entities/codegen/* files are ignored.

stephenh:
@brudil looks great! Maybe run `yarn codegen` to restore the fixed-up files back to the dprint "prettier-ish" format?

Er, I tried to do that locally, but looks like `yarn format --check` is still fixing up `*Codegen` files in some of the test repos. :thinking: Didn't really look into why, so maybe I did something wrong...

brudil:
I had run that in the last commit, and whoops - I thought that was successfully restoring the un-prettiered version. :/

Looks like prettier doesn't walk up to find the ignore (but it does for the config), I've added the --ignore-path arg in the latest commit.

Not sure if yarn supports top level deps, but if so, might be worth considering moving prettier there, and operating on the monorepo as a whole in the future?

stephenh (approved):
This is great! Thanks!

Issue #345, opened 2022-05-30
Add field information to ValidationError


stephenh:
Currently `ValidationError` has entity-level knowledge of what caused an error:

```
export type ValidationError = { entity: Entity } & GenericError;
```

But it doesn't have field-level knowledge. Ideally this should look like:

```
export type ValidationError = { entity: Entity; field?: string } & GenericError;
```

Bonus points would be also defining some sort of generic errors data structures that we could hook into form-state, i.e. something like a GraphQL response of:

```graphql
type ValidationErrors {
  errors: [ValidationError!]!
}

type ValidationError {
  entity: ID,
  field: string,
  message: string!
}
```

So that a form with the author's first name, could get a list of errors, and know to map `entity=a:1` + `field=firstName` errors like `Required` to the text box.

I.e. we may want a prefix-less error message like:

```
{ entity: Author, field: "firstName", message: "Required" }
```

That when for `ValidationError.message` we build up a string like `Author:1 firstName is required", with some hand-waving around "who uppercases or lowercases Required"? 

Error messages to think about:

* "firstName is required"
* "status cannot be changed"
* "Book cannot be deleted"

PR #1232, opened 2024-09-18
feat: optional field & code on ValidationError

brudil:
This riffs off of #345.

I've assume the intent here is to allow rules to return this infomation vs _automagically_ determining the field.

I've verged slightly from the ideas within the issue, by introducing a second new property, `code?: string`. This avoids the overloading of `message` as a user-readable and-also programatic value.

I thought this was important as already we have rules such as min/max number where the message can encode more infomation. You would then be left with a decision of either making it static and easy to find or dynamic and hard.

### DIFF
--- a/packages/orm/src/rules.ts
+++ b/packages/orm/src/rules.ts
@@ -11,7 +11,7 @@ import { MaybePromise, groupBy, maybePromiseThen } from "./utils";
  *
  * Consumers can extend `GenericError` to add fields relevant for their application.
  */
-export type ValidationRuleResult<E extends GenericError> = string | string[] | E | E[] | undefined;
+export type ValidationRuleResult<E extends GenericError> = string | string[] | E | E[] | {field: string, msg: string } | undefined;
 
 /** An entity validation rule. */
 export type ValidationRule<T extends Entity> = (entity: T) => MaybePromise<ValidationRuleResult<any>>;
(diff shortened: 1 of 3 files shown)


packages/orm/src/rules.ts
```diff
@@ -11,7 +11,13 @@ import { MaybePromise, groupBy, maybePromiseThen } from "./utils";
  *
  * Consumers can extend `GenericError` to add fields relevant for their application.
  */
-export type ValidationRuleResult<E extends GenericError> = string | string[] | E | E[] | undefined;
+export type ValidationRuleResult<E extends GenericError> =
+  | string
+  | string[]
+  | E
+  | E[]
+  | { field: string; msg: string }
```
stephenh, line 19: @brudil can we make this `{ fieldName: string; message: string }`?

stephenh (approved):
This is a great idea @brudil !

One minor ask about the key names, but otherwise is great! Thank you!

packages/orm/src/rules.ts
stephenh: ...actually chatting with @zgavin , we're thinking `{ field: string, message: string }`, i.e. field as you have, but message spelled out; wdyt? 

Zach had also mentioned maybe typing this as `field: FieldOf<T>` but then we'd have to pass the `T` generic through, which is doable, but :shrug: I'm more ambivalent about that for this PR.

stephenh:
Fyi @brudil I'm going to close this and land the same change + a few fixups in #1392.

Thanks, and hope all is well!

PR #1236, opened 2024-09-19
fix: Correct handling of setDefault on STI subtypes

chr1sjf0x:
Ensures that `setDefault` works correctly on Single Table Inheritance classes.

Also adds some test cases for sync/async derived fields.


packages/tests/integration/src/SingleTableInheritance.test.ts
```diff
@@ -390,4 +372,40 @@ describe("SingleTableInheritance", () => {
...
+    });
+    // And TaskNew runs its defaults
+    expect(nt).toMatchEntity({
+      syncDefault: "TaskNew",
+      asyncDefault_1: "TaskNew Async1",
+      asyncDefault_2: "TaskNew Async1 Async2",
+    });
+  });
+
+  it("derived fields work as expected", async () => {
```
chr1sjf0x, line 395: NOTE: Despite these tests working here, I'm still experiencing issues with this in graphql-service. I want to ship this as-is, and try to debug the graphql-service use case separately.

packages/tests/integration/src/entities/Task.ts
```diff
@@ -2,7 +2,8 @@ import { TaskCodegen } from "./entities";
 
 import { taskConfig as config } from "./entities";
 
-export class Task extends TaskCodegen {}
```
chr1sjf0x, line 5: @stephenh - I don't know how important testing a concrete base class was. I needed to make it abstract for my test cases, and there wasn't too much changes needed elsewhere to make the desired assertions.

Let me know if you'd like me to introduce a new STI table to keep this one concrete

stephenh (approved):

packages/tests/integration/src/entities/Task.ts
stephenh: Np on making `Task` abstract; I had been doing the same when debugging locally :-) :shrug:

packages/codegen/src/scanEntityFiles.ts
```diff
@@ -13,13 +21,7 @@ export async function scanEntityFiles(config: Config, dbMeta: DbMetadata): Promi
...
         // load the entity.ts file (as a promise with fs/promises)
         const tsCode = (await readFile(join(config.entitiesDirectory, `${entity.name}.ts`))).toString();
-        const defaultableFields = [
-          ...entity.primitives,
-          ...entity.enums,
-          ...entity.pgEnums,
-          ...entity.manyToOnes,
-          ...entity.polymorphics,
-        ];
+        const defaultableFields = recursiveDefaultableFields(dbMeta, entity);
```
stephenh, line 24: @chr1sjf0x is the fix here for a field is defined in a base class / base table, but the `setDefault` call lives only in a specific subtype?

chr1sjf0x: Yes. If you have a field which is shared between the subtypes, so the field itself is defined on the base class, but the subtypes provide different `setDefault` values (and there is no default in the base), then this code used to fail to detect that it should consider the base field to have a default.

Per the note below, I suspect there are issues here, like if one subtype defines a default but another doesn't. Or maybe one subtype wants the field to be async derived while the other wants it editable or sync derived. These sort of cases would need a larger refactor than I was willing to undertake.

PR #1237, opened 2024-09-19
fix: Sync derived fields on STI types

chr1sjf0x:
I've yet to be able to reproduce why this was failing in graphql-service, but working in joist. However, these changes do get it working in graphql-service.

### DIFF
--- a/packages/orm/src/EntityManager.ts
+++ b/packages/orm/src/EntityManager.ts
@@ -1934,15 +1934,16 @@ function recalcSynchronousDerivedFields(todos: Record<string, Todo>) {
     [...new Set(entities.map(getMetadata))].map((m) => {
       return [
         m,
-        Object.values(m.fields)
-          .filter((f) => f.kind === "primitive" && f.derived === "sync")
+        Object.values(m.allFields)
+          .filter((f) => (f.kind === "primitive" || f.kind === "enum") && f.derived === "sync")
           .map((f) => f.fieldName),
       ];
     }),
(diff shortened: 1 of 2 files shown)


packages/orm/src/EntityManager.ts
```diff
@@ -1934,15 +1934,16 @@ function recalcSynchronousDerivedFields(todos: Record<string, Todo>) {
     [...new Set(entities.map(getMetadata))].map((m) => {
       return [
         m,
-        Object.values(m.fields)
-          .filter((f) => f.kind === "primitive" && f.derived === "sync")
+        Object.values(m.allFields)
+          .filter((f) => (f.kind === "primitive" || f.kind === "enum") && f.derived === "sync")
```
chr1sjf0x, line 1938: @stephenh - I'm pretty sure the lack of an enum check here was also a bug. I could add tests if desired.

packages/orm/src/EntityManager.ts
```diff
@@ -1934,15 +1934,16 @@ function recalcSynchronousDerivedFields(todos: Record<string, Todo>) {
...
           .map((f) => f.fieldName),
       ];
     }),
   );
+  if (entities.some((e) => getMetadata(e).type === "TaskNew")) console.log([...derivedFieldsByMeta.entries()]);
```
chr1sjf0x, line 1943:
```suggestion
```

stephenh (approved):

PR #1238, opened 2024-09-19
feat: Revert "feat: sync lens (#1227)" - Temporary

chr1sjf0x:
This reverts commit 1e1043da690533023dfcc771bc5330512bd0e086.

This commit has caused circular dependency type errors in graphql-service which will take some time to resolve - reverting for now.


brudil (approved):

stephenh:
Thanks @chr1sjf0x !

@brudil we've historically seen these "cyclic compile errors" (I'll paste a screenshot when I have time later) that make the domain model's API too even more cyclic than it already is. Honestly I don't really understand why. :-/ 

Understood it's annoying to have something reverted due to a breakage you can't see/debug/triage :-/

brudil:
> Thanks @chr1sjf0x !
> 
> 
> 
> @brudil we've historically seen these "cyclic compile errors" (I'll paste a screenshot when I have time later) that make the domain model's API too even more cyclic than it already is. Honestly I don't really understand why. :-/ 
> 
> 
> 
> Understood it's annoying to have something reverted due to a breakage you can't see/debug/triage :-/

No drama at all. Nature of open source and software in general :) Would be curious if it's possible to distill what triggers it to add a case like this to the integration tests and prevent an issue like this again.

My guess is it's the `this` as part of the return type - there wasn't really a prior usage of that, so I guess this has been the reason before.

If it's not an easy fix, we could consider bringing it back as a top level fn rather than as a method, this should avoid circular typing issues.

stephenh:
@brudil ah yeah, agreed that `this` being in the return type is the likely culprit; would you like to try a reroll without the `this` type, and before merging we'll preemptively test it against our internal app?

I'd love to get this back in...

Fwiw totally agreed we should have a way of exposing these failures in the joist integration tests -- these circular type issues are just tough b/c it seems to be a somewhat arbitrary shape of domain entities or maybe/more likely size (?) of the domain, that tip the complexity too high to where `tsc` complains :thinking: :-/

PR #1240, opened 2024-09-21
fix: Call hookUpBaseTypeAndSubTypes before fireAfterMetadatas

arvinsiva:
Fixes base and sub type metadata access within the `afterMetadata` lifecycle hook by calling `hookUpBaseTypeAndSubTypes` before `fireAfterMetadata` within `configureMetadata`.

### DIFF
--- a/packages/orm/src/configure.ts
+++ b/packages/orm/src/configure.ts
@@ -18,11 +18,11 @@ const typeToMetaMap = new Map<string, EntityMetadata>();
 
 /** Performs our boot-time initialization, i.e. hooking up reactivity. */
 export function configureMetadata(metas: EntityMetadata[]): void {
+  hookUpBaseTypeAndSubTypes(metas);
   fireAfterMetadatas(metas);
   setBooted();
   populateConstructorMaps(metas);
   setImmutableFields(metas);
-  hookUpBaseTypeAndSubTypes(metas);
   reverseIndexReactivity(metas);
   populatePolyComponentFields(metas);
 }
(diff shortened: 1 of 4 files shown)


packages/orm/src/configure.ts
```diff
@@ -18,11 +18,11 @@ const typeToMetaMap = new Map<string, EntityMetadata>();
 
 /** Performs our boot-time initialization, i.e. hooking up reactivity. */
 export function configureMetadata(metas: EntityMetadata[]): void {
+  hookUpBaseTypeAndSubTypes(metas);
```
arvinsiva, line 21: @stephenh Do any of the other functions within `configureMetadata` need to be called before `hookUpBaseTypeAndSubTypes`? I took my best guess at reordering but wouldn't be surprised if this causes other issues I'm not aware of.

stephenh: @arvinsiva well, it's hard to say until a bug happens :laughing: but `populateConstructorMaps` seems "maybe useful?" as does `setImmutableFields` and `populatePolyComponents` fields.

Honestly not entirely sure, but I'd check those.

Fwiw I might just hit merge on this, to unblock you and make the smallest change possible (i.e. mostly likely to get a passing build in `graphql-service` :sweat_smile: ) and then if you wanted to try and move these others around, just on principle/to see what happens, but in a less critical-path PR, that'd be great!

stephenh (approved):

packages/orm/src/configure.ts
stephenh: Actually @arvinsiva my guess from the current test failures is that `setImmutableFields` is probably needing to come first, before `hookUpBaseTypeAndSubTypes`, as it looks like some fields are forgetting that they are immutable.

PR #1247, opened 2024-09-27
feat: Support gt/gte/lte/lt on FK columns.

stephenh:


packages/tests/integration/src/EntityManager.queries.test.ts
```diff
@@ -2693,6 +2693,31 @@ describe("EntityManager.queries", () => {
...
+              // Use all of gt/gte/lte/lt
+              b.author.gt("a:1"),
+              b.author.gte("a:2"),
+              b.author.lte("a:2"),
+              b.author.lt("a:3"),
+            ],
+          },
+        },
+      );
+      expect(books.length).toEqual(1);
```
chr1sjf0x, line 2718: FWIW - would be nice to confirm the expected book was returned

PR #1251, opened 2024-10-01
feat: Update factories to set defaults immediately.

stephenh:
Because factory entities are DeepNew/deeply loaded, it's very likely that what would, in production, usually be an async `setDefault` call (requiring a Promise to load data from the db) can, in a test, actually be done synchronously (by assuming the test instance has the data already in memory).

This PR leans into that assumption, and uses a cute `async () => {}.constructor` check to guess at whether a `setDefault` lambda is synchronous or not.

It's not foolproof but should work.

### DIFF
--- a/packages/orm/src/defaults.ts
+++ b/packages/orm/src/defaults.ts
@@ -1,9 +1,9 @@
 import { Entity } from "./Entity";
-import { EntityMetadata, EnumField, PrimitiveField, getBaseAndSelfMetas, getMetadata } from "./EntityMetadata";
+import { EntityMetadata, EnumField, getBaseAndSelfMetas, getMetadata, PrimitiveField } from "./EntityMetadata";
 import { Todo } from "./Todo";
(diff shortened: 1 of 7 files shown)


packages/orm/src/defaults.ts
```diff
@@ -133,6 +146,34 @@ export class AsyncDefault<T extends Entity> {
...
+    // from a synchronous factory call.
+    if (this.#fn.constructor.name === "AsyncFunction") return;
+    const value = (entity as any)[this.fieldName];
+    try {
+      if (value === undefined || (isLoadedReference(value) && !value.isSet)) {
+        const val = this.#fn(entity, ctx);
+        // Even though we checked `async () => ...` above, a lambda could still return a Promise
+        if (val instanceof Promise) {
+          // ...maybe we should throw here as an invalid usage
+          val.catch(() => {});
```
zgavin, line 160: imo we should definitely throw here

PR #1290, opened 2024-11-06
fix: Fix specialized default dependencies.

stephenh:
A single `config.setDefault("type", { identity: "type" })` might actually have separate dependency resolutions, if subtypes are in place.

See the example in the code, but a regular RPV.type default doesn't have a dependency, b/c the hint's RP.type is not an async default.

But a POV.type, which knows its identity is a PV, it turns out that PV.type does have an async default.

We were already doing some separate of "set defaults on base type separately from defaults on sub types", but weren't letting them have different dependency resolutions.

### DIFF
--- a/packages/orm/src/EntityManager.ts
+++ b/packages/orm/src/EntityManager.ts
@@ -1244,7 +1244,7 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW> {
           // Run our hooks
           let todos = createTodos([...pendingHooks]);
 
-          await setAsyncDefaults(suppressedDefaultTypeErrors, this.ctx, Todo.groupByType(todos));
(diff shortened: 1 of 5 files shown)


packages/orm/src/EntityManager.ts
```diff
@@ -1244,7 +1244,7 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW> {
           // Run our hooks
           let todos = createTodos([...pendingHooks]);
 
-          await setAsyncDefaults(suppressedDefaultTypeErrors, this.ctx, Todo.groupByType(todos));
+          await setAsyncDefaults(suppressedDefaultTypeErrors, this.ctx, Todo.groupInsertsByTypeAndSubType(todos));
```
stephenh, line 1247: Just renaming the method for clarity

PR #1297, opened 2024-11-15
feat: Use TypeMap instead of orm field

### LINKED ISSUE #1295: Maybe use TanStack Router's codegen trick
To drive the type-safety of Joist's APIs, like `em.find`, we have generics like `OptsOf<T>` or `FilterOf<T>` that, given an entity `T` return the codegen-d `AuthorFilters` or `AuthorOpts` or what not.

Today we use a `__orm: { optsType: AuthorOpts }` type-only field to take `T` and return the `AuthorOpts`.

This works well, but it does mean the `__orm: ...lots of types...` is considered part of the `Author` type signature, and so it can throw off type-checking if we don't craft the `Author...` types very carefully.

The high-level problem/goal is simply to "lookup" a given type `T` in a "type map", however:

* `em.find` needs to be able to reference the lookup type, and
* this mapping is fundamentally an application-specific map

This creates a conundrum of how a generic, Joist-module API like `em.find` can use/reference a "type map" that is fundamentally in the user's application code.

One potentially neat trick/approach is using TypeScript's declaration merging, which TanStack Router uses to good effect, by generating TypeScript code that looks like: 

```ts
// this is created by codegen for each individual app
const router = createRouter({
  // ...
})

// this "reaches into" the tanstack types and pushes `typeof router`
// into a location that all of its APIs can reference
declare module '@tanstack/react-router' {
  interface Register {
    router: typeof router
  }
}
```

---

This generally seems like a great idea for us, with the one wrinkle (of course) being inheritance, b/c we want to be able to do:

```ts
const pubOpts = OptsOf<Publisher>; // returns PublisherOpts
const lpOpts = OptsOf<LargePublisher>; // returns LargePublisherOpts
```

But if the implement of the "type map" is something like:

```ts
type OptsTypeMap = {
  Publisher: PublisherOpts,
  LargePublisher: LargePublisherOpts,
}
```

The `Publisher` and `LargePublisher` keys here are just strings, which we'd have to index off of the `T` like:

```ts
type OptsOf<T extends { typeName: string }> = OptsTypeMap[T["typeName"]];
```

But if we do:

```ts
class Publisher {
  declare typeName: "Publisher";
}
class LargePublisher extends Publisher {
  declare typeName: "LargePublisher";
}
```

The `LargePublisher.typeName` will cause a type-error because it's changing the type of `Publisher.typeName`.

The only idea I've had so far to work around this is multiple keys, like:

```ts
type OptsOf<T extends { baseType: string; subType?: string }> =
  T extends { subType: string }
    ? OptsTypeMap[T["subType"]]
    : OptsTypeMap[T["typeName"]];
```

Which I *think* would work, but requires adding a subType key per level of inheritance that we support (which today is only 2, but surely will be at least 3 someday).

stephenh:
Instead of using a hidden `__orm: { ...types... }` field to look up types like the `AuthorOpts` or `AuthorFilter`, we use a hidden `__type: { 0: "Author" } }` field that is "just a string" and then keep the types themselves in a global `TypeMap`, which are later "looked up" by the `OptsOf`, `FilterOf`, etc. helper types.

This keeps the `...Opts` / `...Filter` types from being a part of the entity's type signature, which is a simplification, and will hopefully give us some more freedom wrt filter/etc typing on subtypes (i.e. the filters of a subtype won't need to be co/contravariant with the filters of the base type).

Granted, we still need the `__type` field itself to do the lookups from "type `T`" --> "some other type", that given these "some other type"S are not directly derived/mapped from `T`.

(I.e. the other approach to this problem, is to rewrite all of the `AuthorOpts` / `AuthorFilter` / etc. types to purely be mapped types against `Author`; this is maybe doable, but codegen-ing out the types has so far being simpler to implement.)

Besides more freedom wrt filtering, the other hope was better compilation speeds, just from the entity types being (slightly) simpler. On our internal codebase, this change makes `tsc` ~6% faster, which isn't a crazy win, but still nice to see. 

Fixes #1295

Before:

```
$ time yarn build
________________________________________________________
Executed in   54.25 secs    fish           external
   usr time   72.73 secs  209.00 micros   72.73 secs
   sys time    2.35 secs   75.00 micros    2.35 secs
```

```
$ time yarn build
________________________________________________________
Executed in   53.43 secs    fish           external
   usr time   69.43 secs  276.00 micros   69.43 secs
   sys time    2.43 secs   99.00 micros    2.43 secs
```

```
$ time yarn build
________________________________________________________
Executed in   54.86 secs    fish           external
   usr time   73.87 secs  195.00 micros   73.87 secs
   sys time    2.49 secs   73.00 micros    2.49 secs
```

---

After

```
 $ time yarn build
________________________________________________________
Executed in   50.98 secs    fish           external
   usr time   66.98 secs  206.00 micros   66.98 secs
   sys time    2.52 secs   75.00 micros    2.52 secs
```

```
$ time yarn build
________________________________________________________
Executed in   51.56 secs    fish           external
   usr time   67.88 secs  192.00 micros   67.88 secs
   sys time    2.48 secs   70.00 micros    2.48 secs
```

```
$ time yarn build
________________________________________________________
Executed in   49.32 secs    fish           external
   usr time   65.03 secs  216.00 micros   65.03 secs
   sys time    2.40 secs   79.00 micros    2.40 secs
```


chr1sjf0x (approved):
praise: Surprisingly straight forward 👍

PR #1311, opened 2024-12-12
feat: Skip defaults on already-set fields.

stephenh:


packages/tests/integration/src/EntityManager.defaults.test.ts
```diff
@@ -35,6 +35,17 @@ describe("EntityManager.defaults", () => {
...
   });
 
+  it("does not overwrite existing sync default set to undefined", async () => {
+    const em = newEntityManager();
+    // Create a new book with explicit notes
+    const b = em.create(Book, {
+      title: "Book 1",
+      notes: undefined,
+    });
+    // Then the synchronous default did not overwrite them
```
chr1sjf0x, line 45: thought: Shouldn't there be a flush in here?

stephenh: Right, the test "knows" the sync default will be applied immediately.

Agreed the async would be good to handle as well; I'd already started covering the behavior of `createOrUpdatePartial` which I believe correctly (albeit maybe unintuitively) still *does* let `field: undefined` drive defaults:

https://github.com/joist-orm/joist-orm/pull/1317

Because the semantics of that `em.createOrUpdatePartial` is that "`undefined`-as-a-value means skip".

PR #1319, opened 2024-12-13
fix: Remove serializable isolation level

arvinsiva:
Removes the `serializable` isolation level since it is being used incorrectly (i.e. both reads and writes need to be run within the same transaction) and causing many false errors.

Read [the PostgeSQL Transaction Isolation Docs](https://www.postgresql.org/docs/current/transaction-iso.html#XACT-SERIALIZABLE) for more info.


packages/tests/integration/src/EntityManager.txns.test.ts
```diff
@@ -1,171 +0,0 @@
-import { Stepper } from "@src/Stepper.test";
-import { Publisher, SmallPublisher } from "@src/entities";
-import { select } from "@src/entities/inserts";
-import { newPgConnectionConfig } from "joist-orm";
-import { Pool } from "pg";
-
-import { knex, newEntityManager } from "@src/testEm";
-
-describe("EntityManager", () => {
```
arvinsiva, line 9: Seems like this entire test file was to test anomalies based on serializable isolation level or not

packages/tests/integration/src/EntityManager.txns.test.ts
```diff
@@ -1,171 +0,0 @@
...
-        await steps.on(1, async () => em.find(Publisher, { name: "foo" }));
-        await steps.on(3, async () => em.create(SmallPublisher, { name: "foo", city: "c1" }));
-      });
-      await steps.on(5, async () => {});
-    })();
-
-    const t2 = (async () => {
-      const em = newEntityManager();
-      await em.transaction(async () => {
-        await steps.on(2, async () => em.find(Publisher, { name: "foo" }));
```
stephenh, line 157: I'm kinda wanting to keep this test file, just to play with/repro anomaly failures in the future, even if we do totally remove the isolationLevel capability from the EM.

Also, I think technically this test shows how users _could_ have used the isolation level feature, by explicitly doing their em.finds/loads within the txn as well.

Granted, we don't do that at all, but apps I used in the past did (it was common before "Internet scale" to always do reads and writes in a txn).

stephenh: (afk dropped some of my comment...)

Maybe we keep this test, but update it to fail, to show we explicitly don't support it atm?

stephenh (approved):

PR #1324, opened 2024-12-19
fix: loosen dependencies

blimmer:
I noticed that `joist` is requesting a few specific versions, which make the dependency tree of consuming projects more complex.

Here's how I noticed it in a project:

```shell
>  yarn why prettier
├─ my-project@workspace:.
│  └─ prettier@npm:3.4.2 (via npm:^3.4.2)
|
├─ @types/prettier@npm:3.0.0
│  └─ prettier@npm:3.4.2 (via npm:*)
│
└─ joist-graphql-codegen@npm:1.215.3
   └─ prettier@npm:3.4.1 (via npm:3.4.1)
```

Joist is requiring _exactly_ prettier version 3.4.1, which makes this tree more complex. We should allow `^3.4.1` so the package manager can more effectively dedupe.

### DIFF
--- a/packages/codegen/package.json
+++ b/packages/codegen/package.json
@@ -20,7 +20,7 @@
     "@dprint/json": "0.15.6",
     "@types/inquirer": "^9.0.7",
     "@types/jscodeshift": "^0.12.0",
-    "@types/pluralize": "0.0.33",
+    "@types/pluralize": "^0.0.33",
     "@types/semver": "^7.5.8",
     "change-case": "^4.1.2",
     "fast-glob": "^3.3.2",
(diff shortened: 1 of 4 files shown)


packages/codegen/package.json
```diff
@@ -20,7 +20,7 @@
     "@dprint/json": "0.15.6",
```
blimmer, line 20: I'm not sure why this was reverted a while back: https://github.com/joist-orm/joist-orm/commit/546f346f2a511e8381545805bc09452888ce50ec

It would be nice to also loosen this requirement if possible.

stephenh (approved):

packages/codegen/package.json
stephenh: Er, yeah really should have put a "why" into that revert--not remembering the issue

PR #1328, opened 2024-12-21
feat: Lazily parse Temporal dates

stephenh:
Previously we let pg keep turning DATE, TIMESTAMP, and TIMESTAMPTZ columns into JS `Date`s, and then would do `Date` -> `Temporal.*` conversions at our serde layer.

Now we tell pg to leave these types as "just strings" and we do all parsing into their Temporal counterparts at our serde layer.

This achieves two goals:

* We don't have to hop through `Date` and loose the microsecond precision of `TIMESTAMPTZ`s
* The serde parsing is already lazy and so we'll avoid parsing dates until the application actually reads them. 

Kinda oddly, the `pg` library only has support for parsing wire-to-JS, and not formatting JS-to-wire, but that's fine, we use the serde layer's `mapToDb` codepath to turn the `Temporal` data back into strings-on-the-wire when doing inserts/updates/wheres.

As a final "kinda cute" update, because we're not loosing the microsecond-level precision, we can remove our `date_trunc` hack from our op locking `WHERE` clauses.


packages/codegen/src/generateMetadataFile.ts
```diff
@@ -110,16 +110,9 @@ function generateFields(config: Config, dbMetadata: EntityDbMetadata): Record<st
       } else {
         serdeType = PrimitiveSerde;
       }
-      if (serdeType === PrimitiveSerde || serdeType === DateSerde) {
```
stephenh, line 113: Given `timeZone` is held in this new global `runtimeConfig`, we don't have to pass it into the individual serde constructors.

packages/codegen/src/generate.ts
```diff
@@ -79,6 +79,10 @@ export async function generateFiles(config: Config, dbMeta: DbMetadata): Promise
   const metadataFile: CodegenFile = {
     name: "./codegen/metadata.ts",
     contents: code`
+      ${setRuntimeConfig}({
+        temporal: ${config.temporal ? { timeZone: typeof config.temporal === "boolean" ? "UTC" : config.temporal.timeZone } : "false"},
+      });
```
stephenh, line 84: Seems likely we're going to want to reference `joist-config.json` data at runtime, so added this embedded-into-metadata.ts call to pass things along into a global `runtimeConfig` const.

packages/orm/src/EntityManager.ts
```diff
@@ -1078,11 +1078,11 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW> {
       const now = new Date();
       if (createdAt) {
         const serde = baseMeta.fields[createdAt].serde as TimestampSerde<unknown>;
-        data[createdAt] = serde.mapFromDate(now);
+        data[createdAt] = serde.mapFromNow(now);
```
stephenh, line 1081: Renamed the `mapFromNow` just to extra-highlight for myself what it's used for.

packages/orm/src/serde.ts
```diff
@@ -119,112 +126,73 @@ export class PrimitiveSerde implements FieldSerde {
...
 
   mapFromJsonAgg(value: any): any {
     if (value === null) return value;
     return new Date(value);
   }
+}
 
-  mapFromDate(value: Date) {
-    return value;
+export class PlainDateSerde extends PrimitiveSerde {
```
stephenh, line 140: Disclaimer these `serde` updates are all ~hacky 1st passes, needs a 2nd pass.

packages/orm/src/temporal.ts
```diff
@@ -0,0 +1,44 @@
+import type { Intl, Temporal, toTemporalInstant } from "temporal-polyfill";
+import { fail } from "./utils";
+
+type RequireTemporal = { Temporal: typeof Temporal; toTemporalInstant: typeof toTemporalInstant; Intl: typeof Intl };
```
stephenh, line 4: Just copy/pasted these here from `utils.ts`, no changes

PR #1331, opened 2024-12-27
feat: Remove DISTINCT usage in em.find queries

stephenh:
This is a fairly massive rewrite of the SQL that `em.find` issues for o2m queries.

Specifically when filtering on children, i.e. `em.find(Author, { books: ...conditions... })`, previously we would `LEFT OUTER JOIN` into whatever `N` child rows, and then use `SELECT DISTINCT` against the `a.id`, to avoid having a singular `authors` row returned multiple times (i.e. once per book):

```sql
SELECT DISTINCT a.* FROM authors as A
  LEFT OUTER JOIN books b ON b.author_id = a.id
  LEFT OUTER JOIN comments c ON c.author_id = a.id
```

This `SELECT DISTINCT` worked, but created performance issues if multiple large child collections were queried, i.e. something like `em.find(Author, { books: ...conditions..., comments: ...conditions... })`.

With this query, if a single Author has `N` books and `M` comments, doing a `LEFT OUTER JOIN` into both of these `books` and `comments` tables would create `N x M` tuples, which Postgres had to hold in memory, to evaluate the "did any of this author's books match ...conditions... and any of this author's comments match ...conditions..." before `DISTINCT`-ing all the tuples back down to a single author.

Granted, for small values of `N` and `M` this was fine, but in production we've had ~3-4 queries have big enough values that it caused ~1-2 second long queries.

To avoid this `N x M` explosion, this PR rewrites the SQL that is generated for the query from `SELECT FROM authors LEFT OUTER JOIN books on b.author_id = a.id LEFT OUTER JOIN comments on c.author_id = a.id` (which causes a combination of rows) to use `CROSS JOIN LATERAL`s:

```sql
SELECT * FROM authors as A
  CROSS JOIN LATERAL (
    SELECT count(*) as _ FROM books b WHERE b.author_id = a.id
  ) as b
  CROSS JOIN LATERAL (
    SELECT count(*) as _ FROM comments c WHERE c.author_id = a.id
  )
```

Where now each `CROSS JOIN LATERAL` is an aggregate query (that is guaranteed to return only 1 row, for all of the author's books, and 1 row for all of the author's comments), such that having multiple o2m clauses in an `em.find`, no longer causes the `N x M` combination of tuples.

Which means we can stop adding the `SELECT DISTINCT` to the queries all together, as we'll never have "more than 1 row per author" to worry about.

---

In addition to just better/more performant queries, the "aggregate all children" approach also gives us an easy way to count the number of children (see the `count(*)` above), so we can add that support to the API like:

```
// find authors with exactly 2 books
em.find(Author, { books: { $count: 2 } });

// find authors with more than 3 books
em.find(Author, { books: { $count: { gt: 3 } } });
```

---

Lateral joins are somewhat esoteric, but actually widely supported outside of Postgres, and also had already been a tool we're using for the (opt-in / disabled-by-default) JSON preloading.

The JSON preloading in particular wanted to use the same "aggregate all of my (books) children into a single (per author) row", so `LATERAL JOIN`s were a great fit.

Given `em.find` itself uses `LATERAL JOIN`s, this PR also re-worked the JSON preloading to use the new/blessed lateral join infra, instead of it's original/proof-of-concept version.

---

Moving the child relations into "aggregate only" lateral join rows also meant that `em.find` batching had to be written from a "final join with a really big `ON` clause" to an "up-front cross join than injects each parameter set as a new row".

This is likely a better query pattern anyway.

### DIFF
--- a/packages/orm/src/AliasAssigner.ts
+++ b/packages/orm/src/AliasAssigner.ts
@@ -8,13 +8,19 @@ export class AliasAssigner {
     this.getAlias = this.getAlias.bind(this);
     // If we're assigning aliases into an existing query, get the current assignments
     if (query) {
-      for (const table of query.tables) {
-        // Ignore CTI base/sub aliases
-        if (table.alias.includes("_")) continue;
-        const abbrev = abbreviation(table.table);
-        const i = this.#aliases[abbrev] || 0;
-        const j = Number(table.alias.replace(abbrev, "")) + 1;
-        this.#aliases[abbrev] = Math.max(i, j);
+      const todo = [query];
+      while (todo.length > 0) {
+        const query = todo.pop()!;
+        for (const table of query.tables) {
+          // Ignore CTI base/sub aliases
+          if (table.alias.includes("_")) continue;
+          if (table.join === "lateral") {
+            this.maybeBumpIndex(table.table, table.alias);
+            todo.push(table.query);
+          } else {
+            this.maybeBumpIndex(table.table, table.alias);
+          }
+        }
       }
     }
   }
@@ -24,5 +30,14 @@ export class AliasAssigner {
     const i = this.#aliases[abbrev] || 0;
     this.#aliases[abbrev] = i + 1;
     return i === 0 ? abbrev : `${abbrev}${i}`;
+  }
+
+  /** Given an existing alias, like `a5`, make sure our internal `a = N` is at least 5 or higher. */
+  private maybeBumpIndex(tableName: string, alias: string): void {
+    const tag = abbreviation(tableName);
+    const ourN = this.#aliases[tag] || 0;
+    // Recover the `5` out of `pi5` by stripping the `pi` prefix
+    const newN = Number(alias.replace(tag, "")) + 1;
+    this.#aliases[tag] = Math.max(ourN, newN);
   }
 }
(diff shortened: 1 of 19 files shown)


packages/orm/src/AliasAssigner.ts
```diff
@@ -8,13 +8,19 @@ export class AliasAssigner {
...
     // If we're assigning aliases into an existing query, get the current assignments
     if (query) {
-      for (const table of query.tables) {
-        // Ignore CTI base/sub aliases
-        if (table.alias.includes("_")) continue;
-        const abbrev = abbreviation(table.table);
-        const i = this.#aliases[abbrev] || 0;
-        const j = Number(table.alias.replace(abbrev, "")) + 1;
-        this.#aliases[abbrev] = Math.max(i, j);
+      const todo = [query];
```
stephenh, line 11: Rewrote this to crawl through nested `ParsedFindQuery`s if necessary (nested `ParsedFindQuery`s is how lateral joins are implemented now).

packages/orm/src/QueryParser.ts
```diff
@@ -61,34 +62,73 @@ export interface PrimaryTable {
...
+type ParsedSelectWithBindings = { sql: string; bindings: any[]; aliases: string[] };
+
 /** The result of parsing an `em.find` filter. */
 export interface ParsedFindQuery {
-  selects: string[];
+  selects: ParsedSelect[];
   /** The primary table plus any joins. */
   tables: ParsedTable[];
-  /** Any cross lateral joins, where the `joins: string[]` has the full join as raw SQL; currently only for preloading. */
-  lateralJoins?: { joins: string[]; bindings: any[] };
```
stephenh, line 87: This is the original `LATERAL JOIN` infra from the JSON preloader, which is getting replaced with `tables` having `LateralJoinTable`s in them.

Issue #1338, opened 2025-01-08
Queries with multiple o2m joins are expensive


stephenh:
Joist's `em.find` API makes it easy to "filter by children", i.e. something like:

```
em.find(Author, { books: { title: "b1" } });
```

Will find `Author` have that "at least one" book with a title of `b1`.

Currently, Joist implements this by `LEFT JOIN`-ing into `books` and then `DISTINCT`-ing on the `a.id`.

This means we only return 1 `author` row over the wire (good), but only after asking Postgres to dedupe the author+books rows.

Usually this is fine, until we filter on two o2m collections:

```
em.find(Author, { books: { title: "b1" }, comments: { text: "c1" } });
```

This is saying find `Author`s that have 1 book with "title=b1" and 1 comment with "text=c1".

However, when we `LEFT JOIN books` and `LEFT JOIN comments`, postgres creates 1 tuple per `author x book x comment` combination.

When there many children books/comments, this can create a very large result set that postgres has to create in-memory -- only to `DISTINCT` it away to return "just the author".

Ideally we'd like evaluate this query by still joining into `books` and `comments`, but having those joins produce **only a single row**, effectively a single roll up/aggregate row that presents "all of this author's books" and "all of this author's comments", which will prevent the combinatorial explosion and mean we don't have to `DISTINCT` anymore.

stephenh:
One option is to realize these two conditions should be pushed down together, i.e. something like:

```sql
CROSS JOIN LATERAL (
  SELECT count(*) as _,
  BOOL_OR(b.title LIKE $1 AND b.title LIKE $2) as _combined
  FROM books ...
) AS b
WHERE b._combined
```

However, it's not clear how this would work in more complex conditions like:

```ts
em.find(Author, { as: a, books: b }, conditions: {
  and: [
    b.title.like("fo%"),
    { or: [b.title.like("%ar"), a.firstName.eq("a1") ] },
  ]
});
```

I.e. the two `b.title` references are in separate clauses.

---

We also need to handle queries across levels, i.e.:

```ts
em.find(Author, { books: { as: b, reviews: r } }, conditions: {
  or: [
    { and: [b.title.eq("t1"), br.rating.eq(3)] },
    { and: [b.title.eq("t2"), br.rating.eq(4) ] },
  ]
});
```

Where the tuples of:

* `[book1, [book1's ratings]`
* `[book2, [book2's ratings]`
* `[book3, [book3's ratings]`

[...]

PR #1339, opened 2025-01-08
feat: Revert using lateral joins for o2m/m2m finds

stephenh:
We ran into issues rolling this out internally -- despite our entire / non-trivial / usually-a-pedantic-pita test suite passing, we didn't caught that complex conditions against child/o2m relations had a breaking change in semantics.

So reverting back to the `OUTER JOIN`s + `DISTINCT` approach for now.

See #1338 for more on the issue + ideas to fix it.


stephenh:
:cry:

PR #1358, opened 2025-02-07
feat: Add notNull config support for CTI based subclasses

zgavin:
Also renamed `stiNotNull` to just `notNull`


docs/docs/advanced/class-table-inheritance.md
```diff
@@ -112,6 +112,21 @@ After this, Joist will enforce that all `Animal`s must be either `Dog`s or `Cat`
...
 
+## SubType Configuration
+
+In some situations a subtype may want to override the behavior of a field or relation it inherits from its base type.  In such situations you may manually configure the `joist-config.json` for the subtype to give Joist hints about "which subtype" a given relation should be or about its nullability.
+
+For example, instead of the `Dog.breed` relation (from the `animals.breed_id` FK) being typed as `Breed`, you want it to be typed as `DogBreed` because you know a `Dog` can't be a Maine Coon.
+
+These hints in `joist-config.json` generally look like:
+
+1. Adding `subType: "DogBreed"` to any relation (like `breed`) from the `animals` table that should be limited to a specific subtype in `Cat` or `Dog`
```
stephenh, line 123: Maybe say this is "within the `Dog` section" of the config file.

packages/codegen/src/inheritance.ts
```diff
@@ -29,27 +29,37 @@ export function applyInheritanceUpdates(config: Config, db: DbMetadata): void {
...
     if (entity.baseType) {
+      const entityConfig = config.entities[entity.name];
+      const baseConfig = config.entities[entity.baseType.name];
       // Look through the base's m2os & o2ms, looking for a config specialization
       for (const rel of [...entity.baseType.manyToOnes, ...entity.baseType.oneToManys]) {
+        const notNull =
+          entityConfig?.relations?.[rel.fieldName]?.notNull ??
+          baseConfig?.relations?.[rel.fieldName]?.notNull ??
+          // Probe the otherFieldName for a `subType: "self"` on self-referential relations in STI tables
+          baseConfig?.relations?.[rel.otherFieldName]?.notNull;
```
stephenh, line 40: Maybe not need this one? :thinking:

packages/codegen/src/inheritance.ts
```diff
@@ -125,7 +135,7 @@ function expandSingleTableInheritance(
         // Make fields as required
         function maybeRequired<T extends { notNull: boolean; fieldName: string }>(field: T): T {
           const config = subTypeFields.find(([name]) => name === field.fieldName)?.[1]!;
-          if (config.stiNotNull) field.notNull = true;
+          if (config.notNull) field.notNull = true;
```
stephenh, line 138: Kinda wonder if this `maybeRequired` could go away, now that `setupSubTypeSpecialization` is doing it as well?

packages/tests/integration/src/entities/codegen/LargePublisherCodegen.ts
```diff
@@ -104,13 +115,17 @@ export interface LargePublisherGraphQLFilter extends PublisherGraphQLFilter {
...
+  spotlightAuthor?: AuthorOrder;
 }
 
 export interface LargePublisherFactoryExtras {
 }
 
 export const largePublisherConfig = new ConfigApi<LargePublisher, Context>();
 
+largePublisherConfig.addRule(newRequiredRule("spotlightAuthor"));
+largePublisherConfig.addRule("spotlightAuthor", mustBeSubType("spotlightAuthor"));
```
stephenh, line 127: Little surprised the `mustBeSubType` rule showed up, given `Author` doesn't have any subtypes

packages/tests/integration/src/EntityManager.reactiveRules.test.tsx
```diff
@@ -194,7 +194,7 @@ describe("EntityManager.reactiveRules", () => {
 
   it.withCtx("runs rule added via afterMetadata", async ({ em }) => {
     // Given a publisher
-    const p = newPublisher(em);
+    const p = newPublisher(em, { authors: [{}] });
```
stephenh, line 197: Maybe could have added `authors: [{}]` as a factory default?

stephenh (approved):

packages/tests/integration/src/EntityManager.reactiveRules.test.tsx
zgavin: that actually broke more tests than just manually inserting it in a dozen or so places 🤷

packages/tests/integration/src/entities/codegen/LargePublisherCodegen.ts
zgavin: hmm yeah, why is this not breaking any tests? 🤔

stephenh: I think `mustBeSubType` just gets the "what should I `instanceof` against?" from the runtime metadata, so it's probably just adding a "passes but technically unnecessary" `spotlightAuthor instanceof Author` check. We could probably just leave it.

packages/codegen/src/inheritance.ts
zgavin: Thinking more about it, I think that `notNull` should only be read from `entityConfig` and not fall back to `baseConfig` inside `setupSubTypeSpecialization`.  Since STI descendants don't have per type config, it'd never be read from their config so they'd effectively be skipped here.  Meanwhile, it doesn't make sense for CTI to define a `notNull: true` on the base config since they could just make the db level field required in the first place.

packages/codegen/config-json-schema.json
```diff
@@ -47,7 +47,10 @@
                     "superstruct": { "type": "string" },
                     "zodSchema": { "type": "string" },
                     "type": { "type": "string" },
-                    "serde": { "type": "string" }
+                    "serde": { "type": "string" },
+                    "stiType": { "type":  "string" },
+                    "notNull": { "type": "boolean" },
+                    "stiDiscriminator": { "type": "object" }
```
stephenh, line 53: Thanks for updating this! ...now I have to remember how to upload it to the schema registry in the sky that webstorm/vscode use...

packages/codegen/src/inheritance.ts
stephenh: > I think that notNull should only be read from entityConfig and not fall back to baseConfig inside setupSubTypeSpecialization

Makes sense

> Since STI descendants don't have per type config

I thought we wanted that / were thinking of adding it?

stephenh (approved):

packages/codegen/src/inheritance.ts
zgavin: > I thought we wanted that / were thinking of adding it?

I mean maybe at some point, but right now we don't have it and we don't need it immediately, so we can punt on this until then?

PR #1372, opened 2025-02-19
fix: use the correct meta for inherited recursive relations

zgavin:


packages/orm/src/dataloaders/recursiveChildrenDataLoader.ts
```diff
@@ -21,7 +22,10 @@ export function recursiveChildrenDataLoader<T extends Entity, U extends Entity>(
   em: EntityManager,
   collection: RecursiveChildrenCollectionImpl<T, U>,
 ): DataLoader<Entity, U[]> {
-  const { meta, fieldName } = collection;
+  let { meta, fieldName } = collection;
+  // This could be called from subtypes to get relations defined on the parent. So we need to make sure we are using the
+  // correct meta by walking the inheritance tree until we find the meta that actually has the root o2m field
+  while (!(collection.o2mFieldName in meta.fields) && meta.baseType) meta = getMetadataForType(meta.baseType);
```
stephenh, line 28: Huh, kinda surprised more places haven't needed to do this -- makes me wonder if `collection.meta` itself should be "the maybe-base type that owns the relation" :thinking: 

Or if that would break more than it's worth, given it seems like a lot of places are just fine with the meta always being the child.

stephenh (approved):

PR #1373, opened 2025-02-19
feat: Convert to postgres.js

### LINKED ISSUE #1360: Replace Knex with postgres.js
Joist's internal implementation was originally leveraged/heavily coupled to knex, as in the very early days of Joist, before `em.find` existed, and the first implementation of `em.find`, knex's query builder abstraction was useful to have.

However, as Joist's internals have evolved, we use knex less & less as "a query builder" and more as "a client directly to the database".

Which means it's been getting more & more possible to swap/replace Knex, but we've just never had a reason.

I think there is a good reason now, which is that Bun's new `sql` module is API compatible with the node npm package `postgres`, which is a lower-level "just a database client", similar to/a replacement for node-pg.

Pros/Cons:

1. Pro: Having our primary `PostgresDriver` use `postgres.js` would make Bun `sql` support trivial -- same APIs
 
2. Pro: `postgresql.js` is lower-level than knex, and we don't need a wrapper/abstraction layer around the database client

3. Pro: `postgresql.js` is "the new kid on the block" and supposedly faster than `node-pg` (I can attach some perf results) and definitely has a more active maintainer

4. Con: `node-pg` is rock solid/years of industry adoption, we could validly move to using `node-pg` directly (instead of `postgres.js`), albeit we would loose the instant/for-free Bun adoption

5. Con: We can't share connection pools between knex and `postgres.js` -- it's extremely unlikely our (and others) apps would migrate to knex overnight, and so we'd have a dual knex/postgres.js codebase for awhile. This would likely not be horrible, other than they don't really talk to each other, particularly the connection pools, so we'd have two pools per machine.

   At our scale, this is fine/np, but it's also not really best practice.

stephenh:
Fixes #1360 

Neat things:

* postgres.js instead of knex/node-pg
  * Prepared statements (leverages unnest i.e. fixed param sizes -> better hit rates)
  * Pipelining in `em.flush`

---

Pretty neat, postgres.js sends the string of `INSERT INTO authors (a) VALUES (%1)` to pg as part of `Prepare`, and then pg acks back "the oid of %1 is ...", and then postgres.js will use that info to serialize parameters back out to pg.


stephenh:
This is using the `sql.unsafe` API, which doesn't use prepared statements:

```
[I] ~/o/j/p/t/integration (postgresjs|✔) $ yarn env-cmd tsx ./src/benchmarks/loading-authors.ts
cpu: Intel(R) Core(TM) i9-10885H CPU @ 2.40GHz
runtime: node v23.1.0 (x64-linux)

benchmark        time (avg)             (min … max)       p75       p99      p999
--------------------------------------------------- -----------------------------
• loading 50k authors
--------------------------------------------------- -----------------------------
postgres.js  93'509 ns/iter    (77'829 ns … 400 µs) 97'556 ns    147 µs    330 µs
knex            179 µs/iter       (162 µs … 805 µs)    175 µs    303 µs    650 µs
em.find         395 µs/iter     (348 µs … 1'986 µs)    399 µs    675 µs  1'588 µs

summary for loading 50k authors
  postgres.js
   1.92x faster than knex
   4.22x faster than em.find

• loading 50k books
--------------------------------------------------- -----------------------------
postgres.js  83'321 ns/iter    (74'792 ns … 403 µs) 85'263 ns    110 µs    167 µs
knex            136 µs/iter       (120 µs … 391 µs)    136 µs    207 µs    337 µs
em.find         341 µs/iter     (285 µs … 1'495 µs)    365 µs    523 µs  1'482 µs

summary for loading 50k books
  postgres.js
   1.63x faster than knex
   4.09x faster than em.find
```

And this is changing `PostgresDriver.executeQuery` to use the primary tagged literal API:

```
[I] ~/o/j/p/t/integration (postgresjs|✚1) $ yarn env-cmd tsx ./src/benchmarks/loading-authors.ts
cpu: Intel(R) Core(TM) i9-10885H CPU @ 2.40GHz
runtime: node v23.1.0 (x64-linux)

benchmark        time (avg)             (min … max)       p75       p99      p999
--------------------------------------------------- -----------------------------
• loading 50k authors
--------------------------------------------------- -----------------------------
postgres.js     106 µs/iter    (96'158 ns … 409 µs)    105 µs    158 µs    329 µs
knex            215 µs/iter       (183 µs … 865 µs)    224 µs    422 µs    764 µs
em.find         288 µs/iter     (243 µs … 2'151 µs)    285 µs    645 µs  2'017 µs

summary for loading 50k authors
  postgres.js
   2.03x faster than knex
   2.72x faster than em.find

• loading 50k books
--------------------------------------------------- -----------------------------
postgres.js  82'430 ns/iter    (73'081 ns … 950 µs) 86'695 ns    114 µs    375 µs
knex            136 µs/iter       (117 µs … 379 µs)    142 µs    213 µs    365 µs
em.find         200 µs/iter     (179 µs … 1'430 µs)    199 µs    298 µs  1'377 µs

summary for loading 50k books
  postgres.js
   1.65x faster than knex
   2.42x faster than em.find
```

PR #1376, opened 2025-02-19
fix: findOrCreate should reject collections

### LINKED ISSUE #1375: em.findOrCreate should reject collections

stephenh:
findOrCreate fundamentally assumes it can "eval the `where` clause against all in-memory entities", so that we don't dup against a just-created entity.

Currently we don't assume that entity has non-primitive relations loaded into memory, and even if we did `populate` those into memory, the concept of "this entity is unique based on (...)" really is meant to match "basically what a database constraint" can eval, i.e. just fields/columns directly owned on the entity.

Fixes #1375

### DIFF
--- a/packages/orm/src/dataloaders/findOrCreateDataLoader.ts
+++ b/packages/orm/src/dataloaders/findOrCreateDataLoader.ts
@@ -6,7 +6,7 @@ import { ManyToOneReference, PolymorphicReference, isLoadedReference } from "../
 import { OptsOf } from "../typeMap";
 import { whereFilterHash } from "./findDataLoader";
 
-interface Key<T extends Entity> {
+interface FindOrCreateKey<T extends Entity> {
   where: Partial<OptsOf<T>>;
   ifNew: OptsOf<T>;
   upsert: Partial<OptsOf<T>> | undefined;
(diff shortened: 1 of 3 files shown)


packages/orm/src/dataloaders/findOrCreateDataLoader.ts
```diff
@@ -17,27 +17,44 @@ export function findOrCreateDataLoader<T extends Entity>(
...
   // to avoid creating duplicates. Also use `whereFilterHash` b/c if a new entity is included in `where`,
   // it will use `entity.toString()` to keep it unique from other new entities.
   const whereValue = whereFilterHash(!hasAnyCitext ? (where as any) : (maybeLower(meta, where) as any));
   const batchKey = `${type.name}-${whereValue}-${softDeletes}`;
 
-  return em.getLoader<Key<T>, T>(
+  const invalidKeys = Object.keys(where).filter((key) => {
+    const field = meta.allFields[key];
+    const supported =
+      field.kind === "primitive" || field.kind === "enum" || field.kind === "m2o" || field.kind === "poly";
```
chr1sjf0x, line 47: thought: Could be nice to make an array of these valid keys and use `validKeys.includes(field.kind)` here.

Then the error message could also use the validKeys to build the message.

Granted, it is not likely the validKeys will be changing much in the future 🤷

chr1sjf0x (approved):
TY @stephenh

PR #1399, opened 2025-03-11
fix: skip entityResolver optimization for references to a inheritance based type

zgavin:


bbarbosa-homebound (approved):

stephenh (approved):
Makes sense--would be great to have a test case so I don't regress this in the future, but the in-source comment is good too.

PR #1406, opened 2025-03-19
feat: add includeSchema option to config, with default of public

### LINKED ISSUE #1398: codegen: optionally define schemas to derive metadata from
Long time, no issue. :)

We're evaluating using pg schemas for some lightweight tenancy. This isn't external/per organsiation, but really for creating various ops - such as a "demo" tenant.

The idea is to have a couple schemas say public, demo etc. We have some global tables, which we are considering moving out to a `core` schema.

We'll then be able to set the pg search_path to `{tenant}, core` per request.

Within codegen there's currently https://github.com/joist-orm/joist-orm/blob/1a3fbdee64cfcf9ffafc06bf048c6d6a047891a1/packages/codegen/src/utils.ts#L197-L199 - if this could be optionally configurable within joist-config.json to an array of schemas, we would be able to define `['public', 'core']`.

Afaik nothing else from joist would be needed here, as search_paths allows this to be handled at pg level. Happy to PR if open to this!

brudil:
- Fixes #1398
- Add optional `"includeSchema"` option to `joist-config.json`
- Taking an array of schema names

### DIFF
--- a/packages/codegen/src/config.ts
+++ b/packages/codegen/src/config.ts
@@ -142,7 +142,7 @@ export const config = z
     version: z.string().default("0.0.0"),
 
     /** A list of postgres schemas to use for finding tables */
-    includeSchema: z.optional(z.array(z.string())),
+    schemas: z.optional(z.array(z.string())),
   })
   .strict();
 
--- a/packages/codegen/src/config.ts
+++ b/packages/codegen/src/config.ts
@@ -142,7 +142,7 @@ export const config = z
     version: z.string().default("0.0.0"),
 
     /** A list of postgres schemas to use for finding tables */
-    includeSchema: z.array(z.string()).default(["public"]),
+    includeSchema: z.optional(z.array(z.string())),
   })
   .strict();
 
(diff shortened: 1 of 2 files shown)


packages/codegen/src/config.ts
```diff
@@ -140,6 +140,9 @@ export const config = z
     allowImportingTsExtensions: z.optional(z.boolean()),
     // The version of Joist that generated this config.
     version: z.string().default("0.0.0"),
+
+    /** A list of postgres schemas to use for finding tables */
+    includeSchema: z.optional(z.array(z.string())),
```
stephenh, line 145: @brudil wdyt about calling this `schemas`?

brudil: No quarms here - was torn on a hundred possibilities.

brudil: Updated in [a001e36](https://github.com/joist-orm/joist-orm/pull/1406/commits/a001e36b6dc67241ac04b4732308136946d62d6c)

stephenh (approved):
Looks great, thank you!

PR #1419, opened 2025-04-07
fix: Add isLoaded caching

### LINKED ISSUE #1166: Optimize isLoaded checks
Joist's fundamental entity-based ORM approach allows incrementally loading subgraphs of the database into memory.

With this approach, we often need to check "has this relation (m2o/o2m/m2m) been loaded into memory yet?".

These loaded checks for "base relations" like m2o are pretty simple, i.e. boolean flags of "did I load the `select * from books from author_id = me` yet?", but as our relations have grown in complexity, i.e. are composite relations sitting on top of other base relations, the isLoaded check has become more dynamic.

I.e. for a composite relation (a reactive field with a load hint of `{ ...json hint... }`), we crawl the load hint and ask "are all of the base relations loaded?".

This is not "slow", but when done "a lot" the expense can add up.

We could probably cache this information, with a scheme like:

* For each composite relation, start off with `loaded=false`
* On the first `isLoaded` calc, do the real/deep isLoaded check
* Store/cache either `loaded=true/false` (new behavior, we don't cache this today)
* Also store the "what graph mutations would invalidate this loaded check"
* Instrument our writes (i.e. `m2o.set`, `m2m.add/remove`) to invalidate any cached checks

We could do the invalidations one of three ways:

1. Fine-grained / instance-based, i.e. this composite relation of `a1.favoriteBooks` has crawled `b1.author` and `b2.author` and `b3.author`, and should be invalidate whenever those specific `b1.author`/etc fields change, or
   * Pro: More exact (than alternative 2)
   * Con: More bookkeeping and complexity to implement 
2. Course-grained, i.e. this composite relation of `Author.favoriteBooks` has crawled `Book.author` and should invalidate whenever any `Book.author` changes
   * Pro: Probably simpler to implement?
   * Con: Would over-invalidate cache, i.e. if two `a1` and `a2` author had their `favoriteBooks` cached as loaded, and only one of `a2`'s books was mutated, then both `favoriteBook` relations would have their cache invalidated
3. Super-dumb, just invalidate every composite relation on any graph mutation
   * Pro: Super-simple
   * Pro: Solves the read case, i.e. queries never mutate, so great/solved caching for them
   * Pro: Writes probably go in phases of "do some mutations" / "then do some recalcs" / "then do some mutations" / "then do some recalcs", so this dumb/bulk invalidation probably wouldn't be that bad

...alternatively/disclaimer, ^ seems worthwhile in general, but also the current hot spot might be resolved with something as simple as caching the "reactive field-level hint" -> "load relation-level hint", or some other missed/unknown hot spot.

stephenh:
Fixes #1166

### DIFF
--- a/packages/orm/src/EntityManager.ts
+++ b/packages/orm/src/EntityManager.ts
@@ -1921,7 +1921,7 @@ async function validateReactiveRules(
           e instanceof rule.source &&
           (e.isNewEntity ||
             e.isDeletedEntity ||
-            ((e as any).changes as Changes<any>).fields.some((f) => rule.fields.includes(f))),
+            ((e as any).changes as Changes<any>).fieldsWithoutRelations.some((f) => rule.fields.includes(f))),
       );
       // From these "triggered" entities, queue the "found"/owner entity to rerun this rule
       return followAndQueue(triggered, rule);
--- a/packages/orm/src/EntityManager.ts
+++ b/packages/orm/src/EntityManager.ts
@@ -310,9 +310,9 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
       outputIsLoadedStats() {
         const entries = [...em.#isLoadedStats.entries()];
         entries.sort((a, b) => b[1].count - a[1].count);
-        for (const [key, stats] of entries) {
-          console.log(`${key}.isLoaded = ${stats.count}`);
-        }
+        // for (const [key, stats] of entries) {
+        //   console.log(`${key}.isLoaded = ${stats.count}`);
+        // }
       },
     };
   }
--- a/packages/orm/src/EntityManager.ts
+++ b/packages/orm/src/EntityManager.ts
@@ -5,6 +5,7 @@ import { getField, setField } from "./fields";
 // We alias `Entity => EntityW` to denote "Entity wide" i.e. the non-narrowed Entity
 import { Entity, Entity as EntityW, IdType, isEntity } from "./Entity";
 import { FlushLock } from "./FlushLock";
+import { IsLoadedCache } from "./IsLoadedCache";
 import { JoinRows } from "./JoinRows";
 import { ReactionsManager } from "./ReactionsManager";
 import { JoinRowTodo, Todo, combineJoinRows, createTodos } from "./Todo";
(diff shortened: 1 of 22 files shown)


packages/orm/src/relations/RecursiveCollection.ts
```diff
@@ -250,6 +253,7 @@ export class RecursiveChildrenCollectionImpl<T extends Entity, U extends Entity>
   }
 
   get isLoaded(): boolean {
+    // We could cache this, and add staleness tracking
```
chr1sjf0x, line 256: Curious: Are these `we could cache this` comments in other files all going to be a follow up?

chr1sjf0x (approved):
TY @stephenh

packages/orm/src/relations/RecursiveCollection.ts
stephenh: Most will be follow-ups yes, but I went ahead and did the two in this Recursive parent/child.

packages/orm/src/EntityManager.ts
```diff
@@ -1884,7 +1898,7 @@ async function validateReactiveRules(
           e instanceof rule.source &&
           (e.isNewEntity ||
             e.isDeletedEntity ||
-            ((e as any).changes as Changes<any>).fields.some((f) => rule.fields.includes(f))),
+            ((e as any).changes as Changes<any>).fieldsWithoutRelations.some((f) => rule.fields.includes(f))),
```
stephenh, line 1901: This ended up being another hot-spot b/c calculating "changed fields with m2o and m2m included" requires diffing the m2o/m2m, which is expensive, particularly the `entityCompare` sorter.

packages/orm/src/EntityManager.ts
```diff
@@ -252,16 +254,25 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
...
     // Expose some of our private fields as the EntityManagerInternalApi
     const em = this;
+
     this.__api = {
       preloader: this.#preloader,
+      pendingChildren: this.#pendingChildren,
+      mutatedCollections: new Set(),
+      hooks: this.#hooks,
+      rm: this.#rm,
+      isLoadedCache: this.#isLoadedCache,
```
stephenh, line 264: Reordered the fielsd here, only new addition is this `isLoadedCache`

PR #1425, opened 2025-04-10
fix: support poly relations to a subclass in query parser

zgavin:
I'm gonna be honest, i dont have the motivation to create a test case for this 😅


packages/orm/src/QueryParser.ts
```diff
@@ -257,7 +257,10 @@ export function parseFindQuery(
               // Or together `parent_book_id in (1,2,3) OR parent_author_id IN (4,5,6)`
               // ...if there is a `parent IN [b:1, b:2, a:1, null]` we'd need to pull the `null` out and do an `OR (all columns are null)`...
               const conditions = Object.entries(idsByConstructor).map(([cstrName, ids]) => {
-                const column = field.serde.columns.find((c) => c.otherMetadata().cstr.name === cstrName)!;
+                const column = field.serde.columns.find(
+                  // tagged ids from subclasses always map to the base class, so we should compare to the base class if we don't directly match
+                  (c) => c.otherMetadata().cstr.name === cstrName || c.otherMetadata().baseType === cstrName,
+                )!;
```
stephenh, line 263: Believe me, I get the fatigue of "lol another thing to test in our poor domain model" :sweat_smile: , but we happen to have an existing FK to polys that are subtyped.

Both of these lines fail around here (the 2nd one this `in` handling, the 1st a little earlier but with the same "find component" logic):

```
  it("foo", async () => {
    await em.find(User, { favoritePublisher: "p:1" });
    await em.find(User, { favoritePublisher: ["p:1"] });
  });
```

So should be a quick add

stephenh (approved):

PR #1437, opened 2025-04-14
feat: add support for subtypes to maybeNewPoly

zgavin:


packages/orm/src/newTestInstance.ts
```diff
@@ -311,21 +311,26 @@ function resolveFactoryOpt<T extends Entity>(
   } else {
     // Look for an obvious default
     if (opt === undefined || opt instanceof MaybeNew) {
-      if (field.kind !== "poly") {
-        const [existing, loggerKey] = getObviousDefault(em, meta, use);
+      if (
+        field.kind === "poly" ||
+        (field.otherMetadata().subTypes.length > 0 && opt instanceof MaybeNew && opt.polyRefPreferredOrder.length > 0)
```
stephenh, line 316: Just so I'm following, this is letting callers use the "find one of these types" feature of `maybeNewPoly`, but with non-poly subtypes?

I.e.

```
maybeNewPoly(LargePublisher, { existingSearchOrder: [SmallPublisher, LargePublisher] });
```

I think that makes sense--wondering if we should have a `maybeNewSubType` alias, since there isn't necessarily a poly involved -- or maybe rename `maybeNewPoly` to something that both polys and subtypes could use.

...maybe we just allow `maybeNew` itself be that, i.e. right now we already need a generic to `maybeNew` like `maybeNew<Publisher>({})`, to get type checking working, and if we changed that to `maybeNew(Publisher)`:

* `maybeNew(Publisher)`
* `maybeNew(Publisher, { name: "" })`
* `maybeNew(LargePublisher, { largeAttribute: "" }, { searchOrder: [SmallPublisher, LargePublisher] })`

It's a breaking change to `maybeNew`, but wdyt?

I could try ^ + add tests if I'm following the ask.

zgavin: yeah, was actually thinking along the same lines that we should just get rid of `maybeNewPoly` and require an explicit type for `maybeNew`. `maybeNewPoly` itself is just a wrapper around `new MaybeNew` anyways.

stephenh (approved):

PR #1439, opened 2025-04-14
fix: update handling of polymorphic references so subtypes aren't thrown out

zgavin:
this is only a partial fix, as passing an ID will still get thrown out


stephenh:
Landed in https://github.com/joist-orm/joist-orm/pull/1982

PR #1443, opened 2025-04-15
fix: Fix setting over ReactiveReferences.

### LINKED ISSUE #1442: em.create setting over reactive fields, references should fail

stephenh:
Calling `em.create(Author, { favoriteBook: b1 })` would work because invoking `setOpts` from the constructor meant that `a.favoriteBook` was not initialized yet, so the `current instanceof ...` logic in `setOpts` was not kicking in.

Now, we have `em.create` not call `setOpts` until after the instance is fully initialized.

We don't go all the way to hard deprecating the `new Author(em, ...)` syntax, but we should plan on doing so.

Fixes #1442.


stephenh:
This has since been fixed.

PR #1446, opened 2025-04-17
fix: allow HookFn return value

allangsilva:
Allow hooks like `beforeCreate`/`beforeUpdate`/etc to receive a returnable function, so it can save the `async/await` when doing something else rather than setting a default value.

Before:
```js
config.beforeCreate(async (author) => {
  await maybePublishAuthorBooks(author);
});
```

After:
```js
config.beforeCreate((author) => maybePublishAuthorBooks(author));
```

### DIFF
--- a/packages/orm/src/config.ts
+++ b/packages/orm/src/config.ts
@@ -28,7 +28,7 @@ export type EntityHook =
   | "afterValidation"
   | "beforeCommit"
   | "afterCommit";
-type HookFn<T extends Entity, C> = (entity: T, ctx: C) => MaybePromise<void>;
+type HookFn<T extends Entity, C> = (entity: T, ctx: C) => MaybePromise<unknown>;
 
 export const constraintNameToValidationError: Record<string, string> = {};
 
--- a/packages/tests/integration/src/entities/Author.ts
+++ b/packages/tests/integration/src/entities/Author.ts
@@ -414,7 +414,8 @@ config.beforeCreate((author) => {
 });
 
 config.beforeCreate((author) => {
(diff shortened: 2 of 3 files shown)


packages/tests/integration/src/entities/Author.ts
```diff
@@ -409,7 +410,11 @@ config.beforeFlush(async (author, { em }) => {
 });
 
 config.beforeCreate((author) => {
-  author.transientFields.beforeCreateRan = true;
+  author.transientFields.beforeCreateRan = true
+});
+
+config.beforeCreate((author) => {
+  return Promise.resolve(author.transientFields.beforeCreateAsyncRan = true);
```
stephenh, line 417: I wonder if not 1-lining this would be simpler, i.e.:

```
config.beforeCreate((author) => {
  author.transientFields.beforeCreateAsyncRan = true;
  return Promise.resolve("testing not void");
});
```

Would I think cover the same use case, and maybe be a little cleaner?

allangsilva: ooh yeah, it looks better. Thanks 👏

stephenh (approved):

packages/orm/src/config.ts
```diff
@@ -28,7 +28,7 @@ export type EntityHook =
   | "afterValidation"
   | "beforeCommit"
   | "afterCommit";
-type HookFn<T extends Entity, C> = (entity: T, ctx: C) => MaybePromise<void>;
+type HookFn<T extends Entity, C> = (entity: T, ctx: C) => MaybePromise<unknown>;
```
zgavin, line 31: suggestion: `MaybePromise<any>` should let you avoid needing the explicit return

PR #1457, opened 2025-04-23
feat: Allow clone to skip setting fields.

stephenh:


packages/orm/src/EntityManager.ts
```diff
@@ -711,9 +712,14 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
    * // This will duplicate the author, but skip any book where the title includes `sea`
    * const duplicatedAuthor = await em.clone(author, { skipIf: (original) => original.title?.includes("sea") })
    */
-  public async clone<T extends EntityW, H extends LoadHint<T>>(
+  public async clone<T extends Entity, H extends LoadHint<T>, const K = keyof RelationsIn<T>>(
```
stephenh, line 715: I think the best version of this `skip` parameter would support deep things, i.e. "clone books and reviews but leave the review.order unset" (or something like that) with a `skip: { books: { reviews: "order" } }` sort of thing, but going to skip that for now.

haislip (approved):

PR #1494, opened 2025-05-13
issue: derived value not pulling from async derived field

stephenh:


stephenh:
No longer reproduces, we fixed in https://github.com/joist-orm/joist-orm/pull/1579

Issue #1546, opened 2025-07-17
Add em.fork


stephenh:
* Return a new `em` with a copy of all currently-loaded entities
* Already-loaded relations should maintain their loaded state
  * I.e. if `author.populate({ books: "reviews" })` was run in `em`, the forked em should see `author.books.isLoaded`, `author.books[0].reviews.isLoaded`, etc.

---

The relationship loaded would probably "just work" if we leverage the preloading cache:

```ts
  private getPreloaded(): U[] | undefined {
    if (this.entity.isNewEntity) return undefined;
    return getEmInternalApi(this.entity.em).getPreloadedRelation<U>(this.entity.idTagged, this.fieldName);
  }
```

I.e. if the parent EM was able to shove/encode all loaded relations into the `a:123.books => [b:1, b:2]` cache, then whenever `a.books.isLoaded` runs, it would just find it.

PR #1550, opened 2025-07-25
feat: add em.fork

zgavin:
Implements #1546

### DIFF
--- a/packages/orm/src/EntityManager.ts
+++ b/packages/orm/src/EntityManager.ts
@@ -1907,9 +1907,10 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
     if (pendingEntities.length > 0) {
       fail("Cannot fork an EntityManager with pending changes");
     }
+    // copy the context so that it's distinct between the two ems and so we can update any references from the old em
+    // to the new one later
     const ctx: C = { ...oldEm.ctx };
     const newEm = new EntityManager<C, Entity, TX>(ctx, { em: oldEm });
-    if ("em" in (ctx as any)) Object.assign(ctx as any, { em: newEm });
     // hydrate (ie, create) each entity from the oldEm in the newEm
     oldEm.#entitiesByTag.forEach((entities, tag) => {
       // tags are always defined by the base constructor, so we use that as hydrate expects the base class as its argument
@@ -1948,14 +1949,16 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
         loadedRelations.forEach((relation) => (newEntity as any)[relation].preload());
       });
     });
-    // If there are any entities in the context, then they are from oldEm and need to be replaced with entities from newEm
+    // Inspect the top level keys in the context and replace any references from oldEm with newEm while also replacing
+    // any entities from oldEm with their version in newEm
     const _ctx = ctx as Record<string, any>;
-    Object.entries(_ctx)
-      .filter(([, value]) => isEntity(value))
-      .forEach(([key, oldEntity]: [string, Entity]) => {
-        _ctx[key] = newEm.#entitiesById.get(oldEntity.idTagged)!;
-      });
-
+    Object.entries(_ctx).forEach(([key, value]) => {
+      if (isEntity(value) && value.em === oldEm) {
+        _ctx[key] = newEm.#entitiesById.get(value.idTagged)!;
+      } else if (value === oldEm) {
+        _ctx[key] = newEm;
+      }
+    });
     return newEm;
   }
 }
--- a/packages/orm/src/EntityManager.ts
+++ b/packages/orm/src/EntityManager.ts
@@ -6,7 +6,7 @@ import { IndexManager } from "./IndexManager";
 // We alias `Entity => EntityW` to denote "Entity wide" i.e. the non-narrowed Entity
 import { getReactiveRules } from "./caches";
 import { ReactiveRule, constraintNameToValidationError } from "./config";
-import { getMetadataForType } from "./configure";
+import { getConstructorFromTag, getMetadataForType } from "./configure";
 import { findByUniqueDataLoader } from "./dataloaders/findByUniqueDataLoader";
 import { findCountDataLoader } from "./dataloaders/findCountDataLoader";
 import { findDataLoader } from "./dataloaders/findDataLoader";
@@ -155,10 +155,17 @@ export interface TimestampFields {
   deletedAt: string | undefined;
 }
 
-export interface EntityManagerOpts<TX = unknown> {
-  driver: Driver<TX>;
-  preloadPlugin?: PreloadPlugin;
-}
+export type EntityManagerOpts<TX = unknown> =
+  | {
+      driver: Driver<TX>;
+      preloadPlugin?: PreloadPlugin;
+      em?: EntityManager<any, any, any>;
+    }
+  | {
+      driver?: Driver<TX>;
+      preloadPlugin?: PreloadPlugin;
+      em: EntityManager<any, any, any>;
+    };
 
 export interface FlushOptions {
   /** Skip all validations, including reactive validations, when flushing */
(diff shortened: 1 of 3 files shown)


packages/orm/src/EntityManager.ts
```diff
@@ -233,33 +240,21 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
...
-  constructor(em: EntityManager<C, Entity, TX>);
   constructor(ctx: C, opts: EntityManagerOpts<TX>);
   constructor(ctx: C, driver: Driver<TX>);
-  constructor(emOrCtx: EntityManager<C, Entity, TX> | C, driverOrOpts?: EntityManagerOpts<TX> | Driver<TX>) {
-    if (emOrCtx instanceof EntityManager) {
-      // Setup from the existing EntityManager
-      const em = emOrCtx;
-      this.driver = em.driver;
-      this.#preloader = em.#preloader;
+  constructor(ctx: C, driverOrOpts: EntityManagerOpts<TX> | Driver<TX>) {
```
zgavin, line 238: note: the old behavior of allowing the creation of an `em` from another `em` was, imo, a footgun due to the way it interacted with `ctx`.  I think requiring a context makes it more explicit that it's up to the end user to manage `em` inside that `ctx`, unless they're explicitly calling `fork`.

packages/orm/src/EntityManager.ts
```diff
@@ -155,10 +155,17 @@ export interface TimestampFields {
...
+export type EntityManagerOpts<TX = unknown> =
+  | {
+      driver: Driver<TX>;
+      preloadPlugin?: PreloadPlugin;
+      em?: EntityManager<any, any, any>;
+    }
+  | {
+      driver?: Driver<TX>;
+      preloadPlugin?: PreloadPlugin;
+      em: EntityManager<any, any, any>;
```
stephenh: I'm not totally following this type union:

- if I'm passing `driver`, why would I want to pass `em?` (doesn't passing `driver` imply "make a new em"), and vice versa
- if I'm passing `em` why would I want to pass `driver?` (doesn't passing `em` impl "use the same driver as this em"?)

packages/orm/src/EntityManager.ts
```diff
@@ -1905,6 +1900,67 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
       );
     }
   }
+
+  fork(): EntityManager<C, Entity, TX> {
```
stephenh, line 1920: Some jsdocs for this would be great

packages/orm/src/EntityManager.ts
```diff
@@ -1905,6 +1900,67 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
...
+    // copy the context so that it's distinct between the two ems and so we can update any references from the old em
+    // to the new one later
+    const ctx: C = { ...oldEm.ctx };
+    const newEm = new EntityManager<C, Entity, TX>(ctx, { em: oldEm });
+    // hydrate (ie, create) each entity from the oldEm in the newEm
+    oldEm.#entitiesByTag.forEach((entities, tag) => {
+      // tags are always defined by the base constructor, so we use that as hydrate expects the base class as its argument
+      const baseCstr = getConstructorFromTag(tag);
+      // We copy the raw row data for each entity to avoid any potential conflict. Then we ensure the correct concrete
+      // __class key is set for hydrate to use to actually instantiate the entity.
```
stephenh: 👍 nice reminder

packages/orm/src/EntityManager.ts
```diff
@@ -1905,6 +1900,67 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
...
+      const rows = entities.map((e) => {
+        const { row, data } = getInstanceData(e);
+        return { ...(row ?? data), __class: e.constructor.name };
```
stephenh: Kinda wonder if we should only set `__class` specifically when CTI is involved, just to avoid iirc a few unnecessary lines in `hydrate` that run when `__class` is detected.

packages/orm/src/EntityManager.ts
```diff
@@ -1905,6 +1900,67 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
...
+          loadedRelations.map((field) => [
+            field,
+            // this should be safe because we are only using loaded relations and we made sure to copy every entity between ems
+            toArray((oldEntity as any)[field].get).map((e: Entity) => newEm.#entitiesById.get(e.idTagged)!),
+          ]),
+        );
+        newEm.#preloadedRelations.set(oldEntity.idTagged, cache);
+        const newEntity = newEm.#entitiesById.get(oldEntity.idTagged)!;
+        // force each relation to preload on the new entity
+        loadedRelations.forEach((relation) => (newEntity as any)[relation].preload());
```
stephenh, line 1978: Fwiw I was thinking we could skip explicitly calling `.preload()` from within `.fork()` itself, and just wait until the caller does/does not call the `.get` / `.isLoaded` methods on the relations, which would then lazily/just-in-time call `.preload()`.

Does that work out, or do we need to do this?

packages/orm/src/EntityManager.ts
```diff
@@ -1905,6 +1900,67 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
...
+      });
+    });
+    // Inspect the top level keys in the context and replace any references from oldEm with newEm while also replacing
+    // any entities from oldEm with their version in newEm
```
stephenh: Maybe include an example like `i.e. update ctx.internalUser`

packages/orm/src/EntityManager.ts
```diff
@@ -1905,6 +1900,67 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
...
+            toArray((oldEntity as any)[field].get).map((e: Entity) => newEm.#entitiesById.get(e.idTagged)!),
+          ]),
+        );
+        newEm.#preloadedRelations.set(oldEntity.idTagged, cache);
+        const newEntity = newEm.#entitiesById.get(oldEntity.idTagged)!;
+        // force each relation to preload on the new entity
+        loadedRelations.forEach((relation) => (newEntity as any)[relation].preload());
+      });
+    });
+    // Inspect the top level keys in the context and replace any references from oldEm with newEm while also replacing
```
stephenh: Maybe throw in `...oldEm with newEm (i.e. update ctx.em)`

packages/orm/src/EntityManager.ts
zgavin: none of the relations explicitly call `preload` themselves from any `get` / `isLoaded` path.  It's only called from data loaders which are obviously async.  If that situation changes then we can remove this, but in the meantime i didn't want this PR to sprawl.

packages/orm/src/EntityManager.ts
zgavin: the purpose is to make one of `driver` or `em` required but still allow the other.   They are not exclusive anymore.

> if I'm passing driver, why would I want to pass em? (doesn't passing driver imply "make a new em"), and vice versa

because passing `em` also copies transaction hooks and the preload plugin

> if I'm passing em why would I want to pass driver? (doesn't passing em impl "use the same driver as this em"?)

if you wanted to change the driver of your new em (admittedly this seems like a niche use case), but keep its other properties.  An explicit driver always override's an `em`'s driver.

stephenh: > because passing em also copies transaction hooks and the preload plugin

I buy (while passing an em) wanting to copy the em's txn hooks & preload plugin, but swapping out the driver seems, as you said, pretty niche -- I guess maybe someday if Joist is used for "read from one db/driver, write to the other db/driver" sort of things. :thinking: 

Kinda want to keep the simpler `{ driver: ... } | { em: ... }`, if only b/c I think its simpler for users to understand at a glance, vs. stopping to reason (as I did) about "when/should I be passing both".

packages/orm/src/EntityManager.ts
stephenh: Ah okay, I really thought the `.isLoaded`s probed `isPreloaded` but I guess not; agreed that could be a follow up.

stephenh (approved):

packages/orm/src/EntityManager.ts
zgavin: i updated it to make the type exclusive again

stephenh (approved):
Looks great; excited to have this

PR #1553, opened 2025-07-29
feat: Add em.merge.

stephenh:


packages/tests/integration/src/EntityManager.merge.test.ts
```diff
@@ -0,0 +1,196 @@
...
+
+describe("EntityManager.merge", () => {
+  it("can merge one source and one target", async () => {
+    const em = newEntityManager();
+    // Given two authors with books
+    const a1 = newAuthor(em, { firstName: "a1" });
+    const a2 = newAuthor(em, { firstName: "a2" });
+    const b1 = newBook(em, { title: "b1", author: a1 });
+    const b2 = newBook(em, { title: "b2", author: a2 });
+    await em.flush();
```
zgavin: nit: the flushes in these tests seem unnecessary

zgavin (approved):

packages/tests/integration/src/EntityManager.merge.test.ts
stephenh: Yeah, I'd been thinking having the entities persisted would be more production realistic, but really should have been using the `insert...` methods to make sure the entities are not `DeepNew` / overly loaded when we invoke `em.merge` but was being lazy & just using the joist factories.

Went ahead and swapped over to `insert...`s.

PR #1555, opened 2025-07-29
fix: address issues with em.fork

zgavin:
* add support for forking an `em` with pending changes
* correctly build each new `row` from combination of `data` / `originalData` / `row`
  * requires a new method on each `...Serde` to convert entity data into what that data would look like coming out of postgres 
* implement preloading for `PolymorphicReference`
* update `Temporal.ZonedDateTime` mapper to correctly account for offset
* additional tests

### DIFF
--- a/packages/orm/src/EntityManager.ts
+++ b/packages/orm/src/EntityManager.ts
@@ -46,6 +46,7 @@ import {
   ValidationRuleResult,
   asConcreteCstr,
   assertIdIsTagged,
+  assertLoaded,
   deepNormalizeHint,
   getBaseAndSelfMetas,
   getBaseMeta,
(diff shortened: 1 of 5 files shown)


packages/orm/src/EntityManager.ts
```diff
@@ -2006,79 +2006,124 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
    * - Copy any pending changes (must call .flush() first)
    * - Share any changes made in either EntityManager with the other
    *
+   * @param opts options to control the fork behavior
+   *  - allowPendingChanges - allows the em being forked to have pending changes and forces the resulting em to have
+   *  `mode` set to "in-memory-writes".  Changes to entities will be persisted across ems. Optional. default: false
```
stephenh: "changes to entities will be persisted across ems" makes me think that they'll stay in sync? Like if I fork `em1` into `em2` and then make continual changes in `em1`, I assume they don't show up in `em2`?

packages/orm/src/temporalMappers.ts
```diff
@@ -35,20 +35,28 @@ export const plainDateTimeMapper: CustomSerde<Temporal.PlainDateTime, string> =
...
-      // Should we use the application's time zone here? Afaiu we're using an explicit
-      // offset anyway, so I believe the time zone is effectively irrelevant, albeit maybe
-      // it would be used for DST/etc nuances when doing date calculations.
-      fromDb: (s) => t.ZonedDateTime.from(s.replace(" ", "T") + "[UTC]"),
-      // Match the pg `TIMESTAMPTZ` format, i.e. "2021-01-01 12:00:00-05:00"
-      toDb: (zdt) => `${zdt.toPlainDate().toString()} ${zdt.toPlainTime().toString()}${zdt.offset}`,
+      // Produce a ZDT from a PG output like "2021-01-01 12:00:00-05:00"
+      fromDb: (s) => {
+        const [offset] = s.match(/([+-]\d{2}(?::?\d{2})?)$/) ?? [];
+        return t.ZonedDateTime.from(`${s}[${!offset || /^\+00(:?00)?$/.test(offset) ? "UTC" : offset}]`);
```
stephenh, line 57: So this change is making sure `00` or `00:00` end up as `UTC` instead of `+0`?

packages/orm/src/serde.ts
```diff
@@ -57,6 +57,8 @@ export interface Column {
   dbType: string;
   /** From the given `__orm.data` hash, return this columns value, i.e. for putting in `UPDATE` params. */
   dbValue(data: any, entity: Entity, tableName: string, fixups: InsertFixup[] | undefined): any;
+  /** Used by em.fork to create an __orm.row from an __orm.data. Should output what we would expect from a db query **/
+  rowValue(data: any): any;
```
stephenh, line 61: Wonder if we should rename `dbValue` and `rowValue` to `writeValue` and `readValue`?

`rowValue` is definitely better than `rowValue`, but I kinda like being more explicit about "this is used for writes" / "this is used for reads".

packages/orm/src/EntityManager.ts
```diff
@@ -2604,3 +2649,24 @@ function getDefaultWriteFn(ctx: unknown): WriteFn {
...
+      const value: any =
+        field.fieldName in originalData || field.fieldName in data
+          ? // If our field is in originalData, then the field has been changed since flush. Our `row` should
+            // reflect what would have come from the db so use originalData when present
+            column.rowValue(field.fieldName in originalData ? originalData : data)
+          : // data is lazy and isn't set until it's accessed, so if the field isn't present there, then we should
+            // be safe to pull the raw data out of `row`
+            oldRow[column.columnName];
+      row[column.columnName] = value ?? null;
+    });
```
stephenh, line 2673: Yep this lgtm 👍

packages/orm/src/EntityManager.ts
```diff
@@ -2006,79 +2006,124 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
...
+        const index = parseInt(oldEntity.toString().split("#").pop()!) - 1;
+        return newEm.getEntities(oldEntity.constructor as EntityConstructor<Entity>)[index];
+      } else {
+        return newEm.getEntity(oldEntity.idTagged)!;
+      }
+    }
+
+    // If we are allowing pending changes, then we need to copy any changes across to newEm
+    if (allowPendingChanges) {
+      oldEm.entities.forEach((oldEntity) => {
```
stephenh: Is it worth filtering by `isDirtyEntity` here?

packages/orm/src/temporalMappers.ts
zgavin: yes

packages/orm/src/EntityManager.ts
zgavin: they don't, ill update the text

packages/orm/src/serde.ts
zgavin: makes sense, think i'd prefer not to do that in this PR though

packages/orm/src/EntityManager.ts
zgavin: not sure how much of a performance boost that'd be, but suppose it can't hurt

PR #1556, opened 2025-07-29
feat: add em.migrateEntity

zgavin:
Migrates an entity from one EntityManager to another, ensuring all relations specified in the hint are also migrated.  The principle use case here is to make it so our `cloneContext` in `graphql-service` can be sync and not need to do any round trips to the database


packages/orm/src/EntityManager.ts
```diff
@@ -2128,6 +2129,69 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
...
+   * The migration process:
+   * 1. If the entity already exists in this EntityManager, returns it
+   * 2. Otherwise, creates a new instance with the same data
+   * 3. Recursively migrates any loaded relations specified in the hint
+   *
+   * @param original - The entity to migrate from another EntityManager
+   * @param hint - The load hint specifying which relations to migrate
+   * @returns The migrated entity with all specified relations loaded
+   */
+  migrateEntity<T extends Entity, H extends LoadHint<T>>(original: T, hint: H): Loaded<T, H>;
```
stephenh: Wdyt about `importEntity`? migrate imo implies it's taken out of the oldEm.

stephenh: Should the `original: T` be `original: Loaded<T, H>`? otherwise we're just kinda trusting the caller to "just know" that the `original` is loaded with `hint`?

packages/orm/src/EntityManager.ts
```diff
@@ -2128,6 +2129,69 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
     });
     return newEm;
   }
+
+  /**
+   * Migrates an entity from one EntityManager to another, ensuring all relations specified in the hint are
+   * also migrated.
```
stephenh: So afaiu `em.fork` == "make a new entity that 'imports' all existing entities", and `em.migrateEntity` == "import only a few entities"? But otherwise they're largely doing the same thing?

If so, would be nice to work that compare/contrast into the docs.

zgavin: migrate also only copies over a specific hint, which means iterating through async props / custom relations / etc.  If it just had fork's behavior of copying all loaded relations it would sprawl outside the expected scope and ruin em isolation.

packages/orm/src/EntityManager.ts
zgavin: yeah i think i went about 4 rounds with the naming and still ended up with something i wasnt really happy with, `importEntity` works for me

packages/orm/src/relations/CustomCollection.ts
```diff
@@ -15,6 +15,7 @@ export type CustomCollectionOpts<T extends Entity, U extends Entity> = {
   remove?: (entity: T, other: U) => void;
   /** Whether the reference is loaded, even w/o an explicit `.load` call, i.e. for DeepNew test instances. */
   isLoaded: () => boolean;
+  loadHint?: ((entity: T) => LoadHint<T>) | LoadHint<T>;
```
stephenh, line 22: Maybe jsdoc when/why CustomCollections/References should provide the `loadHint`? Like as a user I will probably be wondering "should I always implement this? What happens / what breaks / doesn't work if I don't?"

stephenh (approved):

PR #1566, opened 2025-08-06
feat: em.fork performance improvements

zgavin:


packages/orm/src/relations/ManyToManyCollection.ts
```diff
@@ -179,6 +179,19 @@ export class ManyToManyCollection<T extends Entity, U extends Entity>
...
+    function map(v: U[] | undefined): U[] | undefined {
+      if (v === undefined) return undefined;
+      const result = new Array<U>(v.length);
+      for (let i = 0; i < v.length; i++) result[i] = findEntity(v[i]);
+      return result;
+    }
+
+    this.#loaded = map(other.#loaded);
+    this.#addedBeforeLoaded = map(other.#addedBeforeLoaded);
+    this.#removedBeforeLoaded = map(other.#removedBeforeLoaded);
```
stephenh, line 192: These `import`s look just fine; they do make me a little nervous wrt to test coverage; i.e. do our `em.fork` tests probe relations for each each of the m2o/o2o/m2m/etc types to make sure they came over correctly? Can be a todo if not.

packages/orm/src/EntityManager.ts
```diff
@@ -17,65 +17,67 @@ import { Driver } from "./drivers";
 import { Entity, Entity as EntityW, IdType, isEntity } from "./Entity";
 import { FlushLock } from "./FlushLock";
 import {
+  asConcreteCstr,
+  assertIdIsTagged,
+  assertLoaded,
```
stephenh, line 22: These re-ordered imports might be unnecessary / from using different tools? Or maybe a version bump of prettier/somethign changed it and it hasn't been updated yet

packages/orm/src/EntityManager.ts
```diff
@@ -2039,115 +2041,105 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
...
-      const [unpersistedEntities, persistedEntities] = allowPendingChanges
-        ? partition(entities, (e) => e.isNewEntity)
-        : [[], entities];
+    for (const [tag, entities] of oldEm.#entitiesByTag.entries()) {
       // tags are always defined by the base constructor, so we use that as hydrate expects the base class as its argument
       const baseCstr = getConstructorFromTag(tag);
-      // We copy the raw row data for each entity to avoid any potential conflict. Then we ensure the correct concrete
-      // __class key is set for hydrate to use, if necessary (ie, cti), to actually instantiate the entity.
-      const rows = persistedEntities.map((e) => createRowFromEntityData(e));
+      let rows = new Array<any>(entities.length);
```
stephenh, line 2053: Nice passing the `entities.length` to the array cstr

packages/orm/src/EntityManager.ts
```diff
@@ -2039,115 +2041,105 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
...
+      let i = 0;
+      for (const entity of entities) {
+        const instanceData = (entity as any).__data as InstanceData;
+        if (!allowPendingChanges && instanceData.pendingOperation !== "none") {
+          fail("Cannot fork an EntityManager with pending changes");
```
stephenh, line 2058: Nice, moving this into the same for loop, so we don't pay two iteration costs

stephenh (approved):

packages/orm/src/relations/ManyToManyCollection.ts
zgavin: all our tests pass in gql-service, so i'm sure it's fine :sweat_smile:

packages/orm/src/EntityManager.ts
zgavin: this is after running `yarn format`

packages/orm/src/relations/ManyToManyCollection.ts
zgavin: tests added, good callout since they revealed 2 issues

PR #1572, opened 2025-08-16
fix: Fix reacting to new-then-deleted o2os.

stephenh:


stephenh:
This is also no longer a bug on current main.
- PR #1830 (https://github.com/joist-orm/joist-orm/pull/1830) fixed created-then-deleted reverse traversal and shipped in v2.2.0.
- referenceHistory now preserves former owners during delete cleanup: fields.ts:83 and reactiveHints.ts:464.
- The direct regression test is at EntityManager.reactions.test.ts:815.
- The test passes in all four variants: stock, preloading, lazy, and lazy-preloading.

PR #1579, opened 2025-09-05
add chain reactivity test

haislip:


packages/orm/src/relations/ReactiveField.ts
```diff
@@ -95,22 +95,29 @@ export class ReactiveFieldImpl<T extends Entity, H extends ReactiveHint<T>, V>
   }
 
   async load(opts?: { forceReload?: boolean }): Promise<V> {
+    // Even without `forceReload=true`, any explicit calls to `.load()` ==> ensure a fresh value,
+    // because `.get` may have cached the stale/previously-calculated value.
+    //
+    // (Currently `em.populate` / `populateDataLoader` has a hack/escape hatch that ignores calling
+    // RF.load(), to avoid pulling in the RF subgraph over using the materialized value, which ideally would be
+    // replaced with us learning to do the maybeDirty approach.)
```
stephenh, line 103: Just this comment since it should apply to the whole `load` method and not just the `if` check

packages/orm/src/relations/ReactiveField.ts
```diff
@@ -95,22 +95,29 @@ export class ReactiveFieldImpl<T extends Entity, H extends ReactiveHint<T>, V>
...
     if (!this.isLoaded || opts?.forceReload) {
-      // Even without `forceReload=true`, any explicit calls to `.load()` ==> ensure a fresh value,
-      // because `.get` may have cached the stale/previously-calculated value.
-      //
-      // (Currently `em.populate` / `populateDataLoader` has a hack/escape hatch that ignores calling
-      // RF.load(), to avoid pulling in the RF subgraph over using the materialized value, which ideally would be
-      // replaced with us learning to do the maybeDirty approach.)
-      this.#isCached = false;
```
stephenh, line 105: This `isCached=false` being here, outside the `populate` was the core bug b/c we would:

* set `isTest.isCached=false`
* start waiting for `isTest` to be loaded
* while that was happening `isTestChain` would say "hey I can go now"
* `isTestChain` would call `isTest`, and `isTest` knew it still wasn't loaded, but would go ahead and return its stale value (which is not incorrect, as long as we remember to `isTestChain` about the new value later, which we do), *but* we'd also set `isTest.isCached=true` as a side-effect
* when `isTest` was finally loaded, `isTest.get` would run to "update isTest" but see "oh hey I'm already `isCached=true` so just stop

packages/orm/src/relations/ReactiveField.ts
```diff
@@ -120,12 +127,13 @@ export class ReactiveFieldImpl<T extends Entity, H extends ReactiveHint<T>, V>
     // Some ReactiveFields can have surprisingly expensive calculations, especially in loops
     // or those that invoke loops, so cache the value.
     if (this.#isCached) return this.fieldValue;
-    // isLoaded will watch for our previously-loaded subgraph being mutated, and if we've
-    // drifted to not-loaded, it's better to fail and tell the user.
+    // isLoaded will watch for our previously-loaded subgraph being mutated (isLoaded is locally cached but hooked into
+    // our IsLoadedCache cache that is immediately invalidated on write-path), and if we've drifted to not-loaded, it's
+    // better to fail and tell the user.
```
stephenh, line 132: Just updated the in-source comment to acknowledge `isLoaded` is cached but also dirtied by writes.

PR #1586, opened 2025-09-13
feat: Lazy relations w/o transforms

stephenh:
This replaces the `transform-properties` AST-rewrite approach (which was coupled to the TypeScript transform API) to creating lazy relations with a pure JavaScript/runtime based approach.

The new approach leverages JavaScript's prototypes to "still create `Author`s" but we avoid going through the `new` operator (which invokes the constructor and assigns/initializes class fields, which is what we want to skip), and instead create the instance using `Object.create(Author.prototype)`.

This gets us all of the normal getters/setters from the base class/prototype, as well as a working `e instanceof Author`, b/c the entities still have `Author.prototype` as their prototype, which insinuates `Author` is their constructor, even though it really was not.

A simple reproduction from the [proof-of-concept repo](https://github.com/stephenh/lazy-ts-fields/blob/main/src/index.ts):

```ts
export class Author {
  books = hasBooks();
}

export function createFastAuthor(): Author {
  // Would only do this once during init
  Object.defineProperty(Author.prototype, "books", {
    configurable: true,
    get(this: any) {
      return (this.relations["books"] ??= hasBooks());
    },
  });
  const a = Object.create(Author.prototype);
  a.relations = {} as Record<string, any>;
  return a as Author;
}
```

During boot, we create a "fake Author" to discover all of the user-defined class fields like `books = hasBooks()` or `numberOfBOoks = hasReactiveField()`, and then `Object.defineProperty` getter versions of each field on the `Author.prototype`.

Luckily `getProperties` was already doing this on-boot property-discover / fake-instance creation, so we just leverage that.

A ~full list of changes:

* Changes all of the `has...` factory functions to return the `lazyRelation` / `RelationConstructor` wrapper
  * Now that `author.books = hasOne(...)` is only ever invoked once, on boot, the return value is a `RelationConstructor` that we can call `hasOne(...).create(entity, fieldName)` against, each time we want a "new books" relation
  * Removes the `entity`, `fieldName` and `otherMetadata` params from the `has...` functions b/c our getter can pass them
* Moves the codegen'd relations like `get books(): Collection` back to fields like `readonly books: ...` b/c now codegen-defined & user-defined relations can be lazied via the same infra/approach
* Makes `transientFields` also a lazy getter, such that it will not be instantiated on instances until `this.transientFields` is called, during which the getter will make a copy of the tranientFields, and only then assign the instance-level `transientFields` value
* Removes the old `transform-properties` code


packages/codegen/src/generateEntityCodegenFile.ts
```diff
@@ -248,6 +248,16 @@ export function generateEntityCodegenFile(config: Config, dbMeta: DbMetadata, me
       };
 
       ${relations.filter((r) => r.kind === "abstract").map((r) => r.line)}
+      ${relations
```
stephenh, line 251: Per PR overview, moved the codegen relations back to fields, since they can be lazied via the same infra now.

This will reduce the LOC in the output, so in some small way might help our "load the world" cost

packages/codegen/src/generateEntityCodegenFile.ts
```diff
@@ -924,19 +915,17 @@ function createPgEnums(meta: EntityDbMetadata) {
...
     if (m2o.derived === "async") {
-      const line = code`
-        abstract readonly ${fieldName}: ${ReactiveReference}<${entity.name}, ${otherEntity.type}, ${maybeOptional}>;
-      `;
+      const line = code`abstract readonly ${fieldName}: ${ReactiveReference}<${entity.name}, ${otherEntity.type}, ${maybeOptional}>;`;
       return { kind: "abstract", line } as const;
     }
     const decl = code`${ManyToOneReference}<${entity.type}, ${otherEntity.type}, ${maybeOptional}>`;
-    const init = code`${hasOne}(this, ${otherEntity.metaType}, "${fieldName}", "${otherFieldName}")`;
+    const init = code`${hasOne}("${otherFieldName}")`;
```
stephenh, line 928: Updated all of the `has...` methods to not need `this` or `fieldName` or `otherMetadata`

packages/orm/src/relations/CustomCollection.ts
```diff
@@ -14,14 +15,22 @@ export type CustomCollectionOpts<T extends Entity, U extends Entity> = {
   add?: (entity: T, other: U) => void;
   remove?: (entity: T, other: U) => void;
   /** Whether the reference is loaded, even w/o an explicit `.load` call, i.e. for DeepNew test instances. */
-  isLoaded: () => boolean;
+  isLoaded: (entity: T) => boolean;
```
stephenh, line 18: The pattern of `isLoaded = () => this...` meant the `isLoaded` lambda's `this` variable was bound to the fake instance on boot.

The old transform-properties technically rewrote the `isLoaded` lambda into the `getter`, so the `this` instance was still "the right entity".

But now that we do eval the lambda only once, and then reuse that `isLoaded` lambda across many instances (each with their own `relationConstructor.create` instance), we need the lambda to really not use `this`.

packages/orm/src/relations/CustomCollection.ts
```diff
@@ -14,14 +15,22 @@ export type CustomCollectionOpts<T extends Entity, U extends Entity> = {
...
   /**
    * A load hint that can be used to speculatively traverse the dependencies of this collection.  Currently
    * only used by em.importEntity and can be omitted if the collection isn't used with it.
    */
   loadHint?: ((entity: T) => LoadHint<T>) | LoadHint<T>;
 };
 
+export function hasCustomCollection<T extends Entity, U extends Entity>(
+  opts: CustomCollectionOpts<T, U>,
+): CustomCollection<T, U> {
```
stephenh, line 28: The `readonly coll = new CustomCollection` callers now need to be `coll = hasCustomCollection`

packages/orm/src/relations/OneToOneReference.ts
```diff
@@ -20,6 +21,24 @@ import { RelationT, RelationU } from "./Relation";
 
 const OneToOne = Symbol();
 
+/** An alias for creating `OneToOneReference`s. */
+export function hasOneToOne<T extends Entity, U extends Entity>(
```
stephenh, line 25: Moved `hasOneToOne` up from the bottom of the file, to match the locations in the other files.

packages/orm/src/BaseEntity.ts
```diff
@@ -1,14 +1,6 @@
 import { IdType } from "./Entity";
 import { Entity, EntityManager, InstanceData, TaggedId, deTagId, getMetadata, keyToNumber } from "./index";
 
-export let currentlyInstantiatingEntity: Entity | undefined;
```
stephenh, line 4: RelationConstructor.create(entity, fieldName) is now able to pass `entity` to all of the relation cstrs without needing the global variable workaround.

packages/orm/src/EntityManager.ts
```diff
@@ -2171,7 +2171,7 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
 
   /** Creates a new `type` and marks it as loaded, i.e. we know its collections are all safe to access in memory. */
   #doCreate<T extends EntityW>(type: EntityConstructor<T>, opts: any, partial: boolean): T {
-    const entity = new (asInternalCstr(type))(this, true);
+    const entity = newEntity(this, type, true);
```
stephenh, line 2173: There are only two places in `EM` that called `new`, and both are going through `newEntity` now

packages/orm/src/getProperties.ts
```diff
@@ -33,7 +33,9 @@ export function getProperties(meta: EntityMetadata): Record<string, any> {
   // on fake instances" getters to throw nonsense errors anyway (which we suppress), so it should be fine.
   const cached = (propertiesCache[key] = {});
 
-  const instance = getFakeInstance(meta);
+  const fakeEm = undefined as any;
+  const instance = new (meta.cstr as any)(fakeEm, true);
```
stephenh, line 37: This is the only place we call `new` to invoke the constructors & field initializers, so we can capture the ones that are `RelationConstructor`s and later turn them into getters.

packages/orm/src/getProperties.ts
```diff
@@ -44,57 +46,56 @@ export function getProperties(meta: EntityMetadata): Record<string, any> {
   // the GraphQL resolvers. So we should probably just remove this filter and let everything
   // get returned as properties.
   const knownPrimitives = Object.values(meta.allFields)
-    .filter((f) => f.kind === "primaryKey" || f.kind === "primitive" || f.kind === "enum")
+    .filter((f) => f.kind === "primaryKey" || ((f.kind === "primitive" || f.kind === "enum") && !f.derived))
     .map((f) => f.fieldName);
 
-  const properties = Object.fromEntries(
```
stephenh, line 50: This is largely the same, just doesn't have the outer `Object.fromEntries`

packages/orm/src/getProperties.ts
```diff
@@ -44,57 +46,56 @@ export function getProperties(meta: EntityMetadata): Record<string, any> {
...
-        // Return the value of `instance[key]` but wrap it in a try/catch in case it's
-        // a getter that runs code that fails b/c of the dummy state we're in.
-        try {
-          return [key, (instance as any)[key] ?? unknown];
-        } catch {
-          return [key, unknown];
-        }
-      })
-      // Purposefully return methods, primitives, etc. so that `entityResolver` can add them to the resolver
-      .filter(([key]) => key !== "fullNonReactiveAccess" && key !== "transientFields"),
```
stephenh, line 68: We temporarily return `transientFields`, so that we can getter-ize it, but otherwise later in this method re-filter it out to keep the same API contract

packages/orm/src/getProperties.ts
```diff
@@ -47,60 +48,78 @@ export function getProperties(meta: EntityMetadata): Record<string, any> {
...
 const unknown = new UnknownProperty();
 
 const propertiesCache: Record<string, any> = {};
-const fakeInstances: Record<string, Entity> = {};
-
-/**
- * Returns a fake instance of `meta` so that user-defined `CustomReference` and `AsyncProperty`s can
- * be inspected on boot.
- */
-export function getFakeInstance(meta: EntityMetadata): Entity {
```
stephenh, line 84: The only caller of `getFakeInstance` really should have been using `getProperties`, so moved them over to that.

packages/codegen/src/generateEntityCodegenFile.ts
zgavin: really love that it's using the same infra now :+1:

packages/orm/src/BaseEntity.ts
zgavin: :open_hands:

docs/src/content/docs/faq.md
```diff
@@ -54,8 +54,6 @@ This can initially feel awkward, but it provides a truly type-safe API, given th
 
 This is often how business logic wants to interact with the domain model--a continual incremental loading of data as needed, as conditional codepaths are executed, instead of an endpoint/program exhaustively knowing up-front exactly what data will be necessary.
 
-If performance is a concern (loading thousands of entities with many custom properties), Joist provides a [ts-patch transform](/advanced/transform-properties) to rewrite the properties as lazy getters in production builds.
```
zgavin, line 57: suggestion: feel like we should still explain somewhere in the documentation that we're runtime re-writing your entity prototypes.   If someone is debugging locally, the current implementation might be confusing since the properties would jump from the entity itself to the prototype.  eg, `Object.entries(entity)` would be different before and after this change if they didn't have transform-properties on.

packages/orm/src/newEntity.ts
```diff
@@ -0,0 +1,100 @@
...
+ * the instance-level class fields/relations like `hooks = hasMany`.
+ *
+ * Instead, we just call `Object.create`, which does not make any constructor/super calls, and instead
+ * let any `this.books` accesses resolve to the getters we've installed on the prototype.
+ */
+export function newEntity<T extends Entity>(em: EntityManager, cstr: EntityConstructor<T>, isNew: boolean): T {
+  if (!(cstr as any)[lazySymbol]) {
+    moveRelationsToGetters(cstr);
+    (cstr as any)[lazySymbol] = true;
+  }
```
zgavin, line 23: question: why do this lazily instead of just at startup?

packages/orm/src/newEntity.ts
```diff
@@ -0,0 +1,100 @@
...
+ * So instead we just do not bother passing metadata consts to `has...` methods, and instead let the methods
+ * one-time resolve `otherMetadata` on their first invocation, which will be after the `metadata.ts` file has
+ * finished importing, and the consts are all defined.
+ */
+export function resolveOtherMeta(entity: Entity, fieldName: string): EntityMetadata {
+  return (getMetadata(entity).allFields[fieldName] as any).otherMetadata();
+}
+
+/** Wraps `has...` relation constructors in an easily-identifiable container. */
+export class RelationConstructor<T extends Entity> {
```
zgavin: suggestion: imo this naming is confusing since it's being used for more than just relations (eg, async methods, reactive fields, async props, etc).  I'd suggest `FieldConstructor` or `PropertyConstructor` instead.  Feels like the entire folder should be renamed for consistency, but would be fine with just this class.

packages/orm/src/newEntity.ts
```diff
@@ -0,0 +1,100 @@
...
+ */
+export function newEntity<T extends Entity>(em: EntityManager, cstr: EntityConstructor<T>, isNew: boolean): T {
+  if (!(cstr as any)[lazySymbol]) {
+    moveRelationsToGetters(cstr);
+    (cstr as any)[lazySymbol] = true;
+  }
+  // This side-steps the `Author` constructor that initializes fields instance-level fields, which instead
+  // we've shoved up to be getters on the prototype, and the only instance field we expect to have is `__data`,
+  // which is set by the `baseEntityCstr` call.
+  const entity = Object.create(cstr.prototype) as T;
```
zgavin, line 27: question: could we use the second option of `Object.create` to install properties so we don't need to muck with the prototype in `moveRelationsToGetters`? ie, that method would just create an immutable properties definition that could be passed here.  Not sure what that does to performance, but i would assume that `Object.create` wouldn't copy the properties definition so it'd just be a simple assignment.

packages/orm/src/relations/CustomCollection.ts
zgavin: praise: nice.  imo this was an oversight originally anyways. It never should have relied on `this` as it's different from the behavior of every other callback.

packages/orm/src/relations/ManyToManyCollection.ts
```diff
@@ -13,29 +13,31 @@ import {
...
-  return new ManyToManyCollection<T, U>(
-    joinTableName,
-    entity,
-    fieldName,
-    columnName,
-    otherMeta,
-    otherFieldName,
-    otherColumnName,
-  );
+  let otherMeta: EntityMetadata<U>;
```
zgavin, line 28: suggestion: copy the comment explaining why we have to resolve this lazily to each of these relation helpers.  If someone is making a change to one of the individual relations/props they might not have the context as to why this needs to be this way.

packages/orm/src/relations/hasManyDerived.ts
```diff
@@ -23,18 +23,19 @@ export function hasManyDerived<T extends Entity, U extends Entity, H extends Loa
...
-      if (load) {
-        return load(entity, opts);
-      } else {
-        return entity.em.populate(entity, { hint: loadHint, ...opts });
-      }
-    },
-    isLoaded: () => isLoaded(entity, loadHint),
-    loadHint,
-    ...(rest as any),
+  return lazyRelation((entity: T) => {
```
zgavin: request: can we use this as an opportunity to make this have the same signature as `hasOneDerived`? eg, instead of taking an `opts` it just takes a `get` function?  If a user wants to customize the other callbacks in `opts`  they can use `newCustomCollection` instead.  I just checked our codebase, and the only thing we are ever using (across 50 `hasManyDerived`) is `get`.

packages/orm/src/newEntity.ts
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * A function for `has...` methods to integrate with the `newEntity` lazy relation system.
+ *
+ * We need TypeScript to still see `books = hasMany(...)` as being typed as `Many<Book>`,
+ * so this method's return type is `R` i.e. the `Many<Book>` relation type.
+ *
+ * But at runtime we actually want this to be a `RelationConstructor` that can be rewritten
+ * into a getter, and only invoked when the relation is actually accessed.
+ */
+export function lazyRelation<T extends Entity, R>(fn: (entity: T, fieldName: string) => R): R {
```
zgavin: suggestion: ditto previous comment about the naming, I guess i'd prefer `lazyProperty`.

zgavin (approved):
:clinking_glasses: really elegant solution

docs/src/content/docs/faq.md
stephenh: Agreed -- not sure where; I have a WIP blog post that describes the approach, but agreed it should end up in the docs.

packages/orm/src/newEntity.ts
stephenh: Well, we have enough "wait for the world to boot" pain that I figured any "don't do it until absolutely necessary" optimization is worth it.

packages/orm/src/newEntity.ts
stephenh: Makes sense, I could rename to `FieldConstructor` / `lazyField` instead of relation--will push that up.

packages/orm/src/newEntity.ts
stephenh: I didn't know about the 2nd arg to `Object.create`, but I think/assume its preferrable to keep them on the prototype b/c it avoids allocating instance-levels slots...

packages/orm/src/relations/hasManyDerived.ts
stephenh: Makes sense but will leave as a follow-up

PR #1591, opened 2025-09-16
feat: add config.addReaction

zgavin:
This pr adds a new `config.addReaction` method that is somewhere between a hook and a reactive field.  The provided callback is intended to be arbitrary code like a hook.  It is called during flush on any change to its reactive hint, like a reactive field or rule.  Unlike a hook, this means it won't be called on every flush for an entity and can be called even when the entity itself does not change.  It also means that it can be called multiple times per flush, so care must be taken to avoid circular dependencies and ensure the result of the callback is idempotent.

### DIFF
--- a/packages/orm/src/config.ts
+++ b/packages/orm/src/config.ts
@@ -176,7 +176,34 @@ export class ConfigApi<T extends Entity, C> {
     this.addHook("afterCommit", fn);
   }
 
+  /**
+   * Adds a reaction that runs during flush whenever fields in the `hint` change.
+   *
+   * Reactions are somewhere in between hooks and reactive fields/references:
+   * 1. Can make arbitrary changes to any entity like a hook
+   * 2. Only run when the provided hint has changes, not on every flush, like an RF/RR
+   * 3. Run when the entity itself has no changes, like an RF/RF
+   * 4. Can run multiple times per flush, like an RF/RF.  Be careful to avoid creating
+   *    circular dependencies in the hint and to make the function idempotent.
+   *
+   * @param hint The fields to watch for changes and load before running the reaction
+   * @param fn The reaction function to run
+   */
   addReaction<H extends ReactiveHint<T>>(hint: H, fn: HookFn<Loaded<T, H>, C>): void;
+  /**
+   * Adds a named reaction that runs during flush whenever fields in the `hint` change.
+   *
+   * Reactions are somewhere in between hooks and reactive fields/references:
+   * 1. Can make arbitrary changes to any entity like a hook
+   * 2. Only run when `hint` has changes, not on every flush, like an RF/RR
+   * 3. Can run when the entity itself has no changes, like an RF/RF
+   * 4. Can run multiple times per flush, like an RF/RF.  Be careful to avoid creating
+   *    circular dependencies in the hint and to make the function idempotent.
+   *
+   * @param name A name to identify this reaction for debugging
+   * @param hint The fields to watch for changes and load before running the reaction
+   * @param fn The reaction function to run
+   */
   addReaction<H extends ReactiveHint<T>>(name: string, hint: H, fn: HookFn<Loaded<T, H>, C>): void;
   addReaction<H extends ReactiveHint<T>>(
     nameOrHint: string | H,


packages/orm/src/logging/ReactionLogger.ts
```diff
@@ -1,8 +1,8 @@
 import ansis from "ansis";
-import { ReactiveField } from "../config";
+import { ReactiveActor } from "../config";
```
stephenh: The term "actor" is often used to mean "erlang actors, akka actors, etc" -- maybe `Reactable`? ... `Reactive...` thinking 🤔

zgavin: it's an actor because it preforms an action.  Really don't want to do all those renames again :sweat_smile:

stephenh: Eh, yeah the terms of pretty unique though, so doing case-sensitive 

* `ReactiveActor` -> `Reactable`
* `reactiveActors` -> `reactables`

Etc with global find/replace (I used the `sd` command line utility last time, as an easier sed) seems easy enough.

It's fair to sit on it, i.e. wait & see if `Reactable` is still the best term after some thinking, but would like to see the rename happen at the last minute/before merging.

stephenh: `Reactable` is my current favor; is just `Reaction` taken as a term? Looks like not

packages/orm/src/config.ts
```diff
@@ -290,6 +312,32 @@ export interface ReactiveField {
...
+  cstr: MaybeAbstractEntityConstructor<any>;
+  /** The name (source location) of the downstream reaction. */
+  name: string;
+  /** The path from this source entity to the downstream entity for this reaction. */
+  path: string[];
+  /** The downstream reaction function. */
+  fn: HookFn<any, any>;
+}
+
+export type ReactiveActor = ReactiveField | Reaction;
```
stephenh: Maybe `Reaction = ReactiveField | ReactiveHook`?

I.e. rename the new `export interface Reaction` to `ReactiveHook` since afaiu its the hook-specific version of `ReactiveField`?

zgavin: I actually did have this as `ReactiveHook` initially which i don't like for a handful of reasons:

* It muddies the water on the difference between this and the other hooks.   I think it should be clear that this does not behave like any of the other hooks and needs to be treated like a rule or reactive field.
* The naming clashes with the existing fields for hooks on `ConfigData`  
* The method itself is `addReaction`, so it seems logical that the thing it creates is a `Reaction`

stephenh: Hm, yeah I get those reasons...

It was just tempting to use "reaction" as the term for "ReactiveField or this-new-thing".

I'm good with "Reaction === this-new-thing", but then the search continues for "the term that means both", with `Reactable` my current best/preferred idea...

packages/orm/src/config.ts
```diff
@@ -83,12 +85,14 @@ export class ConfigApi<T extends Entity, C> {
...
     } else {
       const hint = ruleOrHint;
+      let loadHint: LoadHint<T>;
       // Create a wrapper around the user's function to populate
       const fn = (entity: T) => {
-        // Ideally we'd convert this once outside `fn`, but we don't have `metadata` yet
-        const loadHint = convertToLoadHint(getMetadata(entity), hint);
+        // Ideally, we'd convert this outside `wrappedFn`, but we don't have `metadata` yet. So we do it lazily on the
+        // first call.
+        loadHint ??= convertToLoadHint<T>(getMetadata(entity), hint);
```
stephenh, line 93: Great call making this cached.

packages/orm/src/config.ts
```diff
@@ -172,21 +176,74 @@ export class ConfigApi<T extends Entity, C> {
...
+    maybeFn?: HookFn<Loaded<T, H>, C>,
+  ): void {
+    // Keep the name so we can uniquely identify this reaction later and also aid debugging/tracing
+    const name = typeof maybeFn === "function" ? (nameOrHint as string) : getCallerName();
+    const hint = (maybeFn ? hintOrFn : nameOrHint) as H;
+    const fn = (maybeFn ?? hintOrFn) as HookFn<Loaded<T, H>, C>;
+    this.ensurePreBoot(name, "addReaction");
+    let loadHint: LoadHint<T>;
+    // Create a wrapper around the user's function to populate
+    const wrappedFn = (entity: T, ctx: C) => {
```
stephenh, line 220: Random musing, but seems like eventually we could have these `entities: T[]` to avoid `em.populate` just having to re-combine the individual calls

packages/tests/integration/src/entities/Author.ts
```diff
@@ -388,6 +399,53 @@ config.setDefault(
...
+});
+
+// m2m reaction
+config.addReaction("m2m", { tags: "name" }, (a) => {
+  a.transientFields.reactions.m2m += 1;
+});
+
+// reactive field reaction
+config.addReaction("rf", "search", (a) => {
+  a.transientFields.reactions.rf += 1;
```
stephenh, line 424: Nice, really great adding tests for these

PR #1592, opened 2025-09-17
docs: add note about joist-transform-properties

blimmer:
This PR documents the changes from https://github.com/joist-orm/joist-orm/pull/1586

### DIFF
--- a/docs/astro.config.mts
+++ b/docs/astro.config.mts
@@ -9,6 +9,9 @@ import { glob } from 'glob';
 // https://astro.build/config
 const config = defineConfig({
   site: "https://joist-orm.io",
+  redirects: {
+    "/advanced/transform-properties": "/faq/#why-is-joist-transform-properties-no-longer-required",
+  },
   integrations: [
     starlight({
       title: "Joist",
@@ -94,7 +97,10 @@ function getLlmConfig(): StarlightLllmsTextOptions {
     ],
     exclude: [
       "blog/**", // this is mostly marketing content, not helpful for LLMs
-      "why-joist", // just a redirect
+
+      // redirects
+      "why-joist",
+      "advanced/transform-properties",
     ],
   }
 }
--- a/docs/src/content/docs/faq.md
+++ b/docs/src/content/docs/faq.md
@@ -127,3 +127,18 @@ Then you either need to enable `skipLibCheck: "true"` in your `tsconfig.json` (r
 Joist uses [ts-poet](https://github.com/stephenh/ts-poet) and [dprint-node](https://github.com/devongovett/dprint-node) to generate & format code, as dprint is significantly faster than Prettier when generating large amounts of code.
 
 The ts-poet output attempts to be "prettier-ish", but if you'd like to customize it, you can create a `.dprint.json` file as per the [dprint docs](https://dprint.dev/setup/#hidden-config-file).
+
+## Why is `joist-transform-properties` no longer required?
+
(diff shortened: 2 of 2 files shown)


docs/astro.config.mts
```diff
@@ -9,6 +9,9 @@ import { glob } from 'glob';
 // https://astro.build/config
 const config = defineConfig({
   site: "https://joist-orm.io",
+  redirects: {
+    "/advanced/transform-properties": "/faq/#why-is-joist-transform-properties-no-longer-required",
```
blimmer, line 13: This automatically excludes the page in the `robots.txt` file. It does a client-side redirect since it's a staticly built site.

docs/src/content/docs/faq.md
```diff
@@ -127,3 +127,18 @@ Then you either need to enable `skipLibCheck: "true"` in your `tsconfig.json` (r
...
+
+Starting with version 1.268.0, Joist has moved to a pure JavaScript runtime solution that eliminates the need for `joist-transform-properties`. The new implementation:
+
+- **Uses JavaScript prototypes**: Creates entity instances with `Object.create(Author.prototype)` without invoking constructors
+- **Runtime lazy loading**: Relations are created on-demand using prototype getters instead of compile-time transforms
+- **Simpler build process**: No longer requires TypeScript transform configuration in your build tools
+
+If you're upgrading from an older version of Joist, you can safely remove `joist-transform-properties` from your dependencies and build configuration. All lazy relation functionality continues to work exactly the same from a developer perspective, but now uses a more elegant runtime approach.
+
+**Migration**: Simply remove `joist-transform-properties` from your `package.json` and any transform configuration from your TypeScript/webpack setup. No code changes are required in your entities or application logic.
```
blimmer, line 144: Is this true? Or are code changes required?

blimmer: Ah, from the code changes in an internal project, I think there's more needed. I think we need to at least convert those over to a getter-style reference?

zgavin: I don't think so? they already should be getter style definitions.  There may be some issues with `transientFields` we're still working out.

zgavin: after fixing all our ~1300 test failures, they were almost entirely due to 2 issues:

1. not accounting for inheritance when setting up lazy relations.  This has been fixed in joist and was accounting for ~1000 test failures.
2. Usage of `this` inside custom relations / reactive fields / reactive references / async props.  This was never intended usage and now produces undefined behavior.  We probably need to make this a more explicit failure, as it can be a pain to debug.

Additionally, stray props in entities need to be converted into `transientFields`.

blimmer: Cool, so this migration guide is good as-is?

zgavin: If you're doing things the way you were "suppose" to then there shouldn't be any changes required.  As our repo demonstrates, people dont always do things as they are suppose to.  So might be good to touch on `this` usage and stray props other than `transientFields`.

stephenh: I think this FAQ entry is good/great as-is; I'd like to eventually/soon make a dedicated page on the lazy field approach:

https://github.com/joist-orm/joist-orm/issues/1596

And then probably move ~a few paragraphs of this FAQ over to that + more fully explain the "things you cannot do"

stephenh (approved):

PR #1595, opened 2025-09-18
chore: disallow `this` in property callbacks

zgavin:
Due to the changes in #1586, `this` inside a property initialization callback (such as for `hasAsyncProp`) no longer refers to the instance being instantiated, but rather to the fake instance generated by `getProperties`.  This can result in unpredictable behavior and is tricky to debug, as often the error won't reveal itself immediately upon access.  This PR addresses this by making any access of a property on `this` inside one of these callbacks immediately throw an error.  We accomplish this by deleting the "own" properties on the fake instance and then installing a proxy as its prototype, which allows us to respond to any access by immediately failing.

### DIFF
--- a/packages/orm/src/getProperties.ts
+++ b/packages/orm/src/getProperties.ts
@@ -1,7 +1,7 @@
 import { BaseEntity } from "./BaseEntity";
 import { EntityMetadata } from "./EntityMetadata";
 import { LazyField } from "./newEntity";
-import { partition } from "./utils";
+import { fail, partition } from "./utils";
 
 /**
  * Returns the relations in `meta`, both those defined in the codegen file + any user-defined `CustomReference`s.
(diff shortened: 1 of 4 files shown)


packages/tests/integration/src/entities/Author.test.ts
```diff
@@ -687,4 +687,11 @@ describe("Author", () => {
     const authors = await em.find(Author, { certificate: { eq: new Uint8Array([11, 22]) } });
     expect(authors.length).toBe(1);
   });
+
+  it("fails when trying to use this inside a property callback", async () => {
```
zgavin: note: wasn't sure where this test should live

packages/orm/src/getProperties.ts
```diff
@@ -92,9 +92,20 @@ export function getProperties(meta: EntityMetadata): Record<string, any> {
...
+  // intercept any attempts to access `this` from within the callbacks and fail.
+  Object.setPrototypeOf(instance, instancePrototypeProxy);
+  for (const prop of Object.getOwnPropertyNames(instance)) {
+    if (prop !== "__data") delete instance[prop];
+  }
+
   return cached;
 }
 
+const instancePrototypeProxy = new Proxy({}, { get: () => fail("Cannot use 'this' in a property callback") });
```
stephenh: This is good, would be great to eventually link to a page on joist-orm.io that explains why, maybe like an "Advanced / Lazy Relations" page that explains the approach + the gotchas like this one.

https://github.com/joist-orm/joist-orm/issues/1596

stephenh: Wouldn't mind renaming this to something that shows it is "only for the post-boot phase", like `disallowAfterBoot` / `throwAfterBoot` / I dunno, don't love those names, but something like that.

As-is, just from the name in isolation, it sounds like its always the fake instance's prototype

packages/tests/integration/src/entities/Author.test.ts
stephenh: Maybe `Entity.test.ts`?

stephenh (approved):

PR #1602, opened 2025-09-23
feat: add runOnce option to addReaction

zgavin:
adds an optional `runOnce` flag to `addReaction` that causes it to operate more like a traditional hook, eg only once per flush call regardless of additional changes

### DIFF
--- a/packages/orm/src/config.ts
+++ b/packages/orm/src/config.ts
@@ -30,6 +30,8 @@ export type EntityHook =
   | "afterCommit";
 type HookFn<T extends Entity, C> = (entity: T, ctx: C) => MaybePromise<unknown>;
 
+type AddReactionOpts<T extends Entity, C> = { runOnce?: boolean; fn: HookFn<T, C> };
+
 export const constraintNameToValidationError: Record<string, string> = {};
 
 type Settable<T extends Entity> = keyof SettableFields<FieldsOf<T>> & string;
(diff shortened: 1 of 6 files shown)


packages/orm/src/config.ts
```diff
@@ -205,15 +224,35 @@ export class ConfigApi<T extends Entity, C> {
    * @param fn The reaction function to run
    */
   addReaction<H extends ReactiveHint<T>>(name: string, hint: H, fn: HookFn<Loaded<T, H>, C>): void;
```
stephenh, line 209: Wdyt about reducing the overloads by moving `name` into one of the `opts`? 

Or maybe something like:

```
addReaction(
  hint,
  name | { name, runOnce },
  fn
);
```

So `fn` is always the 3rd parameter (not maybe-in-the-opts), `hint` is always first (which I think? we need for type inference to work), and the 2nd parameter is either "just name" or "an opts with name or runOnce flag".

...and maybe the 2nd arg itself is optional, so there are two overloads:

* `addReaction(hint, fn)` // most boring / like today's hooks
* `addReaction(hint, name | opts, fn)` // if you want to pass name or opts

Admittedly opts is usually the last param, but I kinda like fn being the last top-level param.

Seem fine/wdty?

zgavin: not in love with having `name` as the 2nd param.    With complex hint it'd get lost easily.  my thought for `name` was that it could be used for inheritance eventually or expose it somehow that other hooks could use it as a dependency (eg, reaction A should run after reaction B).

stephenh: Maybe just swap the order then, so:

* `addReaction(hint, fn)` // boring/today's hooks
* `addReaction(name | opts, hint, fn)`

Kinda tempted to force name to always be an opt, like:

```
addReaction({ name: "asdf" }, hint, fn);
```

But letting the 1st arg be `name | opts` seems fine too.

I mostly like keeping hint & fn always as top-level params, and now that you point it out, always together / as the last two params?

zgavin: i like that better, implemented it :+1:

stephenh (approved):

PR #1603, opened 2025-09-24
feat: add basic Plugin and PluginManager support

zgavin:


.gitignore
```diff
@@ -21,5 +21,6 @@ junit*.xml
 .idea/**/usage.statistics.xml
 .idea/**/dictionaries
 .idea/**/shelf
+.idea/material_*.xml
```
zgavin, line 24: this is a file added by the material theme ui plugin, just adding it to .gitignore so i dont have to keep manually ignoring it

stephenh: I'd been using the Material UI theme & Icon plugins, but uninstalled both just to keep the plugin list (and hence potential performance impact) as minimal as possible...

I've been using https://plugins.jetbrains.com/plugin/19668-gerry-themes-pro instead and seems fine.

Disclaimer I have no hard evidence that either of these plugins matter/impact performance, this is just my current cargo culted "seems to work" setup 😅

packages/orm/src/EntityManager.ts
```diff
@@ -303,6 +305,8 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
       get fieldLogger() {
         return em.#fieldLogger;
       },
+
+      pluginManager: new PluginManager(em),
```
stephenh: I'd kinda like this to be `undefined` until any plugins exist, that way call-sites like

```
pluginManager.beforeSetField(entity, fieldName, newValue);
```

Can pretty trivially become

```
pluginManager?.beforeSetField(entity, fieldName, newValue);
```

And we skip calling into the method even for "extremely cheap" hash lookups/etc.

stephenh (approved):

packages/orm/src/EntityManager.ts
zgavin: I thought about doing that but it seemed like such a micro-optimization that I didn't bother.  I can do that, though.

zgavin: Implemented this. Also made it so the functions themselves aren't actually defined on `PluginManager` until a plugin that uses them is added.

packages/orm/src/PluginManager.ts
```diff
@@ -0,0 +1,48 @@
...
+          for (const plugin of this.#pluginsByCallback[method]!) {
+            (plugin[method] as (...args: any[]) => unknown)(...args);
+          }
+        };
+      }
+    }
+  }
+}
+
+export interface PluginManager extends PluginMethods {}
```
stephenh, line 48: Ah, I missed this on the 1st/2nd read and was wondering how the conditionally-added methods got added to `PluginManager`; neat!

stephenh (approved):

PR #1604, opened 2025-09-24
fix: assign ids and recalc reactables pending ids within recalcPendingReactables

zgavin:
This PR addresses the following scenario:

1. the initial run of a reactable during flush fails due to a missing id
2. when the reactable runs again during `recalcReactablesPendingAssignedIds` it mutates a field that has its own reactive dependencies
3. the mutated field is on an entity that is already in `alreadyRanHooks` 
4. `findPendingFlushEntities` puts nothing in `pendingHooks`
5. the dependencies of the mutated field are never re-run

### DIFF
--- a/packages/orm/src/EntityManager.ts
+++ b/packages/orm/src/EntityManager.ts
@@ -1411,8 +1411,6 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
 
     this.#fl.startLock();
 
-    this.#rm.clearSuppressedTypeErrors();
-
     await this.#fl.allowWrites(async () => {
       // Cascade deletes now that we're async (i.e. to keep `em.delete` synchronous).
       // Also do this before calling `recalcPendingReactables` to avoid recalculating
@@ -1612,6 +1610,7 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
       if (e instanceof InMemoryRollbackError) return [...allFlushedEntities];
       throw e;
     } finally {
+      this.#rm.clearSuppressedTypeErrors();
       this.#fl.releaseLock();
     }
   }
--- a/packages/orm/src/EntityManager.ts
+++ b/packages/orm/src/EntityManager.ts
@@ -1411,6 +1411,8 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
 
     this.#fl.startLock();
 
+    this.#rm.clearSuppressedTypeErrors();
+
     await this.#fl.allowWrites(async () => {
       // Cascade deletes now that we're async (i.e. to keep `em.delete` synchronous).
       // Also do this before calling `recalcPendingReactables` to avoid recalculating
@@ -1427,7 +1429,6 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
     // Make sure two ReactiveQueryFields don't ping-pong each other forever
     let hookLoops = 0;
     let now = getNow();
-    this.#rm.clearSuppressedTypeErrors();
     const suppressedDefaultTypeErrors: Error[] = [];
 
     // Make a lambda that we can invoke multiple times, if we loop for ReactiveQueryFields
--- a/packages/orm/src/EntityManager.ts
+++ b/packages/orm/src/EntityManager.ts
@@ -1471,20 +1471,8 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
 
           // The hooks could have deleted this-loop or prior-loop entities, so re-cascade again.
           await this.flushDeletes();
-          let loops = 0;
-          do {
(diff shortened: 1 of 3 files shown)


packages/orm/src/ReactionsManager.ts
```diff
@@ -216,7 +212,15 @@ export class ReactionsManager {
...
-  async recalcRelationsPendingAssignedIds(): Promise<void> {
+  get hasPendingReactiveQueries(): boolean {
+    return this.needsRecalc.query;
+  }
+
+  get hasPendingReactables(): boolean {
+    return this.needsRecalc.populate || this.needsRecalc.reaction;
+  }
+
+  async recalcReactablesPendingAssignedIds(): Promise<void> {
```
zgavin: note: this rename got missed in the original `addReaction` PR

packages/orm/src/EntityManager.ts
```diff
@@ -1461,13 +1461,20 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
 
           // The hooks could have deleted this-loop or prior-loop entities, so re-cascade again.
           await this.flushDeletes();
-          // The hooks could have changed fields, so recalc again.
-          await this.#rm.recalcPendingReactables("reactables");
+          let loops = 0;
```
stephenh: Makes sense... I probably thought RM would handle looping internally, but right I guess after doing PendingAssignedIds it doesn't loop...

Maybe it should? Like could we re-organize this a bit such that the "if hasFieldsPendingAssignedIds, do an em.assignIds" actually happens inside of `recalcPendingReactables`?

Maybe with a param to say "assignIdsIfNeeded"?

Just wondering if then maybe this could happen within RM's existing loop infra/check, instead of introducing an additional one.

zgavin: we could just immediately call `assignIdsIfNeeded` inside the rm if it gets an id error and then retry the reactable immediately instead of deferring to later? wdyt?

stephenh: Ah yeah :thinking: ... would we need to worry about multiple RFs failing and kicking off more than 1 `assignIdsIfNeeded`?

That sort of "don't have multiple assign ids if needed in flight at once" is probably why I did the current phase approach, but not against doing the assign ids right away...

zgavin: we could data loader assign ids? then have it remove itself once it finishes.

zgavin: alternatively we could still collect the missing ids, but instead of deferring to later we could do the retry directly in each loop of `recalcPendingReactables`

zgavin: @stephenh rewrote this with the approach from :arrow_up:

packages/orm/src/EntityManager.ts
```diff
@@ -1427,7 +1427,6 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
     // Make sure two ReactiveQueryFields don't ping-pong each other forever
     let hookLoops = 0;
     let now = getNow();
-    this.#rm.clearSuppressedTypeErrors();
```
zgavin, line 1430: note: since errors in reactive fields are immediately populated rather than deferring to `recalcRelationsPendingAssignedIds`, this was clearing out the errors prematurely.  So now we just clear at the end of flush.

packages/tests/integration/src/relations/ReactiveQueryField.test.ts
```diff
@@ -42,8 +42,8 @@ describe("ReactiveQueryField", () => {
     // After the INSERTs, there is a SELECT to calc the data, and then an `UPDATE`
     expect(queries).toMatchInlineSnapshot(`
      [
-       "WITH _find (tag, arg0) AS (VALUES ($1::int, $2::character varying), ($3, $4) ) SELECT array_agg(_find.tag) as _tags, a.* FROM authors AS a CROSS JOIN _find AS _find WHERE a.deleted_at IS NULL AND a.last_name = _find.arg0 GROUP BY a.id ORDER BY a.id ASC LIMIT $5",
        "select nextval('publishers_id_seq') from generate_series(1, 1) UNION ALL select nextval('authors_id_seq') from generate_series(1, 1) UNION ALL select nextval('books_id_seq') from generate_series(1, 2) UNION ALL select nextval('book_reviews_id_seq') from generate_series(1, 2)",
+       "WITH _find (tag, arg0) AS (VALUES ($1::int, $2::character varying), ($3, $4) ) SELECT array_agg(_find.tag) as _tags, a.* FROM authors AS a CROSS JOIN _find AS _find WHERE a.deleted_at IS NULL AND a.last_name = _find.arg0 GROUP BY a.id ORDER BY a.id ASC LIMIT $5",
```
zgavin, line 46: note: the order changed due to `assignIds` now being called before this rqf runs.  I believe the behavior is still correct otherwise.

stephenh (approved):

PR #1605, opened 2025-09-29
feat: expose isFlushing via getEmInternalApi

zgavin:


packages/orm/src/EntityManager.ts
```diff
@@ -2273,6 +2276,7 @@ export interface EntityManagerInternalApi {
   get fieldLogger(): FieldLogger | undefined;
   get isLoadedCache(): IsLoadedCache;
   pluginManager: PluginManager | undefined;
+  isFlushing: boolean;
```
stephenh: Wdyt about having a top-level function like `isRefreshing`?

Kinda thinking that I'm fine with both `isRefreshing` and `isFlushing` between basically public APIs, and so its kosher to write code against those, but anything in `EMInternalApi` is more strictly "do not use outside of joist internals".

Maybe we should just let `isRefreshing` and `isFlushing` be getters directly on the `EntityManager`?

I was probably hestitent to do that for `isRefreshing`, but 🤷 now that we have two ... and given the alternatives, I think getters on `EM` is probably fine.

sgty?

zgavin: Yeah that works.  The reason I didn't make it an exported function is that such a function wouldn't have access to `#fl`, but a method on the em itself would so i can do that.

stephenh (approved):

PR #1607, opened 2025-09-30
feat: rename and add new entity manager transaction hooks

zgavin:


packages/orm/src/drivers/PostgresDriver.ts
```diff
@@ -80,17 +82,20 @@ export class PostgresDriver implements Driver<Knex.Transaction> {
     if (em.txn) {
       return fn(em.txn as Knex.Transaction);
     }
-    return this.knex.transaction(async (txn) => {
+    await beforeTransactionStart(em, this.knex);
```
stephenh: I get going for completeness and parallelism/consistency with the dual `(before|after)...(Commit|Start)`.

Not sure the `beforeTransactionStart` is useful though, b/c afaiu it would be executed in isolation, i.e. with a `this.knex` that is still "whatever connection from the pool" and so not really tied to the upcoming `txn`.

So not sure what use cases it could really drive. Maybe txn pool stats, but seems like the txn pool itself could report on those.

I'd also like to keep `em.beforeCommit` / `em.afterCommit` to match the current `config`-level names... 🤔 

Not sure of a great name for `afterTransactionStart` (current `beforeTransaction`), which afaiu is where we'd want to slot in our cyanaudit user / other connection-level parameters.

Ironically I kinda like the previous name of `beforeTransaction` even though I get it happens "technically after a txn is started", but I guess to me the "before" is "before Joist itself starts issuing SQL within the txn".
 
...still kinda like `beforeTransaction` + `beforeCommit` (the renamed afterTxn) + new EM-level `afterCommit` (and skip the `beforeTransactionStart`).

...if you really want `afterTransactionStart`, I guess that'd be fine, but do kinda want "just beforeCommit / afterCommit" for succinctness / matching the `config` API.

Is that fine?

zgavin: > Not sure the beforeTransactionStart is useful though, b/c afaiu it would be executed in isolation, i.e. with a this.knex that is still "whatever connection from the pool" and so not really tied to the upcoming txn.

This is also true of `afterTransactionCommit` (ie what you want to call `afterCommit`).  Why is one useful and the other is not?

More to the point, my plan was to use `beforeTransactionStart` to enable cyanaudit in tests, and then set the user via `afterTransactionStart`, then finally use `afterTransactionCommit` to disable cyanaudit . That would replicate how it actually works in practice (the user is set in the transaction, cyanaudit is enabled globally).

> I'd also like to keep `em.beforeCommit` / `em.afterCommit` to match the current config-level names... 🤔

They behave differently, though.  These hooks are only called once per transaction while the config level hooks will be called for each flush inside a transaction.  I don't think they should be conflated by naming them the same thing.  If anything, tbh, the config level hooks are inappropriately named.

Also they collide inside `EntityManger.ts` since they're both top level functions.  Obviously that can be worked around, but it's nice that with the current implementation the naming is consistent throughout the codebase for each feature.

stephenh: The `afterTxnCommit` / `afterCommit` have the known use-case of "you want to make I/O calls after the commit is complete" ... granted, I really prefer graphile worker jobs for those sorts of use cases, to get retries, but I still think its basically valid.

For `beforeTxnStart`, if you enable cyanaudit on that knex connection, are we guaranteed the connection given to the next `knex.transaction` / `afterTxnStart` will be that same connection from the pool that has been one-off prepped to enable cyanaudit?

My guess is no, hence just the curiosity of "when would this be useful".

zgavin: > The afterTxnCommit / afterCommit have the known use-case of "you want to make I/O calls after the commit is complete"

But that is not a known use case for `afterTxnCommit`.  We literally don't have any known use cases for `afterTxnCommit` because we didn't have it until this PR.

I am happy to rename `fooTransaction` -> `fooTxn` for brevity 

> will be that same connection from the pool that has been one-off prepped to enable cyanaudit?

It doesn't matter which connection turns off a global setting in the database? only one test runs at a time per DB.

stephenh: > no known use cases for afterTxnCommit

I've already mentioned wire calls, after data is committed, which I know we don't do b/c background jobs are the better way to do that, but imo its still a known use case.

I.e. `after_commit` is a rails hook for the same "do stuff after commit" use case.

This is what the current `afterTransaction`, imo, was supposed to be for, and we're just fixing/restoring it (while also adding a separate/dedicated `beforeCommit` for the add jobs use case).

> not the same connection between the two hooks

Ah okay, I didn't realize/forgot the cyanaudit setting was global, so yes that would work.

> Also they collide inside EntityManger.ts

As you said, we can work around this, but imo its not a factor for this particular user-facing API decision.

I do want `EM.beforeCommit` and `EM.afterCommit`.

You're right that our current `config` hooks are different aka wrong, but we should just fix those (as a follow-up/someday), instead of using that incorrect behavior to inform this decision.

Stepping back, I think I just don't like the "(before|after)...(start)" naming pattern -- ofc I see the the rationale & technical correctness, but in my pseduo-dyslexic head, having "two temporal terms" i.e. "before start" / "after start" just doesn't jive as "ah immediately intuitive".

I get I'm supposed to noun-ize "transactionstart", so there is only one temporal term, but I'd like to avoid doing that.

...one way of doing that is to only have "beforeTransaction" be the "okay I'm technically within a txn but have not issued any SQL yet", i.e. what it was before.

I know this doesn't give us the "actually before even starting the txn" you're looking for, but I don't want that at this point.

If you want to turn on cyanaudit from "not the current `em.beforeTransaction`" hook, couldn't we issue a `ctx.knex` immediately within the `if (enableAuditTrail)`? I.e. does it have to be something done from a Joist hook?

tldr, still want `EM.beforeCommit` / `afterCommit`, no `beforeTransactionStart`, open to better names for `beforeTransaction` that doesn't have the "two temporal terms" but would kinda like to ship with my asks & move on for now; as everyone can tell, we have no problems making breaking API changes when/if we think of better names down the road. 😅

zgavin: I don't really see what the harm is in having something like `beforeTransactionStart` is? It costs us nothing to support and it's not a huge amount of overhead to loop over an additional very small array once per transaction.

how about `beforeBegin` / `afterBegin`? I suppose it's still two "temporal" terms but if we also have `beforeCommit` / `afterCommit` then all the hooks mirror the actual SQL going out over the wire.

zgavin: Updated to use `type EntityManagerHook = "beforeBegin" | "afterBegin" | "beforeCommit" | "afterCommit"`

zgavin: Also for a future rename of the entity level hooks, wdyt about `afterValidation` or `afterRules` to replace `beforeCommit` and `afterFlush` or `afterWrite` to replace `afterCommit`?  IMO these would align with what they actually do.  The equivalent rails callbacks are `after_validation` and `after_save`.

Also after going over our `afterCommit`s yesterday, there's still quite a few with wire calls (iirc all eventbridge or cognito) we should probably turn into jobs.

stephenh: I like `beforeBegin` / `afterBegin` -- much better, ty.

Interesting re `afterValidation` / `afterFlush` -- atm I think I prefer "current names but fixing the semantics", but I guess not totally against the "keep current semantics but better names".

Kinda wonder if we'd end up with all four names, i.e. afterValidation + afterFlush (run per loop) + beforeCommit + afterCommit (match the EM-level before/afterCommit semantics).

packages/orm/src/drivers/PostgresDriver.ts
```diff
@@ -80,17 +82,20 @@ export class PostgresDriver implements Driver<Knex.Transaction> {
     if (em.txn) {
       return fn(em.txn as Knex.Transaction);
     }
-    return this.knex.transaction(async (txn) => {
+    await driverBeforeBegin(em, this.knex);
```
stephenh, line 85: Ngl kinda surprised you don't have a `maybePromise` "put these all in a list that compacts something something" trick up your sleeve that avoids the `await` calls 😅 

(Agreed they won't really matter to perf, but I do enjoy our trend of leaning into micro-optimizations, for when we resurrect / ship the joist-vs-other-orms benchmarks.)

packages/orm/src/drivers/PostgresDriver.ts
zgavin: I'm not sure we actually have a need for a "true" `beforeCommit` / `afterCommit`.   Those would be tricky to implement as we'd need to track every flushed entity across the transaction in addition to across a flush.

Our current hooks using `beforeCommit` almost always rely on `changes` (either directly or through `transientFields`) which would be wiped out each flush, so if we waited until before `COMMIT` is going to be sent we wouldn't see them.

Our `afterCommit` fall into 3 camps:
* reset `transientFields` so the next flush doesn't pick them up again.  These shouldn't wait until after the `COMMIT` is sent.
* Wire calls.  These probably do want to wait for the `COMMIT` to be sent so that the id actually exists, but as we've discussed they should be probably moved to graphile jobs rather than investing time into making them safer from a request.
* Raw sql updates.  These definitely shouldn't want to wait for after `COMMIT` as we'd want them to rollback the transaction if they failed.

packages/orm/src/drivers/PostgresDriver.ts
zgavin: i did initially but it was pretty gross to read a bunch of chained and nested `.then` calls while also propagating the  result through the commit hooks

PR #1611, opened 2025-10-01
feat: add beforeFind / afterFind callbacks support to plugins

zgavin:


packages/orm/src/dataloaders/loadDataLoader.ts
```diff
@@ -9,13 +9,16 @@ import { keyToNumber } from "../keys";
 import { LoadHint } from "../loadHints";
 import { abbreviation, indexBy } from "../utils";
 
+export const loadOperation: `load-${boolean}` = "load-true";
```
stephenh: Seems like the plugin shouldn't care about `load-true` vs. `load-false` as separate operations?

zgavin: yeah i guess it really isn't necessary to pass that through

zgavin: fixed

packages/orm/src/EntityManager.ts
```diff
@@ -181,6 +190,21 @@ export interface FlushOptions {
...
+  | typeof lensOperation
+  | typeof loadOperation
+  | typeof manyToManyLoadOperation
+  | typeof manyToManyFindOperation
+  | typeof oneToManyLoadOperation
+  | typeof oneToManyFindOperation
+  | typeof oneToOneLoadOperation
+  | typeof populateOperation
+  | typeof recursiveChildrenOperation
+  | typeof recursiveParentsOperation;
```
stephenh, line 207: I'm a little worried this is going to end up being too-granular, but 🤷 shouldn't hurt

stephenh (approved):

packages/orm/src/EntityManager.ts
zgavin: I think the only one we actually would care about right now is `findCount`

PR #1615, opened 2025-10-01
chore: Add documentation for reactions

zgavin:


docs/src/content/docs/modeling/reactions.md
```diff
@@ -0,0 +1,135 @@
...
+);
+```
+
+## Best Practices
+
+1. **Keep reactions focused**: Each reaction should handle a single concern
+2. **Make reactions idempotent**: Since they can run multiple times, ensure they produce the same result
+3. **Avoid circular dependencies**: Don't create reactions where A triggers B which triggers A
+4. **Use read-only relations**: Mark relations as `:ro` when you only need to read them, not react to them
+5. **Consider `runOnce`**: If your reaction has side effects (like sending notifications), use `runOnce: true`
```
stephenh: Do we want to throw in "while being aware that you might 'miss' changes after your only run)"?

docs/src/content/docs/modeling/reactions.md
```diff
@@ -0,0 +1,135 @@
...
+6. **Name your reactions**: For easier debugging, especially in complex domain models
```
stephenh: I like the names, but not sure we need to promote them to as best practice; if they were "that strictly best practice" I think we'd require them them all the time? So maybe just delete this point.

stephenh (approved):

docs/src/content/docs/modeling/reactions.md
zgavin: I kinda do feel they *could be* best practice eventually.  If we get around to adding more features based on the name (eg, overriding inherited reactions, making reactions depends on each other, etc) then it'd be good to have names.  I'll remove for now though as I don't feel super strongly about this :shrug:

PR #1617, opened 2025-10-02
chore: improve ESM tests

blimmer:
I'm looking to add ESM support to the graphql-codegen plugin. I want to use the `tests/esm-misc` test package to validate that it produces the expected output.

However, there are a few things about the current test suite that seem a bit strange to me. I've corrected those issues in this PR.

See inline comments for granular details on the changes.


.circleci/config.yml
```diff
@@ -43,7 +43,7 @@ jobs:
             equal: [<< parameters.node-version >>, bun]
           steps:
             - node/install:
-                node-version: "23"
+                node-version: "24"
```
blimmer, line 46: Use current LTS for bun test environment (this is just necessary to run `yarn`)

.circleci/config.yml
```diff
@@ -66,23 +66,15 @@ jobs:
           name: Run codegen
           command: yarn codegen
 
-      # For all Node targets, run everything -- except tests/bun and tests/esm-misc (Jest doesn't load `jest.config.ts` in Node 18/20)
+      # For all Node targets, run everything -- except tests/bun (run in a separate job)
```
blimmer, line 69: Node 20 can run ESM tests, so I'm pulling this back into the main test suite runtime.

.circleci/config.yml
```diff
@@ -66,23 +66,15 @@ jobs:
...
             not:
               equal: [<< parameters.node-version >>, bun]
           steps:
             - run:
                 name: Run tests
-                command: yarn workspaces foreach --worktree --parallel --include 'packages/tests/*' --include packages/orm --exclude packages/tests/bun --exclude packages/tests/esm-misc run test
-      # For Node 23, run esm-misc
-      - when:
-          condition:
-            equal: [<< parameters.node-version >>, 23]
```
blimmer, line 81: Node 23 isn't in the matrix build, so these tests were never running in CI.

packages/tests/esm/src/entities/codegen/AuthorCodegen.ts
```diff
@@ -37,8 +37,8 @@ import {
   type ValueFilter,
   type ValueGraphQLFilter,
 } from "joist-orm";
-import type { Context } from "../../context.ts";
-import { Author, authorMeta, Book, type BookId, type Entity, EntityManager, newAuthor } from "../entities.ts";
+import type { Context } from "../../context.js";
+import { Author, authorMeta, Book, type BookId, type Entity, EntityManager, newAuthor } from "../entities.js";
```
blimmer, line 41: The logic to allow importing .ts extensions is a nice feature. However, I wouldn't describe it as the "happy path" for ESM. The happy path for TS + ESM is to import using .js extensions, since it is compatible with both TypeScript and in the transpiled code.

packages/tests/esm/src/entities/codegen/AuthorCodegen.ts
```diff
@@ -37,8 +37,8 @@ import {
   type ValueFilter,
```
blimmer, line 37: I renamed this test package to just `esm` vs. `esm-misc`.

packages/tests/esm/jest.config.ts
```diff
@@ -2,10 +2,13 @@ import type { Config } from "@jest/types";
 import { createDefaultEsmPreset } from "ts-jest";
 
 export default {
-  ...createDefaultEsmPreset({ isolatedModules: true }),
+  ...createDefaultEsmPreset(),
```
blimmer, line 5: This setting trigger this warning:

```
ts-jest[config] (WARN)
    The "ts-jest" config option "isolatedModules" is deprecated and will be removed in v30.0.0. Please use "isolatedModules: true" in /Users/blimmer/code/homebound/joist-orm/packages/tests/esm/tsconfig.json instead, see https://www.typescriptlang.org/tsconfig/#isolatedModules
```

So, I've added `isolatedModules` to `tsconfig.json`

packages/tests/esm/jest.config.ts
```diff
@@ -2,10 +2,13 @@ import type { Config } from "@jest/types";
...
   testMatch: ["<rootDir>/src/**/*.test.(ts|tsx)"],
   globalSetup: "<rootDir>/src/setupTestEnv.ts",
   testEnvironment: "node",
+  moduleNameMapper: {
+    "^(\\.{1,2}/.*)\\.js$": "$1",
+  },
```
blimmer, line 11: Back in https://github.com/joist-orm/joist-orm/pull/1353, @stephenh mentioned that he had to use the `.ts` imports via `tsconfig.json` because Jest required it.

This is how you use ESM-style imports with ts-jest.

packages/tests/esm-misc/package.json
```diff
@@ -35,8 +35,5 @@
     "ts-node": "^10.9.2",
     "tsconfig-paths": "^4.2.0",
     "typescript": "5.9.2"
-  },
-  "imports": {
-    "src/": "src/*"
```
blimmer, line 40: Node subpath imports must start with `#`, so this configuration was invalid. I'm working on subpath import support in another branch and will likely create a new integration test for that, specifically.

packages/tests/esm/tsconfig.json
```diff
@@ -1,16 +1,16 @@
 {
   "compilerOptions": {
-    "target": "es2022",
-    "module": "Node16",
-    "moduleResolution": "Node16",
+    "target": "esnext",
+    "module": "nodenext",
+    "moduleResolution": "nodenext",
```
blimmer, line 5: These are the settings we should use for modern ESM typescript projects.

packages/tests/esm/jest.config.ts
stephenh: I vaguely recall struggling this, i.e. getting jest to rewrite the .js-s back to .ts so it can load it off the file system, but then it would (to my recollection) break loading random things out of `node_modules` and other "wow I hate this" sort of weirdness.

Weird but makes sense solution that we just drop the js, don't add it back, and let jest reprobe "maybe that should have been js or ts or tsx or what not".

stephenh (approved):

PR #1620, opened 2025-10-03
feat: support ESM in graphql-codegen

blimmer:
## Background

This PR resolves issues in `joist-graphql-codegen` related to ESM.

Specifically, this PR:

- Fixes a bug in the root `joist-codegen` that did not apply the esm extension to enum re-exports 
- Updates all grapqhl-codegen (ESM or not) to use `type`-only imports when appropriate (required for `verbatimModuleSyntax` tsconfig setting)
- Produces `grapqhl-codegen-joist.mjs` (vs. just `graphql-codegen-joist.js`) in ESM format
- Updates import paths to use `.js` (or `.ts` if Joist's `allowImportingTsExtensions` is `true`) when `esm` is `true` in the config. Existing (e.g., CJS / the default) imports are not updated.

Fixes #1619

## Remaining Fixes

The `@homebound/grapqhl-typescript-simple-resolvers` package still needs to updated to produce ESM-compatible import statements:

```shell
 ~/code/joist-orm/packages/tests/esm   fix-codegen  yarn tsc --noEmit
src/generated/graphql-types.ts:2:10 - error TS1484: 'GraphQLResolveInfo' is a type and must be imported using a type-only import when 'verbatimModuleSyntax' is enabled.

2 import { GraphQLResolveInfo, GraphQLScalarType } from "graphql";
           ~~~~~~~~~~~~~~~~~~

src/generated/graphql-types.ts:3:10 - error TS1484: 'Context' is a type and must be imported using a type-only import when 'verbatimModuleSyntax' is enabled.

3 import { Context } from "src/context.js";
           ~~~~~~~

src/generated/graphql-types.ts:4:37 - error TS5097: An import path can only end with a '.ts' extension when 'allowImportingTsExtensions' is enabled.

4 import { Author, Book, Color } from "src/entities/index.ts";
                                      ~~~~~~~~~~~~~~~~~~~~~~~

src/generated/graphql-types.ts:107:23 - error TS5097: An import path can only end with a '.ts' extension when 'allowImportingTsExtensions' is enabled.

107 export { Color } from "src/entities/index.ts";
                          ~~~~~~~~~~~~~~~~~~~~~~~
```

PR #1623, opened 2025-10-03
chore: regenerate integration test codegen files

blimmer:
During development of #1620 , I noticed that many of the files in the integration test were generated with old logic. This PR removes the codegen'ed files, regenerates them, and restores existing tests.

This makes the test suite more realistic and up-to-date with current best practices.


packages/tests/integration/src/resolvers/adminUser/adminUserResolvers.test.ts
```diff
@@ -0,0 +1,17 @@
+import { newAdminUser } from "src/entities";
```
blimmer, line 1: This file appears to have never been generated, although it was in the `history` file. Is there a reason we don't want it?

packages/tests/integration/src/resolvers/largePublisher/saveLargePublisherMutation.ts
```diff
@@ -3,7 +3,7 @@ import { MutationResolvers } from "src/generated/graphql-types";
 import { saveEntity } from "src/resolvers/utils";
 
 export const saveLargePublisher: Pick<MutationResolvers, "saveLargePublisher"> = {
-  async saveLargePublisher(root, args, ctx) {
+  async saveLargePublisher(_, args, ctx) {
     return { largePublisher: await saveEntity(ctx, LargePublisher, args.input) };
   },
 };
```
blimmer, line 9: I think all the files below were generated with an older filesystem layout setup (`/objects/`), but we have a new layout. `git` idenfiied some of these as file moves, but not others.

PR #1626, opened 2025-10-05
fix: Fix recalc RFs with new/deleted entities.

stephenh:


packages/orm/src/ReactionsManager.ts
```diff
@@ -202,20 +206,11 @@ export class ReactionsManager {
...
         results.forEach((result) => {
-          if (result.status === "rejected") {
-            if (result.reason instanceof TypeError) {
-              // Defer to the validation error to catch this
-              this.suppressedTypeErrors.push(result.reason);
-            } else {
-              failures.push(result.reason);
-            }
-          }
+          if (result.status === "rejected") failures.push(result.reason);
```
stephenh: What if a RF failed once b/c of an IdError but then fails as 2nd time due to a `TypeError`?

packages/orm/src/ReactionsManager.ts
```diff
@@ -233,6 +228,39 @@ export class ReactionsManager {
...
+    );
+    const endTime = this.logger?.now() ?? 0;
+    this.logger?.logLoadingTime(this.em, endTime - startTime);
+
+    const failures: any[] = [];
+    results.forEach((result) => {
+      if (result.status === "rejected") {
+        if (result.reason instanceof TypeError) {
+          // Defer to the validation error to catch this
+          this.suppressedTypeErrors.push(result.reason);
```
stephenh, line 254: Kinda wonder if `suppressedTypeErrors` should be a map from RF -> reason, so that if somehow the RF ends up passing in some subsequent loop, we could remove it from the `suppressedTypeErrors` map. Not for this PR, just wondering.

packages/orm/src/EntityManager.ts
```diff
@@ -1510,6 +1513,16 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
           await this.flushDeletes();
           // The hooks could have changed fields, so recalc again.
           await this.#rm.recalcPendingReactables("reactables");
+          // We may have reactables that failed earlier, but will succeed now that hooks have been run and cascade
+          // deletes have been processed
+          if (this.#rm.hasPendingTypeErrors) {
+            await this.#rm.recalcPendingTypeErrors();
+            // If we successfully re-ran and have dirtied something new, then we need to run the entire flush loop
```
stephenh: By "something new" I think we mean "an entity that has already had hooks ran (so not a new entity)" but "a new/not-yet-calced RF on some already-looped entity, that, when we recalc the net RF, then dirties other RFs"?

...afaiu we'd only need to do the flush loop again if there are actually newly-dirtied entities, which is already how the flush loop works, which is why it seems like "just another `recalcPendingReactables(...)` here would be fine?

B/c if that recalc touches/dirties "entities that haven't had hooks ran yet" then our flush loop will already pick them up, but if it hits "entities that have had hooks run" then we won't rerun the hooks, so why do the flush loop?

packages/orm/src/ReactionsManager.ts
zgavin: ah yeah good point.  dang.

packages/orm/src/ReactionsManager.ts
zgavin: So like 

1. first flush loop iteration runs 
2. some reactions dirty entities
3. a reaction fails
4. second iteration runs
5. hooks from dirtied entities fix the failure
6. failure reaction runs again and succeeds (since presumably it'd be dependent on whatever was fixed changing)
7. we fail after validation because there's still `suppressedTypeErrors`

Seems like an edge case but definitely possible :thinking:

packages/orm/src/EntityManager.ts
zgavin: The issue is if you run `recalcPendingReactables` here again, those could also have type errors that now won't be in `suppressedTypeErrors`.  You have to run `recalcPendingTypeErrors` again to get those to transition, but they could be resolved if you let hooks run again so you don't want to just immediately run it again.  

An alternative would be, after `findPendingFlushEntities` but still in the loop, to do something like: 

`if (pendingHooks.size === 0 && this.#rm.hasPendingTypeErrors) await this.#rm.recalcPendingTypeErrors()`

packages/orm/src/EntityManager.ts
```diff
@@ -1510,11 +1511,22 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
...
+            if (this.#rm.needsRecalc("reactables")) await this.#rm.recalcPendingReactables("reactables");
+          }
 
           for (const e of pendingHooks) hooksInvoked.add(e);
           pendingHooks.clear();
           // See if the hooks mutated any new, not-yet-hooksInvoked entities
           findPendingFlushEntities(this.entities, hooksInvoked, pendingFlush, pendingHooks, alreadyRanHooks);
+          // The final run of recalcPendingReactables could have left us with pending type errors and no entities in
+          // pendingHooks.  If so, we need to re-run recalcPendingTypeErrors to get those errors to transition into
+          // suppressed errors so that we will fail after simpleValidation.
```
stephenh, line 1527: Nice, makes sense & great comment

zgavin (approved):
🙃

PR #1635, opened 2025-10-09
fix: bump inquirer to resolve `tmp` CVE

### LINKED ISSUE #1634: joist-codegen/joist-graphql-codegen transitively require a version of `tmp` with a CVE
https://github.com/advisories/GHSA-52f5-9888-hmc6

```shell
> yarn why -R tmp
└─ graphql-service@workspace:.
   ├─ exceljs@npm:4.4.0 (via npm:^4.4.0)
   │  └─ tmp@npm:0.2.5 (via npm:^0.2.0)
   ├─ joist-codegen@npm:1.275.0 (via npm:^1.275.0)
   │  ├─ inquirer@npm:9.3.7 (via npm:^9.3.7)
   │  │  └─ external-editor@npm:3.1.0 (via npm:^3.1.0)
   │  │     └─ tmp@npm:0.0.33 (via npm:^0.0.33)
   │  └─ jscodeshift@npm:17.3.0 [c1533] (via npm:^17.3.0 [c1533])
   │     └─ tmp@npm:0.2.5 (via npm:^0.2.3)
   └─ joist-graphql-codegen@npm:1.275.0 (via npm:^1.275.0)
      └─ joist-codegen@npm:1.275.0 (via npm:1.275.0)
```

So we need to update inquirer to fix this.

blimmer:
## Problem

The `joist-codegen` package had a transitive dependency on a vulnerable version of `tmp` (v0.0.33) via the following chain:
```
inquirer@9.3.7 → external-editor@3.1.0 → tmp@0.0.33
```

## Solution

Rather than simply bumping `inquirer` to a newer patch version, this PR migrates to the modern `@inquirer/prompts` package, which:
- No longer depends on `external-editor` (and thus not `tmp@0.0.33`)
- Has a cleaner, more ergonomic API
- Includes TypeScript types natively (no separate `@types/inquirer` needed)
- Is the officially recommended package from the Inquirer.js team

Fixes #1634


packages/codegen/src/codemods/index.ts
```diff
@@ -48,9 +46,7 @@ export async function maybeRunTransforms(config: Config): Promise<void> {
 
   // Otherwise run them
   for (const mod of mods) {
-    const run = await inquirer.prompt({
-      name: "run",
-      type: "confirm",
+    const run = await confirm({
       message: `Do you want to run ${mod.description}?`,
     });
     if (!run) continue;
```
blimmer, line 52: The migration **incidentally fixes a latent bug** in the codemod logic:

**Before:** The old `inquirer.prompt()` API returned `{ run: boolean }`, but the code checked `if (!run)` (checking if the object is falsy). Since objects are always truthy in JavaScript, users could never actually opt out of running codemods.

**After:** The new `confirm()` API returns a boolean directly, so `if (!run)` now works correctly and properly respects user choices.

stephenh: Ah! Thank you!

PR #1636, opened 2025-10-09
feat: Add cte.kind=ast to ParsedFindQuery.

stephenh:


packages/tests/integration/src/drivers/PostgresDriver.cte.test.ts
```diff
@@ -47,4 +47,51 @@ describe("PostgresDriver.cte", () => {
...
+
+    const parsed: ParsedFindQuery = {
+      selects: ["a.*"],
+      tables: [{ join: "primary", alias: "a", table: "authors" }],
+      ctes: [
+        {
+          alias: "min_age_cte",
+          columns: [{ columnName: "min_age", dbType: "int" }],
+          query: {
+            kind: "ast",
```
stephenh, line 67: I don't love the `kind` being here, b/c it leads to a cte.query.query; thinking should probably push the kind up into the cte type itself?

PR #1646, opened 2025-10-18
chore: add CTE support to buildKnexQuery

zgavin:


packages/orm/src/drivers/index.ts
```diff
@@ -1,3 +1,4 @@
+export * from "./buildRawQuery";
```
zgavin, line 1: note: ~~exposing this so users can generate a `ParsedFindQuery` to pass to handlers (eg, `beforeFind`)~~

edit: I meant exposing this so users can generate raw query from a `ParsedFindQuery` for debugging in handlers (eg, `beforeFind`)

stephenh: Like in tests?

I was kinda thinking today that we might eventually usages like:

```
const query = buildRawQuery(someFind, ...);
// customize the knex query...
```

To maybe be a `em.buildQuery` so that, after the initial parsing, it could pass the AST through any plugins, before returning the knex query to callers to further muck with.

Granted, callers are on their own for dealing with "not just the AST -> query" but "AST -> plugin(s) -> query", and handling any complexity/brittleness that results...

zgavin: yeah it was basically for debugging so i could see what queries looked like

PR #1652, opened 2025-10-24
feat: Make em.create ids stable in toString.

### LINKED ISSUE #1625: Make entity.toString stable for new entities
Currently the output of `a.toString()` on a new entity is `a#1` or `a#2` etc, where the `#n` is based on `a`s index in the list of unsaved/new `Author`s.

However once we `em.flush` and/or assign ids, the output switches to `a:1` / `a:2` / etc, using the assigned id, which is:

- a) basically good, but
- b) means we can't easily which "was this a:2 previously a#2 or a#1"?
- c) if the em is still used, some other new instance might take `a#1`

A more stable algorithm would be:

* We keep a counter of `i` per meta
* Each new `em.create`-d new gets assigned its `i++` (would basically match today's behavior)
* The counters are not reset by `em.flush`, so we'd never see `a#1`-used-again (new change)
* The output of `a.toString()` for new-then-assigned entities would be `a#1:1` i.e. both the "#1 you're a new entity" and the ":1 your assigned id" (new change)

The output of `a#1:1` is definitely "different" so we shouldn't jump in to this, but it seems like once we got used to it, the improved ergonomics of tracking "who exactly is a#1 over time" would probably be worth it.

stephenh:
This should help debugging b/c we don't have to guess "which new #1 is now which :1 entity".

It also fixes the performance issue of "calling entity.toString() on a bunch of new entities is basically an O(n^2)".

Fixes #1625


packages/orm/src/BaseEntity.ts
```diff
@@ -141,6 +135,25 @@ export abstract class BaseEntity<EM extends EntityManager, I extends IdType = Id
...
+  // Start with `Author` or `a`, depending on the caller
+  let result = prefix;
+  // Add `#1` if this is an em.create-d entity
+  const data = getInstanceData(entity);
+  const createId = data.createId;
+  if (createId) {
+    result += `#${createId}`;
+  }
+  // Add the `:1` but only after our flush is complete, otherwise the id is
+  // internally assigned but not really flushed/committed to the db yet
```
zgavin, line 148: question: do we really need to filter the ids of the  "not yet flushed" entities? seems like it could still be useful for debugging, ie "here is what i think my id should be"

stephenh: That is what I originally did, but then any validation rule that failed, already had the ids assigned, so changed from "author#1" to "author#1:1" and it just created more noise/changes in the assertions than seemed necessary/worthwhile.

PR #1664, opened 2025-11-14
feat: Trim string values.

stephenh:
This builds on our existing "sanitize empty string to undefined", and just extends it to leading/trailing spaces.

This is automatically disabled if the varchar's default value is `""`, which is how the varchar can declare "i'm okay with empty strings" (obviously, since it has that has its default value).

We could also add sanitize knobs to `joist-config.json` if/as needed in the future.


packages/orm/src/dataloaders/findOrCreateDataLoader.ts
```diff
@@ -20,25 +20,14 @@ export function findOrCreateDataLoader<T extends Entity>(
...
-      ? whereFilterHash
-      : // Otherwise rewrite each of the `{ firstName: "AaA" }` values in each of where/ifNew/upsert hashes
-        (key) =>
-          whereFilterHash({
-            where: maybeLower(meta, key.where),
-            ifNew: maybeLower(meta, key.ifNew as any),
-            upsert: maybeLower(meta, key.upsert),
-          } as any) as FindOrCreateKey<T>;
+  // Do some extra logic to make unclean strings/citext not create duplicate entities
+  const hasAnyStrings = Object.values(meta.allFields).some((f) => f.kind === "primitive" && f.type === "string");
```
stephenh, line 24: Per @chr1sjf0x 's point of `findOrCreate` having the same trim/clean behavior as setters, we now always munge/sanitize `findOrCreate` params that are strings, with the same behavior as setters.

(But we only handed citext correctly, i.e. to do case insensitive string matches)

packages/orm/src/dataloaders/findOrCreateDataLoader.ts
```diff
@@ -66,7 +55,7 @@ export function findOrCreateDataLoader<T extends Entity>(
       // This is fundamentally asking to create the same entity, but with different ifNew/upsert
       // conditions, which we could fail on, but for now just take the first tuple of
       // {where/ifNew/upsert} and assume it wins
-      const [{ where, ifNew, upsert }] = keys;
+      const [{ ifNew, upsert }] = keys;
```
stephenh, line 58: I removed the `where` from the dataloader key, b/c we already have it as part of the batch key, and the batch key has the trimmed values.

packages/orm/src/fields.ts
```diff
@@ -55,15 +55,22 @@ export function isFieldSet(entity: Entity, fieldName: string): boolean {
 export function setField(entity: Entity, fieldName: string, newValue: any): boolean {
   ensureNotDeleted(entity, "pending");
   const { em } = entity;
+  const api = getEmInternalApi(em);
+  const { rm, pluginManager, indexManager, fieldLogger, isLoadedCache } = api;
 
-  getEmInternalApi(em).checkWritesAllowed();
+  api.checkWritesAllowed();
```
stephenh, line 61: We had several repeated calls to `getEmInternalApi` so made it a local variable.

packages/orm/src/utils.ts
```diff
@@ -207,8 +207,13 @@ export function mergeNormalizedHints(target: any, source: any): void {
 const newLine = /\n/g;
 const doubleSpace = /  +/g;
 
-export function cleanStringValue(value: unknown) {
-  return typeof value === "string" && value.trim() === "" ? undefined : value;
+export function cleanStringValue(value: unknown): string | undefined {
+  // If this is not a string, coerce it -- or maybe fail?
+  let s = typeof value !== "string" && value !== undefined && value !== null ? String(value) : value;
```
stephenh, line 212: @zgavin this is the new string coercion

packages/tests/integration/src/entities/Comment.ts
```diff
@@ -6,7 +6,10 @@ export class Comment extends CommentCodegen {
     "parentTags",
     { parent: { commentParentInfo: {}, tags: "name" } },
     (c) => {
-      return [c.parent.get?.commentParentInfo.get, ...(c.parent.get?.tags.get.map((t) => t.name) ?? [])].join("-");
+      return (
+        // Use `|| "-"` because the db field is required, and empty string will be coerced to null
+        [c.parent.get?.commentParentInfo.get, ...(c.parent.get?.tags.get.map((t) => t.name) ?? [])].join("-") || "-"
```
stephenh, line 11: @zgavin this was a RF in our test suite that was returning `""` and not getting swapped to null, so didn't even need new test cases for this one 😬

packages/tests/integration/src/EntityManager.findOrCreate.test.ts
```diff
@@ -334,6 +334,22 @@ describe("EntityManager.findOrCreate", () => {
...
+    const a2 = await em.findOrCreate(Author, { firstName: " a1 " }, {});
+    expect(a2).toMatchEntity(a1);
+  });
+
+  it("coalesces empty string", async () => {
+    const em = newEntityManager();
+    const a1 = em.create(Author, { firstName: "a1" });
+    await em.flush();
+    const a2 = await em.findOrCreate(Author, { firstName: "a1", lastName: "" }, {});
+    expect(a2).toMatchEntity(a1);
```
stephenh, line 351: @chr1sjf0x here are test cases for what you pointed out; ty!

PR #1665, opened 2025-11-14
feat: add plugin to sync data between ems during run invocations

zgavin:


packages/orm/src/relations/ManyToManyCollection.ts
```diff
@@ -198,7 +198,7 @@ export class ManyToManyCollection<T extends Entity, U extends Entity>
     ensureNotDeleted(this.entity, "pending");
     if (this.#loaded === undefined) {
       // This should only be callable in the type system if we've already resolved this to an instance
-      throw new Error("get was called when not loaded");
+      throw new Error(`${this.entity}.${this.fieldName}.get was called when not loaded`);
```
zgavin, line 201: note: just felt this was a helpful change for debugging downstream code

packages/orm/src/EntityManager.ts
```diff
@@ -340,6 +343,18 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
...
+
+      clearDataloaders() {
+        em.#dataloaders = {};
+      },
+
+      clearPreloadedRelations() {
+        em.#preloadedRelations = new Map();
+      },
+
+      setIsRefreshing(isRefreshing: boolean) {
```
zgavin, line 355: note: I left this in since I thought it might be useful.  A previous iteration of this branch needed it, but the current version no longer does.

packages/orm/src/EntityManager.ts
```diff
@@ -2214,11 +2232,18 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
...
+    hint: H | undefined = undefined,
+    opts: { normalizedHint?: H; allowDirty?: boolean } = {},
   ): L {
-    if (source.isNewEntity) fail("cannot import new entities");
+    let { normalizedHint, allowDirty = false } = opts;
+
+    if (!source.idMaybe) fail("cannot import new entities");
     if (source.isDeletedEntity) fail("cannot import deleted entities");
-    if (source.isDirtyEntity) fail("cannot import dirty entities");
+    if (!allowDirty && source.isDirtyEntity) fail("cannot import dirty entities");
```
stephenh: Maybe in-source comment something like "To support the refreshPlugin, we support importing technically-dirty entities, but later will make sure to copy the originalData, i.e. the flushed-to-the-database data, and not the wip values".

packages/orm/src/EntityManager.ts
```diff
@@ -2234,7 +2259,7 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
 
     if (!result) {
       const meta = getMetadata(source);
-      const row = createRowFromEntityData(source);
+      const row = createRowFromEntityData(source, !allowDirty);
```
stephenh: The negation makes make brain feel like it needs an in-source comment ... afaiu usually we prefer the `originalData` values, when importing, but if `allowDirty` is on, we're being used by the refreshPlugin, so set ...

* when allowDirty=false, preferOriginal=true, I get that...
* when allowDirty=true, preferOriginal=false, I thought if we're importing a dirty entity, for refreshPlugin, wouldn't we still want preferOriginal?

packages/orm/src/PluginManager.ts
```diff
@@ -35,23 +35,33 @@ interface PluginMethods {
    * @param rows The raw database rows returned from the query
    */
   afterFind?(meta: EntityMetadata, operation: FindOperation, rows: any[]): void;
-}
 
-const pluginMethods = ["beforeSetField", "beforeFind", "afterFind"] as (keyof PluginMethods)[];
+  afterWrite(entityTodos: Record<string, Todo>, joinRowTodos: Record<string, JoinRowTodo>): void;
```
stephenh, line 39: Any reason for `afterWrite` instead of `beforeWrite`?

Wondering if maybe `beforeWrite` would end up being usable for things like auth / auditing / etc sort of things, if for the refreshPlugin it doesn't matter which.

I suppose an `afterWrite` could still blow up the transaction if it wanted / saw something the plugin thought as a mistake/shouldn't be allowed?

packages/orm/src/PluginManager.ts
```diff
@@ -35,23 +35,33 @@ interface PluginMethods {
...
  * Base class for plugins that hook into entity lifecycle events.
  *
  * Plugins can implement any of the PluginMethods to intercept entity operations
  * and are automatically registered with the EntityManager when added via PluginManager.
  */
 export abstract class Plugin {
-  private [emSymbol]: EntityManager | undefined;
+  // We need to store the EntityManager references in a WeakRef array so that they can be garbage collected.  If we
+  // didn't do this, then multiple EMs with the same plugin instance could keep each other from being garbage collected.
+  [emsSymbol]: WeakRef<EntityManager>[] = [];
```
stephenh, line 53: Not against the weakref, but surely the GC would realize the "two EMs pointing at each other through a plugin" are a self-contained cycle, not reaching from any of the GC roots, and still collect them?

zgavin: I more meant that as long as 1 em had a strong ref to it from somewhere, the other couldn't be GCd even if it was otherwise unused

packages/orm/src/EntityManager.ts
zgavin: i actually think that the plugin should just call `createRowFromEntityData` directly instead of using import and hydrate itself.  Then we wouldn't need the `allowDirty` at all or special behavior in import.

>  I thought if we're importing a dirty entity, for refreshPlugin, wouldn't we still want preferOriginal

We very explicitly do not want the original data since that's whats already in the original em.  We want the changes, which is what is in just `data`

packages/orm/src/PluginManager.ts
zgavin: imo if you're doing auth at the write phase it's already too late :sweat_smile:

packages/orm/src/PluginManager.ts
stephenh: I wonder if we should clone each plugin for the new em...

Just thinking that, if I was writing a plugin, I would probably use instance-level state with the general assumption that "all of the instance data / entities are from 'my one EM'" and then get surprised when that's not the case. 🤔

stephenh: Maybe `shouldCopy` instead of a boolean could be renamed to like "maybeCloneForNewEm" that returns `undefined` | this | copy-if-they-want? (I don't like that name necessarily, just the concept)

packages/orm/src/serde.ts
```diff
@@ -526,7 +530,9 @@ export class SuperstructSerde implements FieldSerde {
     return JSON.stringify(data[this.fieldName]);
   }
 
-  // JSON is returned by postgres already parsed, so if we are trying to recreate that we don't need to stringify
+  // JSON is returned by postgres already parsed, so we should just be able to return our data directly. Unlike with
```
stephenh, line 533: A fwiw from when I was working on postgres.js things, the data for literally everything comes over the wire as strings (kinda surprising imo 🤔 🤷 ), and then the driver like node-pg does `Number(...)` / `JSON.parse` / etc based on its internal + user config.

packages/orm/src/PluginManager.ts
zgavin: That sgtm, but I think it'd need to be opt in rather than opt out if we did that?  The base plugin class wouldn't know how to instantiate subclasses.

packages/orm/src/serde.ts
zgavin: that seems kinda surprising, you'd think it'd send just the raw bytes for efficiency.  You're sure that's over the wire and not just a postgres.js thing?

packages/test-utils/src/RunPlugin.ts
```diff
@@ -0,0 +1,326 @@
...
+  if (otherSide.isPreloaded) return;
+  const otherEntities =
+    otherSide instanceof OneToManyCollection
+      ? em
+          .getEntities(meta.cstr)
+          .values()
+          .filter((e) => !e.isDeletedEntity)
+          .filter((e) => ((e as any)[otherSide.otherFieldName] as ConcreteReference).idTaggedMaybe === other.idTagged)
+          .toArray()
+      : remove
```
stephenh: Maybe in-source this is a o2o

packages/test-utils/src/RunPlugin.ts
```diff
@@ -0,0 +1,326 @@
...
+  ManyToOneReferenceImpl,
+  OneToManyCollection,
+  OneToOneReferenceImpl,
+  PolymorphicKeySerde,
+  PolymorphicReferenceImpl,
+  ReactiveReferenceImpl,
+} from "joist-orm";
+import { EntityManager, getEmInternalApi } from "joist-orm/build/EntityManager";
+import { JoinRow } from "joist-orm/build/JoinRows";
+import { Plugin } from "joist-orm/build/PluginManager";
```
stephenh: Maybe export

stephenh (approved):

PR #1669, opened 2025-11-18
feat: make mapResultToOriginalEm synchronous

zgavin:


packages/test-utils/src/run.ts
```diff
@@ -68,47 +68,26 @@ export function newContext<C extends Context>(ctx: C): C {
   return newCtx;
 }
 
-function gatherEntities(result: any): Entity[] {
+function mapResultToOriginalEm<R>(em: EntityManager, result: R): R {
   if (isEntity(result)) {
-    return [result];
+    return (em.findExistingInstance(result.idTagged) ?? fail(`Could not find entity ${result.idTagged}`)) as R;
```
stephenh, line 75: It looks like previously we had an `em.load`, which was driving the `async` -- but my assumption is that we can actually just assume the original EM has already em.refresh/refreshPlugin-d any entity we'd be asked to map?

If so, maybe add an in-source comment for this rationale/assumption.

packages/test-utils/src/run.ts
```diff
@@ -68,47 +68,26 @@ export function newContext<C extends Context>(ctx: C): C {
   return newCtx;
 }
 
-function gatherEntities(result: any): Entity[] {
+function mapResultToOriginalEm<R>(em: EntityManager, result: R): R {
```
stephenh, line 73: Do we have any callers to `mapResultToOriginalEm` that could be updated now that its not async?

packages/test-utils/src/run.ts
zgavin: yeah that's the assumption, since the only place `mapResultToOriginalEm` is called is after `RunPlugin` is installed.

I'm actually not sure the `em.load` was even necessary before, since `em.refresh()` would put any new entities into the original em anyways.

packages/test-utils/src/run.ts
zgavin: no, there's 2 callers and they're both just `return mapResultToOriginalEm` so there was no explicit `await` for them.

PR #1672, opened 2025-11-21
fix: Use more type imports.

stephenh:
This doesn't actually change (positive or negative) import types in large domain models, but seems stylistically preferable since I have it working anyway.


packages/codegen/src/generateEntityCodegenFile.ts
```diff
@@ -134,7 +134,7 @@ export function generateEntityCodegenFile(config: Config, dbMeta: DbMetadata, me
   // Set up the codegen artifacts to extend from the base type if necessary
   const baseEntity = dbMeta.entities.find((e) => e.name === meta.baseClassName);
   const subEntities = dbMeta.entities.filter((e) => e.baseClassName === meta.name);
-  const base = baseEntity?.entity.type ?? code`${BaseEntity}<${EntityManager}, ${idType}>`;
+  const base = baseEntity?.entity.typeSymbol ?? code`${BaseEntity}<${EntityManager}, ${idType}>`;
```
stephenh, line 137: For `Author extends AuthorCodegen` we can't have the type import.

PR #1673, opened 2025-11-21
fix: Import/export fixes.

stephenh:


packages/drivers/bun-pg/src/BunPgDriver.ts
```diff
@@ -3,22 +3,24 @@ import {
...
   ParsedFindQuery,
   SequenceIdAssigner,
+  Todo,
 } from "joist-orm";
-import { JoinRowTodo, Todo } from "joist-orm/build/Todo";
-import { buildValuesCte } from "joist-orm/build/dataloaders/findDataLoader";
-import { DeleteOp, generateOps, InsertOp, UpdateOp } from "joist-orm/build/drivers/EntityWriter";
-import { buildCteSql } from "joist-orm/build/drivers/buildRawQuery";
-import { getRuntimeConfig } from "joist-orm/build/runtimeConfig";
-import { batched, cleanSql } from "joist-orm/build/utils";
```
stephenh, line 21: This file was one of the biggest offenders of `joist-orm/build...` imports.

I didn't want to export all of these, so hide some of them behind a `driverApi` const -- at least for the functions, I couldn't figure out how to do the same sort of thing for types, but figured it didn't really matter atm.

It might also be easier for this driver to just live in the core `joist-orm` package anyway 🤔

packages/orm/src/relations/index.ts
```diff
@@ -37,7 +37,12 @@ export {
 export { ReactiveField, hasReactiveField, isReactiveField } from "./ReactiveField";
 export { ReactiveGetter, hasReactiveGetter, isReactiveGetter } from "./ReactiveGetter";
 export { hasReactiveQueryField, isReactiveQueryField } from "./ReactiveQueryField";
-export { ReactiveReference, ReactiveReferenceImpl, hasReactiveReference } from "./ReactiveReference";
+export {
+  ReactiveReference,
+  ReactiveReferenceImpl,
+  hasReactiveReference,
+  isReactiveReference,
```
stephenh, line 44: `isReactiveReference` was missing

packages/orm/src/index.ts
```diff
@@ -36,10 +36,18 @@ export { InstanceData } from "./InstanceData";
...
+export { visitConditions } from "./QueryVisitor";
 export { JoinRowTodo, Todo } from "./Todo";
 export * from "./changes";
 export { ConfigApi, EntityHook, resetBootFlag } from "./config";
-export { configureMetadata, getConstructorFromTaggedId, maybeGetConstructorFromReference } from "./configure";
+export {
+  configureMetadata,
+  getConstructorFromTaggedId,
+  getMetadataForTable,
+  getMetadataForType,
```
stephenh, line 47: The `getMetadataFor(table|type)` are new ... a few places in graphql-service use them.

PR #1687, opened 2026-01-03
feat: Reactive m2m collections.

stephenh:
Adds a `hasReactiveCollection` that is similar to `hasReactiveField` or `hasReactiveReference`, but for m2m collections.

This means m2m collections that are derived can be declared with a reactive hint, and Joist will track/recalc the collection just like the other field/reference reactables:

```
  /** Example of a ReactiveCollection - a derived m2m that auto-calculates its membership. */
  readonly bestReviews: ReactiveCollection<Author, BookReview> = hasReactiveCollection(
    { books: { reviews: "rating" } },
    (a) => {
      a.transientFields.bestReviewsCalcInvoked++;
      return a.books.get.flatMap((b) => b.reviews.get).filter((r) => r.rating >= 5);
    },
  );
```

### DIFF
--- a/packages/test-utils/src/RunPlugin.ts
+++ b/packages/test-utils/src/RunPlugin.ts
@@ -19,6 +19,8 @@ import {
   Plugin,
   PolymorphicKeySerde,
   PolymorphicReferenceImpl,
+  ReactiveManyToManyImpl,
+  ReactiveManyToManyOtherSideImpl,
   ReactiveReferenceImpl,
   Todo,
 } from "joist-orm";
@@ -325,10 +327,21 @@ function getReferences(entity: Entity): IteratorObject<ConcreteReference> {
     );
 }
 
-function getManyToManys(entity: Entity): IteratorObject<ManyToManyCollection<Entity, Entity>> {
+function getManyToManys(
+  entity: Entity,
+): IteratorObject<
+  | ManyToManyCollection<Entity, Entity>
+  | ReactiveManyToManyImpl<any, any, any>
+  | ReactiveManyToManyOtherSideImpl<any, any>
+> {
   return getRelations(entity)
     .values()
-    .filter((r) => r instanceof ManyToManyCollection);
+    .filter(
+      (r) =>
+        r instanceof ManyToManyCollection ||
+        r instanceof ReactiveManyToManyImpl ||
+        r instanceof ReactiveManyToManyOtherSideImpl,
+    );
 }
 
 function getCollections(entity: Entity) {
@@ -336,6 +349,10 @@ function getCollections(entity: Entity) {
     .values()
     .filter(
       (r) =>
-        r instanceof ManyToManyCollection || r instanceof OneToManyCollection || r instanceof OneToOneReferenceImpl,
+        r instanceof ManyToManyCollection ||
+        r instanceof OneToManyCollection ||
+        r instanceof OneToOneReferenceImpl ||
+        r instanceof ReactiveManyToManyImpl ||
+        r instanceof ReactiveManyToManyOtherSideImpl,
     );
 }


packages/codegen/src/generateEntityCodegenFile.ts
```diff
@@ -473,9 +476,11 @@ function generateOptsFields(meta: EntityDbMetadata): Code[] {
...
     return code`${fieldName}?: ${otherEntity.type}[];`;
   });
-  const m2m = meta.manyToManys.map(({ fieldName, otherEntity }) => {
-    return code`${fieldName}?: ${otherEntity.type}[];`;
-  });
+  const m2m = meta.manyToManys
+    .filter(({ derived }) => !derived) // Skip ReactiveCollections
+    .map(({ fieldName, otherEntity }) => {
+      return code`${fieldName}?: ${otherEntity.type}[];`;
+    });
```
zgavin, line 483: nit: seems like this could be 1 line now that there's a filter in front of it so it wraps regardless.  ditto for the similar snippets below

packages/codegen/src/generateEntityCodegenFile.ts
```diff
@@ -1016,6 +1023,15 @@ function createRelations(config: Config, meta: EntityDbMetadata, entity: Entity)
   // Add ManyToMany
   const m2m: Relation[] = meta.manyToManys.map((m2m) => {
     const { joinTableName, fieldName, columnName, otherEntity, otherFieldName, otherColumnName } = m2m;
+    if (m2m.derived === "async") {
```
zgavin, line 1026: suggestion: `switch` on  `m2m.derived` maybe?

packages/orm/src/relations/ReactiveCollection.ts
```diff
@@ -0,0 +1,277 @@
...
+  fn: (entity: Reacted<T, H>) => readonly MaybeReactedEntity<U>[],
+): ReactiveCollection<T, U> {
+  return lazyField((entity: T, fieldName) => {
+    const m2m = getMetadata(entity).allFields[fieldName] as ManyToManyField;
+    return new ReactiveCollectionImpl<T, U, H>(entity, m2m, hint, fn);
+  });
+}
+
+/** Implements ReactiveCollection, initially a copy/paste of ReactiveReference but for m2m. */
+export class ReactiveCollectionImpl<T extends Entity, U extends Entity, H extends ReactiveHint<T>>
```
zgavin: thought: I find the naming slightly confusing here, since this is only for m2m collections and not just a way to generically define any collection reactively.  Part of this is that there no "other side" for reactive references, which I would otherwise naively expect to be `ReactiveCollectionOtherSide`.  I'd prefer if the naming explicitly included `ManyToMany`.

packages/orm/src/relations/ReactiveCollection.ts
```diff
@@ -0,0 +1,277 @@
...
+
+  [RelationT]: T = null!;
+  [RelationU]: U = null!;
+}
+
+/** Type guard utility for determining if an entity field is a ReactiveCollection. */
+export function isReactiveCollection(
+  maybeReactiveCollection: any,
+): maybeReactiveCollection is ReactiveCollection<any, any> {
+  return maybeReactiveCollection instanceof ReactiveCollectionImpl;
```
zgavin: chore: we need to also update `isCollection` to have this check

packages/orm/src/relations/ReactiveManyToManyOtherSide.ts
```diff
@@ -0,0 +1,198 @@
...
+    return this.#doGet({ withDeleted: true });
+  }
+
+  set(): void {
+    fail(`Cannot set ${this.entity}.${this.fieldName} - it is the read-only other side of a ReactiveCollection.`);
+  }
+
+  setFromOpts(_others: U[]): void {
+    throw new Error(`ReactiveCollectionOtherSide ${this.entity}.${this.fieldName} cannot be set via opts`);
+  }
```
zgavin, line 93: nit: `fail` then `throw new Error` immediately after 🙃

packages/orm/src/relations/ReactiveManyToManyOtherSide.ts
```diff
@@ -0,0 +1,198 @@
...
+    return this.#field.columnNames[0];
+  }
+
+  get otherColumnName(): string {
+    return this.#field.columnNames[1];
+  }
+
+  get otherFieldName(): string {
+    return this.#field.otherFieldName;
+  }
```
zgavin, line 133: thought: having to redefine these on each relation makes me miss ruby mixins

packages/orm/src/configure.ts
```diff
@@ -245,18 +245,20 @@ function reverseIndexReactivity(metas: EntityMetadata[]): void {
       }
     }
 
-    // Look for ReactiveFields to reverse
-    const reactiveFields = Object.values(meta.allFields).filter(
+    // Look for ReactiveFields, ReactiveReferences, & ReactiveCollections to reverse
+    const reactiveRelations = Object.values(meta.allFields).filter(
```
zgavin, line 249: thought: actually feel like the previous naming was more accurate.  Relations are a subset of fields. Granted, that would also be somewhat confusing since "reactive field" is also a concrete type 🤔   Also it's confusing that the naming change doesn't propagate to the for loop variable.

packages/orm/src/relations/ReactiveReference.ts
```diff
@@ -377,11 +382,11 @@ export class ReactiveReferenceImpl<
   }
 
   public get hasBeenSet(): boolean {
-    return false;
+    return true;
   }
 
   public toString(): string {
-    return `PersistedAsyncReference(entity: ${this.entity}, hint: ${this.loadHint}, fieldName: ${this.fieldName}, otherMeta: {
+    return `ReactiveReference(entity: ${this.entity}, hint: ${this.loadHint}, fieldName: ${this.fieldName}, otherMeta: {
```
zgavin, line 389: praise: 🤯

packages/orm/src/relations/ReactiveReference.ts
```diff
@@ -377,11 +382,11 @@ export class ReactiveReferenceImpl<
   }
 
   public get hasBeenSet(): boolean {
-    return false;
+    return true;
```
zgavin, line 385: question: what did this change address?

packages/codegen/src/generateEntityCodegenFile.ts
```diff
@@ -530,7 +535,7 @@ function generateFieldsType(meta: EntityDbMetadata, idType: "string" | "number")
 // in the partial type and of course the caller will only be setting one.
 function generateOptIdsFields(meta: EntityDbMetadata): Code[] {
   const m2o = meta.manyToOnes
-    .filter(({ derived }) => !derived)
+    .filter(({ derived }) => !derived) // Skip ReactiveCollections
```
zgavin: thought: continuing the previous though about naming, this comment seems to agree with my naive assumption that the other side of a `ReactiveReference` ought to be a `ReactiveCollection`.

packages/orm/src/relations/ReactiveCollection.ts
```diff
@@ -0,0 +1,277 @@
...
+    const meta = getMetadata(this.entity);
+    return (meta.config.__data.cachedReactiveLoadHints[this.fieldName] ??= convertToLoadHint(meta, this.reactiveHint));
+  }
+
+  setFromOpts(_others: U[]): void {
+    throw new Error(`ReactiveCollection ${this.entity}.${this.fieldName} cannot be set via opts`);
+  }
+
+  maybeCascadeDelete(): void {
+    // ReactiveCollections don't cascade delete
```
zgavin: question: why do they not cascade delete?  It seems like it'd be safe for the other side not to cascade delete, but it seems like the main side would need to tell its join rows to go away when it does?  Or is that handled somewhere else in the cascade delete behavior (iirc we still do immediate and on flush cascade delete passes)?

imo would be good to update the comment here with whatever the explanation is

packages/orm/src/relations/ReactiveManyToMany.ts
```diff
@@ -0,0 +1,277 @@
...
+
+  public get hasBeenSet(): boolean {
+    return true;
+  }
+
+  get isPreloaded(): boolean {
+    return false;
+  }
+
+  preload(): void {
```
zgavin, line 233: issue: we need to implement `import` in order for `fork` / `importEntity` to work.  Ideally preload/import would be merged into a single behavior, but we aren't there yet so we do need to implement `import`.

packages/orm/src/relations/ReactiveManyToManyOtherSide.ts
```diff
@@ -0,0 +1,198 @@
...
+
+  get hasBeenSet(): boolean {
+    return false;
+  }
+
+  get isPreloaded(): boolean {
+    return false;
+  }
+
+  preload(): void {
```
zgavin, line 143: issue: ditto about `import`

packages/orm/src/relations/ReactiveCollectionOtherSide.ts
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  [RelationT]: T = null!;
+  [RelationU]: U = null!;
+}
+
+export function isReactiveCollectionOtherSide(
+  maybeCollection: any,
+): maybeCollection is ReactiveCollectionOtherSide<any, any> {
+  return maybeCollection instanceof ReactiveCollectionOtherSideImpl;
```
zgavin: chore: ditto about `isCollection`

packages/tests/integration/src/entities/Author.ts
```diff
@@ -274,6 +277,15 @@ export class Author extends AuthorCodegen {
     },
   );
 
+  /** Example of a ReactiveCollection - a derived m2m that auto-calculates its membership. */
+  readonly bestReviews: ReactiveCollection<Author, BookReview> = hasReactiveCollection(
```
zgavin: thought: really pushing our toy model to the limits 😂

packages/tests/integration/src/relations/ReactiveCollection.test.ts
```diff
@@ -0,0 +1,308 @@
+import { Author, Book, BookReview, Entity, EntityManager, newAuthor, newBook, newBookReview } from "@src/entities";
+import { insertAuthor, insertAuthorToBestReview, insertBook, insertBookReview, select } from "@src/entities/inserts";
+import { newEntityManager } from "@src/testEm";
+
+describe("ReactiveCollection", () => {
```
zgavin: chore: probably should have tests for the new collections for `fork` / `importEntity`

packages/orm/src/relations/ReactiveCollection.ts
stephenh: Yeah, I thought about `ReactiveManyToMany` ... agreed that is better.

packages/orm/src/configure.ts
stephenh: Ah; I'd thought of fields & relations as siblings, i.e. fields are primitives & enums

packages/codegen/src/generateEntityCodegenFile.ts
stephenh: Surprised to see you suggest `switch` :sweat_smile: , but yeah added else ifs/assertNever to make it exhaustive.

packages/orm/src/relations/ReactiveCollection.ts
stephenh: :+1: added

packages/orm/src/relations/ReactiveReference.ts
stephenh: I was futzing with `ReactiveManyToMatch.hasBeenSet` which was auto-completed to false, and I was like "should it be false?" so looked at other impls, and think a `true` default for these reactive/auto-set relations makes more sense.

It's only used by `setDefault` to see if "things have been set yet", which I don't think we can use for reactive relations anyway, so its kind of a noop.

packages/orm/src/relations/ReactiveCollection.ts
stephenh: I was assuming m2ms never cascade, but I guess they do (conditionally); will copy/paste over the m2m impl:

```
    if (this.isCascadeDelete) {
      this.current({ withDeleted: true }).forEach((e) => this.entity.em.delete(e));
    }
```

packages/orm/src/relations/ReactiveManyToMany.ts
stephenh: ...yeah...I might merge this and do those as a follow-up. Scale of 1-10 (10 is most interested), how interested are you in doing said follow-up PR? :sweat_smile: Really np either way.

packages/orm/src/configure.ts
zgavin: well in the metadata they're all just in `fields` :shrug:

PR #1688, opened 2026-01-04
feat: Simplify hasMany, hasReactiveField, etc

stephenh:
Simplifies (mostly) the codegen output by removing now-unnecessary params to our `has...` methods:

image

---

Since the recently-introduced `lazyField` now knows the `fieldName`, we can use `meta.allFields[fieldName]` to get all of the info that previously we passed through the `has(...)` method.

Having the `OneToManyField` / `ManyToOneField` / etc in hand also means we can pass it into our `OneToManyImpl(...)` constructors, and reduce the number of instance fields they hold, from ~3-4 of "other field", "other column", "other meta" to "just the entity & the field".

And then we add back getters like `getOtherFieldName { return this.#field.otherFieldName }` to keep the API the same.

Most of the updates are "just re-codegen" b/c the majority of `has...` methods are called from `...Codegen` classes.

That said, this also fixes `hasReactiveField` & `hasReactiveReference` unnecessarily taking the field names, and removes those -- I didn't write a codemod for this, so it currently requires manual updates.

### DIFF
--- a/packages/codegen/src/generateEntityCodegenFile.ts
+++ b/packages/codegen/src/generateEntityCodegenFile.ts
@@ -92,11 +92,11 @@ export interface ColumnMetaData {
 // A local type just for tracking abstract vs. concrete relations
 type Relation =
   // I.e. `abstract ReactiveReference` that the user must implement
-  | { kind: "abstract"; line: Code }
+  | { kind: "abstract"; line: Code; comment?: string }
   // I.e. a `get author(): ManyToOne<...>`
-  | { kind: "concrete"; fieldName: string; decl: Code; init: Code }
+  | { kind: "concrete"; fieldName: string; decl: Code; init: Code; comment?: string }
   // I.e. a `get author(): { super.author as ... }`
-  | { kind: "super"; fieldName: string; decl: Code };
+  | { kind: "super"; fieldName: string; decl: Code; comment?: string };
 
 /** Creates the base class with the boilerplate annotations. */
 export function generateEntityCodegenFile(config: Config, dbMeta: DbMetadata, meta: EntityDbMetadata): Code {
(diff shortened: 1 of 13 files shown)


packages/orm/src/relations/ManyToManyCollection.ts
```diff
@@ -13,59 +13,37 @@ import {
...
 
 /** An alias for creating `ManyToManyCollections`s. */
-export function hasManyToMany<T extends Entity, U extends Entity>(
-  joinTableName: string,
-  columnName: string,
-  otherFieldName: keyof U & string,
-  otherColumnName: string,
-): Collection<T, U> {
-  let otherMeta: EntityMetadata<U>;
+export function hasManyToMany<T extends Entity, U extends Entity>(): Collection<T, U> {
```
zgavin, line 22: thought: I actually kind of liked having these arguments in the codegen because it was easy to reference when you needed the column/table information.  In particular I appreciated it for m2ms because the table names are generally adhoc (since nothing is derived from them we've never needed to enforce standards) and the collections themselves are often explicitly named.   To get this info now you'd need to go digging into the metadata which is more tedious.

Generally I really like the more simple definitions for the relation/field helpers.  Definitely better for the user-defined functions. But, maybe we could add this info back as jsdoc comments or something, at least for m2ms?

packages/orm/src/relations/ManyToManyCollection.ts
```diff
@@ -13,59 +13,37 @@ import {
...
+    return new ManyToManyCollection<T, U>(entity, m2m);
   });
 }
 
 export class ManyToManyCollection<T extends Entity, U extends Entity>
   extends AbstractRelationImpl<T, U[]>
   implements Collection<T, U>
 {
-  readonly #fieldName: keyof T & string;
+  readonly #field: ManyToManyField;
```
zgavin, line 33: praise: neat, this should make it a lot easier to do reflection on relations and have fewer `getMetadata` calls

packages/tests/integration/src/EntityManager.reactiveRules.test.ts
```diff
@@ -452,7 +452,7 @@ describe("EntityManager.reactiveRules", () => {
     expect(bRfs[i++]).toEqual({
       kind: "reaction",
       cstr: "Book",
-      name: "Book.ts:118",
+      name: "Book.ts:117",
```
zgavin, line 463: question/nit: why does this file have a `.tsx` extension?

stephenh: :shrug: fixed

packages/orm/src/relations/ManyToManyCollection.ts
stephenh: Sure; although my 1st attempt did not work :sweat_smile: 

image

PR #1689, opened 2026-01-04
feat: Closure table example.

stephenh:
Adds a closure table example:

```ts
  /** Example of a closure table. */
  readonly menteesClosure: ReactiveCollection<Author, Author> = hasReactiveCollection(
    "menteesRecursive",
    (a) => [a, ...a.menteesRecursive.get],
  );
```

This required getting recursiveChildren hints to be valid reactive hints, which was a ~slightly longer than anticipated tangent, but good to have done.

I tried to leave gratuitous comments.

### DIFF
--- a/packages/migration-utils/src/utils.ts
+++ b/packages/migration-utils/src/utils.ts
@@ -189,16 +189,20 @@ type ManyToManyColumn = {
   column?: string;
   /**
    * The name of the collection that *points to* these rows, i.e. for `books_to_tags` and the `book_id`
-   * column, this might be `taggedBooks`, i.e. the name of the collection within `Tag` that loads these
-   * books.
+   * column, `collectionName=taggedBooks` b/c `t1.taggedBooks` does a `SELECT book_id WHERE tag_id=t:1`.
    *
-   * For self-referential m2m tables, i.e. `author_to_mentors` m2m table, these names can be confusing, i.e.
-   * an `author_to_mentors.author_id` with a `collectionName=mentors` does not mean "my mentors are all rows
-   * where author_id=me", it means "the author_id entities are members of the `mentors` collection of other
-   * authors".
+   * For self-referential m2m tables, i.e. `author_to_mentors`, these names can be confusing, i.e.:
    *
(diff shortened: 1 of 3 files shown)


packages/test-utils/src/RunPlugin.ts
```diff
@@ -325,10 +327,21 @@ function getReferences(entity: Entity): IteratorObject<ConcreteReference> {
...
-    .filter((r) => r instanceof ManyToManyCollection);
+    .filter(
+      (r) =>
+        r instanceof ManyToManyCollection ||
+        r instanceof ReactiveCollectionImpl ||
+        r instanceof ReactiveCollectionOtherSideImpl,
+    );
 }
 
 function getCollections(entity: Entity) {
```
zgavin, line 347: chore: probably need to update this too? forgot about these helpers in the other PR

packages/tests/integration/src/entities/Author.ts
```diff
@@ -286,6 +297,11 @@ export class Author extends AuthorCodegen {
     },
   );
 
+  /** Example of a closure table. */
```
zgavin, line 299: question: what makes this a "closure" table? i guess what even is a "closure" table? :sweat_smile:

zgavin (approved):

packages/tests/integration/src/entities/Author.ts
stephenh: Yeah, it's a weird term, but it's a pattern for materializing hierarchical parent/child relations into "a flat set of rows" so that joins are easy:

https://fueled.com/blog/closure-table/

I'm hoping to eventually use them for "auth grants on a market inherit to all projects in the market" sort of thing.

packages/test-utils/src/RunPlugin.ts
stephenh: Ah yeah -- I can add -- fwiw I have no idea if ReactiveCollectionImpl/OtherSideImpl are going to work with `run`, my guess is no.

PR #1690, opened 2026-01-06
fix: Implement ReactiveManyToMany.import.

stephenh:


packages/orm/src/EntityManager.ts
```diff
@@ -2154,8 +2156,7 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
 
     function findEntity(oldEntity: Entity): Entity {
       if (oldEntity.isNewEntity) {
-        const index = parseInt(oldEntity.toString().split("#").pop()!) - 1;
-        return newEm.getEntities(oldEntity.constructor as EntityConstructor<Entity>)[index];
```
stephenh, line 2158: I had thought this `getEntities(cstr)[index]` was broken now w/the change to `toString`s that went from "index is em.entities.indexOf(this)" to "index is the em.create assigned instanceData.createId".

That said, maybe this was working? b/c a) the test I wrote to repro the failure did not actually fail, and b) somehow the `newEm.entities` had all of the new entities at the front, even if I `em.create`-d those entities *after* doing am em.find 🤔 .

Which I'm not sure how that re-ordering happened, but I think it would have actually made this index-based approach still work, i.e. the `createId`s from toString() still lined up with "index within getEntities" 🤔 

Either way I think teaching `getEntity` to take the tagged `a#1` ids is fine / more reliable & potentially useful for other things.

packages/test-utils/src/toMatchEntity.ts
```diff
@@ -88,7 +88,7 @@ export type MatchedEntity<T> =
         | {
             [K in keyof T]?: T[K] extends Reference<any, infer U, any>
               ? MatchedEntity<U> | U
-              : T[K] extends Collection<any, infer U>
+              : T[K] extends ReadOnlyCollection<any, infer U>
```
stephenh, line 91: Fixes a type error when readonly collections didn't work in `toMatchEntity`

packages/tests/integration/src/EntityManager.test.ts
```diff
@@ -1563,6 +1567,21 @@ describe("EntityManager", () => {
...
+      const em = newEntityManager();
+      // Move the book to a2
+      const a1 = await em.findOneOrFail(Author, {}, { populate: "books" });
+      const [b1] = a1.books.get;
+      const [a2, a3] = twoOf(() => newAuthor(em));
+      b1.author.set(a3);
+      const result = em.fork({ allowPendingChanges: true });
+      const b = result.getEntity("b:1") as Loaded<Book, "author">;
+      expect(b.author.get.toTaggedString()).toBe("a#2");
+      expect(result.entities).toMatchEntity(["a#1", "a#2", "a:1", "b:1"]);
```
stephenh, line 1582: These `a#1`, `a#2` being at the front is what I didn't expect, but had the code still work.

packages/orm/src/relations/ReactiveManyToMany.ts
```diff
@@ -234,6 +234,16 @@ export class ReactiveManyToManyImpl<T extends Entity, U extends Entity, H extend
     // No-op for now - could implement if needed
   }
 
+  import(other: ReactiveManyToManyImpl<T, U, H>, findEntity: (e: U) => U): void {
+    if (other.#loaded !== undefined) {
+      this.#loaded = other.#loaded.map(findEntity);
```
zgavin: maybe-chore: the other `import` collection implementations have a custom `map` function that was a performance improvement vs the built in `map`.   

I'm not really sure how that was the case, though, other than the custom implementation allocates the entire result array ahead of time.  I don't know why the built-in `map` wouldn't do that as well.  Maybe I just made that change at the same time I made other performance improvements and falsely attributed it to that pattern 🤔 

Also seems like it'd actually be faster if the `map` was just a helper function defined externally to the class that took `findEntity` as a callback rather than capturing it from the definition scope.

Regardless I think all the implementations should use the same approach or one of us will wander in and be confused why they're different later.

packages/orm/src/EntityManager.ts
zgavin: The re-ordering is from the loop above this.  New entities are created immediately in `newEm`, while  persisted entities are simply turned into a `row` which is added to an array and then all those rows are passed to `hydrate`.  So new entities end up at the start of the array in the `newEm`.

Fine with this change though.

packages/tests/integration/src/EntityManager.test.ts
```diff
@@ -1876,6 +1923,30 @@ describe("EntityManager", () => {
...
+      const a2 = em2.importEntity(a1, "bestReviews");
+      expect(a2.bestReviews.isLoaded).toBe(true);
+      expect((a2 as any).bestReviews.get).toMatchEntity([{ id: "br:1" }]);
+    });
+
+    it("can import an entity with a loaded reactive m2m (other side)", async () => {
+      await insertAuthor({ first_name: "a1" });
+      await insertBook({ author_id: 1, title: "b1" });
+      await insertBookReview({ book_id: 1, rating: 5 });
+      await insertAuthorToBestReview({ author_id: 1, book_review_id: 1 });
```
zgavin, line 1943: thought: it'd be neat if we could make these `insert...` methods chainable and then `await`ing the result would dispatch a single query.  Not really sure it matters since the test suite runs so quickly, but just an idea for the future 🤷 

eg, this could could become 


```typescript 
      await insertAuthor({ first_name: "a1" })
        .insertBook({ author_id: 1, title: "b1" })
        .insertBookReview({ book_id: 1, rating: 5 });
        .insertAuthorToBestReview({ author_id: 1, book_review_id: 1 });
```

packages/orm/src/EntityManager.ts
stephenh: Ah! Gotcha, yeah I missed that nuance, but makes sense now.

packages/tests/integration/src/EntityManager.test.ts
stephenh: Yeah 🤔 -- maybe something just different like:

```
await insert({
  authors: [...],
  books: [...],
});
```

...they would probably still turn into "send, wait, send, wait, send, wait" over the wire anyway, unless `insert` used the pipelining feature of pg / postegres.js to "simultaneously" send all the INSERTs at once.

packages/orm/src/relations/ReactiveManyToMany.ts
stephenh: Yeah, I was thinking of passing in both a `findEntity` + `findEntities` that pulled the optimized function up into a helper method in em.fork that all callers could use if they wanted.

Will do that.

PR #1692, opened 2026-01-10
feat: Use unnest for batched SQL

stephenh:
* Uses `unnest` for all INSERTs/UPDATEs/batched SELECTs, which creates a fixed number of parameters
* Brings back `unnest_2d_1d` (renamed to `unnest_arrays`), revamped to support rectangular-izing arrays
  * We used `unnest` in bulk statements ~awhile ago, but removed it b/c it couldn't be used with array columns
  * This reintroduces `unnest_2d_1d` & fixes the gotchas of "must be rectangular" and "how to update nullable columns" (see in-source docs)

About 9% faster on joist-benchmark (`joist_v2` here is just "this branch"):

image

There are 60 queries instead of 66 queries b/c the "batch by max 10k parameters" is no longer necessary.

We also send dramatically less SQL over the wire, 550kb (half a mb) vs. 83kb, albeit in pathological "tens of thousands of records" cases.


packages/codegen/src/generateMetadataFile.ts
```diff
@@ -115,7 +115,7 @@ function generateFields(config: Config, dbMetadata: EntityDbMetadata): Record<st
         serdeType = PrimitiveSerde;
       }
       serde = isArray
-        ? code`new ${serdeType}("${fieldName}", "${columnName}", "${columnType}[]", true)`
+        ? code`new ${serdeType}("${fieldName}", "${columnName}", "${columnType}[]", true, ${!p.notNull})`
```
stephenh, line 118: The `unnest_arrays` needs a hint whether the array column is nullable or not.

If it is nullable, we prefix an extra null value to indicate "actually set this value to null"

packages/orm/src/unnest.ts
```diff
@@ -0,0 +1,63 @@
...
+
+// Because postgres array-of-arrays must be rectangular, we fill all arrays up the same max size
+// to put on the wire, and then later strip the padded nulls during the unnest_arrays call.
+function fillArrayWithNulls(c: OpColumn, array: any[] | null, maxSize: number): any[] {
+  const wasNull = array === null;
+  const result = array ? [...array] : [];
+  const nullsToAdd = Math.max(0, maxSize - (array?.length ?? 0));
+  for (let i = 0; i < nullsToAdd; i++) result.push(null);
+  // Add our unset marker
+  if (c.isNullableArray) result.unshift(wasNull ? null : 1);
```
stephenh, line 61: This `: 1` will probably need to be column type aware at some point, but atm it is working fine...

packages/orm/src/drivers/PostgresDriver.ts
```diff
@@ -96,114 +96,28 @@ export class PostgresDriver implements Driver<Knex.Transaction> {
...
+    entityTodos: Record<string, Todo>,
+    joinRows: Record<string, JoinRowTodo>,
+  ): Promise<void> {
+    const txn = (em.txn ?? fail("Expected EntityManager.txn to be set")) as Knex.Transaction;
+    await this.#idAssigner.assignNewIds(entityTodos);
+    const ops = generateOps(entityTodos);
     // Do INSERTs+UPDATEs first so that we avoid DELETE cascades invalidating oplocks
     // See https://github.com/joist-orm/joist-orm/issues/591
-    // We want 10k params maximum per batch insert
-    const parameterLimit = 10_000;
```
stephenh, line 112: We don't need the param limit batching anymore, b/c we have a fixed # of query params now.

Going to assume we won't be updating tables with 10k columns 😅

packages/orm/src/drivers/PostgresDriver.ts
```diff
@@ -96,114 +96,28 @@ export class PostgresDriver implements Driver<Knex.Transaction> {
...
-
-    for (const del of ops.deletes) {
-      await batchDelete(knex, del);
-    }
-  }
-
-  async flushJoinTables(em: EntityManager, joinRows: Record<string, JoinRowTodo>): Promise<void> {
-    const knex = (em.txn ?? fail("Expected EntityManager.txn to be set")) as Knex.Transaction;
-    for (const [joinTableName, { m2m, newRows, deletedRows }] of Object.entries(joinRows)) {
-      if (newRows.length > 0) {
```
stephenh, line 145: This code is mostly the same, just pulled out into helper methods down below

PR #1697, opened 2026-01-12
fix: o(n^2) on clone complexity

allangsilva:
When cloning a large number of entities, em.clone has O(n²) complexity due to [this line](https://github.com/joist-orm/joist-orm/blob/main/packages/orm/src/EntityManager.ts#L963):

```ts
const existing = this.entities.find((e) => sameEntity(e, existingIdOrEntity));
```

This performs a linear scan through all entities for each clone operation. Since there's already a Map (#entitiesById) that provides O(1) lookup via findExistingInstance(), we can use that instead.

### DIFF
--- a/packages/orm/src/EntityManager.ts
+++ b/packages/orm/src/EntityManager.ts
@@ -947,7 +947,7 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
 
       return [entity, clone] as const;
     });
-    const entityToClone = new Map(clones);
+    const entityToClone: Map<EntityW, EntityW> = new Map(clones);
 
     // 3. Now mutate the m2o relations. We focus on only m2o's because they "own" the field/column,
     // and will drive percolation to keep the other-side o2m & o2o updated.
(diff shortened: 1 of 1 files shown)


packages/orm/src/EntityManager.ts
```diff
@@ -960,7 +960,8 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
         ) {
           // What's the existing entity? Have we cloned it?
           const existingIdOrEntity = getField(clone, fieldName);
-          const existing = this.entities.find((e) => sameEntity(e, existingIdOrEntity));
+          // Use O(1) lookup via #entitiesById Map instead of O(N) .find() scan
+          const existing = this.findExistingInstance(existingIdOrEntity) as Entity;
```
stephenh: @allangsilva my worry is that `getField(clone, fieldName)` might sometimes return a new entity, and not always an id.

I.e. if we do something like:

```
const a = new Author(); // no id yet
book.author.set(a);
getField(book, "author"); // probably returns Author and not a string
```

So that is probably why I used `sameEntity` here in the first place.

I think what we could do instead is something like:

```
const existingIdOrEntity = getField(clone, fieldName);
const existingId = existingIdOrEntity instanceof BaseEntity ? existingIdOrEntity.toTaggedString() : existingIdOrEntity;
const existing = em.getEntity(existingId);
```

That's a little cumbersome, but the `.toTaggedString()` will give us a value like `a#1` that `getEntity` will accept, even for new/not-yet-flushed entities.

...granted, doing an `em.clone` with new entities involved is probably rare, but would be good to handle this.

stephenh: ...actually if `existingIdOrEntity` is an entity, we don't need to do anything 😅 

```
const existingIdOrEntity = getField(clone, fieldName);
const existing = isEntity(existingIdOrEntity) ? existingIdOrEntity : em.getEntity(existingIdOrEntity);
```

allangsilva: That's a simpler solution! Just validated it works fine for the big volume of clones. Thanks 👏

PR #1701, opened 2026-01-15
feat: allow run calls to execute in parallel

zgavin:


packages/orm/src/EntityManager.ts
```diff
@@ -1499,7 +1500,16 @@ export class EntityManager<C = unknown, Entity extends EntityW = EntityW, TX ext
    *
    * It returns entities that have changed (an entity is considered changed if it has been deleted, inserted, or updated)
    */
-  async flush(flushOptions: FlushOptions = {}): Promise<Entity[]> {
+  flush(flushOptions: FlushOptions = {}): Promise<Entity[]> {
+    const api = getEmInternalApi(this);
+    api.flushPromise = this.#flush(flushOptions).then((result) => {
```
stephenh: Should we fail here if `flushPromise` is already set? Or do we assume `#flush` will fail when the flush lock doesn't succeed?

Feels a little weird to always over-write the existing `flushPromise`

packages/orm/src/EntityManager.ts
```diff
@@ -2441,6 +2451,7 @@ export interface EntityManagerInternalApi {
   clearDataloaders(): void;
   clearPreloadedRelations(): void;
   setIsRefreshing(isRefreshing: boolean): void;
+  flushPromise: Promise<any> | undefined;
```
stephenh: I'll probably forget what this is here, maybe:

```
// Allow `run` calls that are invoked in parallel to share a single em.flush
```

packages/test-utils/src/run.ts
```diff
@@ -14,8 +14,8 @@ export async function run<C extends Context, T>(
   contextFn: ContextFn<C> = newContext,
 ): Promise<T> {
   const { em } = ctx;
-  // Ensure any test data we've setup is flushed
-  await em.flush();
+  // Ensure any test data we've setup is flushed.  If the em is already flushing, just wait on that promise.
```
stephenh: Could we store the `em.flush` promise as a module const here in `run`? That would make me feel better than touching `em.flush` for this.

...that said, maybe it needs to be "per em" and not a module-level const in case multiple tests (w/separate ems/dbs) are running all at once? Probably...

packages/orm/src/EntityManager.ts
zgavin: The latter.  Fine with failing here, though, too.

packages/test-utils/src/run.ts
stephenh: Maybe we could store it on `ctx`...

packages/orm/src/EntityManager.ts
stephenh: Cool, yeah, I think an explicit `flush is progress` immediate error here would be nice.

packages/test-utils/src/run.ts
zgavin: great suggestion.  Could use a `WeakMap` to still make it per `em` :upside_down_face:

zgavin: > ...that said, maybe it needs to be "per em" and not a module-level const in case multiple tests (w/separate ems/dbs) are running all at once? Probably...

If multiple tests are running they'd be in separate processes, so i think that'd actually be safe.  Only one test is actually running per process at a time.

stephenh (approved):

PR #1734, opened 2026-02-16
feat!: Replace Knex with node-pg

stephenh:
## Summary

This PR removes the Knex dependency from `PostgresDriver` and replaces it with direct usage of the native `pg` library. The driver now manages its own connection pool and transaction lifecycle instead of delegating to Knex.

## Key Changes

- **Removed Knex dependency**: `PostgresDriver` constructor now accepts a `pg.Pool` instead of a `Knex` instance
- **Direct pg usage**: Replaced all `knex.raw()` calls with `client.query()` calls using native pg
- **Manual transaction management**: Implemented explicit `BEGIN`/`COMMIT`/`ROLLBACK` instead of relying on Knex's transaction wrapper
- **Parameter conversion**: Added `toPgParams()` utility function to convert `?` placeholders to pg's native `$1, $2, ...` numbered parameters
- **Connection pooling**: Driver now directly manages the pg connection pool lifecycle
- **Updated test infrastructure**: All test setup files updated to create and manage `pg.Pool` instances alongside Knex (Knex is retained for test utilities like `flush_database()`)

## Implementation Details

- The `transaction()` method now explicitly calls `BEGIN`, `COMMIT`, and `ROLLBACK` on the client, with proper error handling and cleanup via `client.release()`
- All batch operations (`batchInsert`, `batchUpdate`, `batchDelete`, `m2mBatchInsert`, `m2mBatchDelete`) updated to use `pg.PoolClient` instead of `Knex.Transaction`
- The `getMaybeInTxnClient()` method replaces `getMaybeInTxnKnex()` to return either the transaction client or the pool
- Knex is still used in test infrastructure for convenience utilities but is no longer a core dependency of the driver itself

https://claude.ai/code/session_01FLm22XP3wqDHuzn4R36peW


packages/core/src/EntityGraphQLFilter.ts
```diff
@@ -97,8 +97,7 @@ export const opToFn: Record<Exclude<Operator, "in" | "nin" | "between">, string>
   containedBy: "<@",
   // containsSome / hasSome
   overlaps: "&&",
-  // Kinda weird, but escape the operator for knex
-  jsonPathExists: "@\\?",
```
stephenh, line 101: The `toPgParams` method below is smart enough to not need the `\\` anymore

PR #1750, opened 2026-02-27
fix: Fix m2o missing reactions.

stephenh:
There was a bug where:

* We set a `a.publisher=p2`
* em.flush
* The reaction runs and observes "publisher is now p2"
* A hook says "no, put publisher p1 back"
* setField calls RM.dequeue to see if we can dequeue the reaction
* ...the reaction is not told about the restored p1

And the RM was correctly trying to see "we already ran a reaction, so need to run it again to observe the restored value" but missed setting a state flag that meant we ended up skipping the queued reaction.


packages/core/src/ReactionsManager.ts
```diff
@@ -83,6 +83,7 @@ export class ReactionsManager {
           //   thinks we can dequeue the reactable
           // - but actually our reactable needs to be re-run with the restored value
           pending.todo.add(entity);
+          this.#needsRecalc[r.kind] = true;
```
stephenh, line 86: We missed setting the flag back to `true`, so then the ReactionsManager didn't realize it had work to do 😬

packages/tests/integration/src/entities/Author.ts
```diff
@@ -138,6 +139,7 @@ export class Author extends AuthorCodegen {
       afterMetadata: 0,
       runOnce: 0,
       observedNickNames: [] as string[],
+      observedPublishers: [] as any[],
```
stephenh, line 142: Added this to track "what did the reaction observe"

...I probably could have used the existing `observedNickNames` b/c this bug was not actually specific to m2os but 🤷

PR #1781, opened 2026-04-06
feat: Add hasAsyncQueryProperty for non-persisted queries.

stephenh:


docs/src/content/docs/modeling/relations.md
```diff
@@ -317,6 +317,45 @@ class BookReview extends PublisherCodegen {
...
+
+```typescript
+export class Publisher extends PublisherCodegen {
+  readonly numberOfAuthors: AsyncQueryProperty<Publisher, number> = hasAsyncQueryProperty((p) =>
+    p.em.findCount(Author, { publisher: p.id }),
+  );
+}
+```
+
+The value is cached until the next `em.flush`, at which point it is invalidated because the database state may have changed:
```
zgavin, line 334: thought: this seems like it would break loaded typing.  eg, a `Loaded<Publisher, "numberOfAuthors">` that is flushed would no longer be delivering on its type promise.  Not really sure what we could do about that, though.  Kind of feel like this is just a shade too cute, and we should just cache indefinitely and rely on either explicit `load` calls or have a `forceReload` option.

zgavin (approved):

docs/src/content/docs/modeling/relations.md
```diff
@@ -317,6 +317,45 @@ class BookReview extends PublisherCodegen {
 }
 ```
 
+### hasAsyncQueryProperty
+
+`hasAsyncQueryProperty` creates a derived value calculated from a SQL query, rather than from in-memory graph data. This is useful when pulling all the related data into memory would be too expensive, and you'd rather let the database do the work.
+
+Unlike `hasAsyncProperty`, there is no load hint or reactive hint — the lambda receives the entity directly and is expected to perform its own queries:
```
stephenh, line 324: @zgavin to your point about naming, we could rename:

* `hasAsyncProperty` -> `hasProperty`
* Frees up `hasAsyncQueryProperty` -> ... `hasAsyncProperty`
* Then maybe `hasReactiveQueryField` -> `hasAsyncReactiveField`?

Would probably do ^ as a separate PR

docs/src/content/docs/modeling/relations.md
stephenh: Yeah, I guess I could see:

* I populate some AsyncQueryPropertys
* I happen to em.flush during my `saveAuthors`
* I put the `author` on the wire and `AsyncQueryProperty` ... actually if it reports "not loaded" and the resolver call `.load()`, it would just recalc

I dunno, I go back/forth between "wow that's really annoying that I have to re-load it" and "wow that's really annoying I returned stale data".

Seems like the higher risk is returning stale data, so maybe go w/this until we have enough real-world usage to see if that was the wrong decision?

docs/src/content/docs/modeling/relations.md
zgavin: Yeah, that just seems like a much bigger breaking change :sweat_smile: 

Would we be renaming the underlying classes too? seems like it'd be weird to have `readonly foo:AsyncProperty<...> = hasProperty(...)` while we we might also have `readonly bar:AsyncQueryProperty<...> = hasAsyncProperty<...>` right beside it.

Also should add `hasAsyncMethod` to the list

stephenh: Fyi I'm also planning on adding `...Batched` flavors, i.e.:

* `hasAsyncBatchedProperty` -> the lambda takes a `mv[]` instead of the individual `mv` entities

And then same for `hasAsyncBatchedReactiveField`...

Seem good?

PR #1785, opened 2026-04-13
fix: Reactive hint subtype matching

henrysipp:
### Summary
- Fix `isTypeOrSubType` to walk an entity's base types when checking subtype compatibility.
- Export the helper and add coverage for concrete, base-type, sibling, and unrelated type cases.

### Test Plan
- `yarn jest src/reactiveHints.test.ts`

### DIFF
--- a/packages/core/src/reactiveHints.ts
+++ b/packages/core/src/reactiveHints.ts
@@ -4,7 +4,6 @@ import {
   EntityMetadata,
   getBaseAndSelfMetas,
   getMetadata,
-  getSubMetas,
   ManyToManyField,
   ManyToOneField,
   OneToManyField,
@@ -12,7 +11,6 @@ import {
   PolymorphicFieldComponent,
 } from "./EntityMetadata";
 import { Changes, FieldStatus, ManyToOneFieldStatus } from "./changes";
-import { getMetadataForType } from "./configure";
 import { isChangeableField } from "./fields";
 import { getProperties } from "./getProperties";
 import { Loadable, Loaded, LoadHint } from "./loadHints";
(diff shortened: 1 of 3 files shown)


packages/core/src/reactiveHints.ts
```diff
@@ -626,17 +624,15 @@ function maybeApplyTypeFilter(loadPromise: Promise<Entity | Entity[]>, viaType:
...
+ * a Publisher via inheritance.
+ */
+export function isTypeOrSubType(entity: Entity, typeName: string): boolean {
   const meta = getMetadata(entity);
-  // Easy check for the name is the same
-  if (meta.type === typeName) return true;
-  // Otherwise see if the entity is a subtype of the typeName, i.e. if our poly/type
-  // filter is `@Publisher`, and we're a `SmallPublisher`, that's valid to traverse.
-  for (const other of getSubMetas(getMetadataForType(typeName))) {
-    if (other.type === typeName) return true;
```
henrysipp, line 637: Naive change is to just make this line into `other.type == meta.type` but going bottom-up should be more efficient than this top-down loop.

packages/core/src/reactiveHints.ts
```diff
@@ -626,17 +624,15 @@ function maybeApplyTypeFilter(loadPromise: Promise<Entity | Entity[]>, viaType:
...
-  }
-  return false;
+  return meta.type === typeName || meta.baseTypes.some((b) => b.type === typeName);
```
henrysipp, line 635: Theoretically could be just `meta.baseType === ...` according to @zgavin since we only support single-level inheritance right now.

Though in that case baseTypes should only be a single element, and this is resistant to future changes.

packages/core/src/reactiveHints.ts
```diff
@@ -626,17 +624,15 @@ function maybeApplyTypeFilter(loadPromise: Promise<Entity | Entity[]>, viaType:
...
 
-/** Handle `viaType` filtering with subtype awareness. */
-function isTypeOrSubType(entity: Entity, typeName: string): boolean {
+/** Returns true if `entity` is exactly `typeName` or a subtype of `typeName`.
+ *
+ * Used to honor poly/type filters like `@Publisher` during reverse-hint walks:
+ * a `SmallPublisher` entity should pass an `@Publisher` filter because it is
+ * a Publisher via inheritance.
+ */
+export function isTypeOrSubType(entity: Entity, typeName: string): boolean {
```
henrysipp, line 633: Not a *massive* fan of exporting this, but the other methods of testing were significantly more involved in modifying our test harnesses.

packages/core/src/reactiveHints.ts
```diff
@@ -626,17 +624,15 @@ function maybeApplyTypeFilter(loadPromise: Promise<Entity | Entity[]>, viaType:
   return loadPromise;
 }
 
-/** Handle `viaType` filtering with subtype awareness. */
-function isTypeOrSubType(entity: Entity, typeName: string): boolean {
+/** Returns true if `entity` is exactly `typeName` or a subtype of `typeName`.
+ *
+ * Used to honor poly/type filters like `@Publisher` during reverse-hint walks:
+ * a `SmallPublisher` entity should pass an `@Publisher` filter because it is
+ * a Publisher via inheritance.
```
stephenh, line 631: 🎉 comment

packages/core/src/reactiveHints.ts
stephenh: 👍

stephenh (approved):

PR #1786, opened 2026-04-15
fix: stable m2m collection order across processes

allangsilva:
## Problem

`ManyToManyCollection` does not guarantee a stable item order across runs — the same code against the same database can return the same items in a different order. That is fine for most callers, but it shows up in snapshot-style comparisons (diffing GraphQL responses, approval tests, etc.) when an m2m relation is involved.

## Root cause

m2m rows are stored in a shared `JoinRows` index backed by a `Map`, which keeps insertion order. Rows get inserted in the order DataLoader batches run, and batch timing depends on JS promise scheduling — which is not deterministic across processes. So the final order can change between runs.

Example: a manual `INSERT` or edit-cycle in prod can produce join-row ids that are not in the same order as the target entity ids. When that happens, sorting by the join-row id no longer matches the target entity id order.

## Fix

Both loading paths now sort by the **target entity's id**:

1. **`manyToManyBatchLoader`** — sorts the result of `joinRows.getOthers(...)` before setting the preloaded relation. Uses `localeCompare(..., { numeric: true })` so tagged ids like `t:2` sort before `t:10`.
2. **`JsonAggregatePreloader`** — changes the `json_agg(... order by ...)` to use the target entity alias, matching the batch-loader's order (follow-up to #1784, which made it sort by join-row id).

### DIFF
--- a/packages/core/src/batchloaders/manyToManyBatchLoader.ts
+++ b/packages/core/src/batchloaders/manyToManyBatchLoader.ts
@@ -59,21 +59,9 @@ async function loadBatch<U extends Entity>(collection: ManyToManyLike, keys: str
   for (const [column, id] of tuples) {
     const entity = em.getEntity(id);
     if (!entity) continue;
-    // Sort by target-entity id so that callers see a stable order regardless of how many
-    // batches contributed rows to `JoinRows`. The underlying `Map` preserves insertion
-    // order, which reflects the sequence of batches — and that sequence depends on
-    // DataLoader timing, which is non-deterministic across processes. Using the target
-    // entity's id gives a stable, plan-independent order.
-    const others = (joinRows.getOthers(column, entity) as U[]).slice().sort(byIdMaybe);
+    const others = joinRows.getOthers(column, entity) as U[];
     // Determine the field name from the column
     const fieldName = column === collection.columnName ? collection.fieldName : collection.otherFieldName;
     api.setPreloadedRelation(entity.idTagged!, fieldName, others);
   }
-}
-
-/** Sorts entities by their tagged id with natural/numeric ordering (e.g. `t:2` before `t:10`). */
-function byIdMaybe(a: Entity, b: Entity): number {
-  const aId = String(a.idMaybe ?? "");
-  const bId = String(b.idMaybe ?? "");
-  return aId.localeCompare(bId, undefined, { numeric: true });
 }
--- a/packages/core/src/batchloaders/manyToManyBatchLoader.ts
+++ b/packages/core/src/batchloaders/manyToManyBatchLoader.ts
@@ -59,9 +59,21 @@ async function loadBatch<U extends Entity>(collection: ManyToManyLike, keys: str
   for (const [column, id] of tuples) {
     const entity = em.getEntity(id);
     if (!entity) continue;
-    const others = joinRows.getOthers(column, entity) as U[];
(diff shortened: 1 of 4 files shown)


packages/core/src/batchloaders/manyToManyBatchLoader.ts
```diff
@@ -59,9 +59,21 @@ async function loadBatch<U extends Entity>(collection: ManyToManyLike, keys: str
   for (const [column, id] of tuples) {
     const entity = em.getEntity(id);
     if (!entity) continue;
-    const others = joinRows.getOthers(column, entity) as U[];
+    // Sort by target-entity id so that callers see a stable order regardless of how many
+    // batches contributed rows to `JoinRows`. The underlying `Map` preserves insertion
+    // order, which reflects the sequence of batches — and that sequence depends on
+    // DataLoader timing, which is non-deterministic across processes. Using the target
+    // entity's id gives a stable, plan-independent order.
+    const others = (joinRows.getOthers(column, entity) as U[]).slice().sort(byIdMaybe);
```
stephenh: @allangsilva just curious, but does "order by m2m row id" or "order by target entity id" really matter? Or do we just need "well the order is at least deterministic"? 

I'm wondering if we could do the sorting within the `JoinRows`, maybe around here:

```
  /** Cheat and use the index to return "the other entities" for `entity`. */
  getOthers(columnName: string, entity: Entity): JoinRow[] {
    // It's empty m2m collection won't have any other entities, and so no index entries
    const map = this.indexes[columnName].get(entity);
    return map ? Array.from(map.values()) : [];
  }
```

I know with the JoinRows themselves, we won't have row.ids -- but maybe if they just come last, that would be fine?

packages/core/src/preloading/JsonAggregatePreloader.ts
```diff
@@ -213,12 +213,10 @@ function calcLateralJoins<I extends EntityOrId>(
...
-          // For m2m, order by the join-row's id so preloaded results match the
-          // order the lazy batch loader returns (which orders by join-row id).
-          // Otherwise, order by the target entity's id.
-          selects: [
-            `json_agg(json_build_array(${selects.join(", ")}) order by ${kq(m2mAlias ?? otherAlias)}.id) as _`,
-          ],
+          // Order by the target entity's id so preloaded results match the lazy batch
+          // loader's post-sort (see `manyToManyBatchLoader` which sorts `getOthers(...)`
+          // by target id to keep order stable across multi-batch populations of JoinRows).
+          selects: [`json_agg(json_build_array(${selects.join(", ")}) order by ${kq(otherAlias)}.id) as _`],
```
stephenh: Now that `JoinsRows.getOthers` is sorting by join-row id, should we undo this change?

packages/core/src/batchloaders/manyToManyBatchLoader.ts
allangaldinosilva: Good call @stephenh, we need the order _at least deterministic_, I moved the sort into `JoinRowIndex.getOthers` but it's still sorting by join-row id (new rows without id sort last). This worked for our use-case.

packages/core/src/preloading/JsonAggregatePreloader.ts
allangaldinosilva: yep, done 👍

stephenh (approved):

PR #1830, opened 2026-05-14
fix: Fix ReactiveReferences not being recalculated.

stephenh:
Specifically if:

* the RR was watching a o2o
* And the o2o was for a poly
* And the poly was set to a created-then-deleted entity


packages/core/src/relations/ReactiveReference.ts
```diff
@@ -184,6 +184,10 @@ export class ReactiveReferenceImpl<
         }
       }
     }
+    // The ReactiveField.load() always sets `#isCached = false`, but our assertion is that our
+    // IsLoadedCache/resetIsLoaded infra is good enough to "see all invalidating mutations",
+    // so we don't need this explicit reset.
+    // this.#isCached = false;
```
stephenh, line 190: LLM thought this was the fix, but it was just a bandaid -- real fix is down below

packages/core/src/InstanceData.ts
```diff
@@ -117,13 +117,22 @@ export class InstanceData {
...
   markDeleted(entity: Entity): boolean {
     if (this.#deleted === undefined) {
-      // Let any OneToManyCollection.get caches know that they should recalc (i.e. their filterDeleted logic)
-      // by asserting the `deleteBook.author` field is changing.
-      //
-      // Technically, the o2m caches use the "naive" cache, so we don't really need to pass along the field.
+      // Let any OneToManyCollection/ReactiveField/ReactiveReference.get caches know that they should recalc
+      // (i.e. their filterDeleted logic) by asserting the `deleteBook.author` field is changing.
       for (const field of Object.values(this.metadata.allFields)) {
-        if (field.kind === "m2o") {
```
stephenh, line 125: This `markDelete` only `ressetIsLoaded`-ing m2o fields was the real bug -- the `OptionPricingConfig.parent` is a `poly` so that's why this line skipped it, and its `#isCached` was not reset.

packages/core/src/changes.ts
```diff
@@ -95,9 +95,7 @@ class ManyToOneFieldStatusImpl<T extends Entity>
   /** Returns the original entity instance, which may not be loaded into memory yet, or undefined. */
   get originalEntity(): Promise<T | undefined> {
     const originalValue = this.originalValue;
-    if (isEntity(originalValue)) {
```
stephenh, line 98: The LLM asserts that `originalValue` is never an Entity, so this was safe to remove 🤔

PR #1868, opened 2026-06-03
fix: avoid recursive property populate deadlocks

### LINKED ISSUE #1489: Recursive dependency in hasAsyncProperty causes deadlock during batching
### Description:

We’re encountering a deadlock when a `hasAsyncProperty` field depends on a recursive relation, and that relation is also being loaded as part of the same GraphQL query. This seems to happen when both fields are batched together (e.g. by `em.populate`) and the async property accesses the recursive relation before it has been made available, resulting in a circular wait.

It can be reproduced with an async field like this:

```
readonly reputationScore: AsyncProperty<Author, number> = hasAsyncProperty(
  { mentor: "reputationScore" },
  (a) => a.localReputation ?? a.mentor.get?.reputationScore.get ?? 0,
);
```

And a query like:

```
query {
  authors(filter: { id: ["a1", "a2"] }) {
    id
    reputationScore
  }
}
```
Where: 
*  `mentor` is a recursive reference of `Author`
* `a2.mentor` is `a1`
* `a1` it's called by the loading hint and it's also returned in the query.

Poyraxx:
## Summary
- avoid recursive property promise cycles by letting SQL-backed relation batches finish before property-like fallback loads
- keep fallback relation loads batched in parallel so recursive relation/reaction paths still populate fully
- add a regression test that reproduces the recursive property deadlock with an in-memory author graph

Fixes #1489.

## Validation
- `yarn build`
- `yarn workspace joist-core test`
- local in-memory recursive property populate probe
- local in-memory recursive `menteesClosure` populate/mutation probe


stephenh:
@Poyraxx wow, this is a great PR, thank you!

The description of "SQL-backed relation batches finish before property-like fallback loads" sounds exactly right -- give me a chance to more thoroughly review, but at first glance everything looks great.

Ngl I'm kinda surprised you hit this, b/c we use Joist "a lot" and rarely/never hit it, except for that #1489 you found :thinking: that was from awhile ago -- so that is curious, but either way really great to have the fix! Thank you!

Poyraxx:
> @Poyraxx wow, this is a great PR, thank you!
> 
> The description of "SQL-backed relation batches finish before property-like fallback loads" sounds exactly right -- give me a chance to more thoroughly review, but at first glance everything looks great.
> 
> Ngl I'm kinda surprised you hit this, b/c we use Joist "a lot" and rarely/never hit it, except for that #1489 you found :thinking: that was from awhile ago -- so that is curious, but either way really great to have the fix! Thank you!
> 
> 

My dream job needed some SQL projects or contributions so i deep dived issues and codes then i noticed this issue is forgotten. I hope I have been able to help the project development, even a little.

packages/tests/integration/src/EntityManager.joins.test.ts
```diff
@@ -14,10 +14,34 @@ import {
 import { isPreloadingEnabled, newEntityManager, queries, resetQueryCount } from "@src/testEm";
 import { testing } from "joist-orm";
 import { jan1, jan2 } from "src/testDates";
-import { Author, Book, Critic, LargePublisher, Publisher } from "./entities";
+import { Author, Book, Critic, LargePublisher, Publisher, newAuthor } from "./entities";
 
 const { partitionHint } = testing;
 
+function attachRecursiveScore(author: Author): void {
```
stephenh: Instead of `Object.defineProperty`-ing this `recursiveScore` onto `Author`, can we just declare this as a real/regular field in the `Author.ts` file?

Assuming that still reproduces the issue, that would be much more "black box" oriented, and so be more robust to future refactorings/changes in how joist's relations / internals work.

packages/tests/integration/src/EntityManager.joins.test.ts
```diff
@@ -217,6 +241,21 @@ describe("EntityManager.joins", () => {
...
+    const a2 = newAuthor(em, { firstName: "a2", mentor: a1 }) as Author;
+    attachRecursiveScore(a1);
+    attachRecursiveScore(a2);
+    const result = await Promise.race([
+      em.populate([a1, a2], "reputationScore" as any).then(() => "done"),
+      new Promise((resolve) => setTimeout(() => resolve("timeout"), 1000)),
+    ]);
+    expect(result).toBe("done");
+    expect((a1 as any).reputationScore.get).toBe(0);
+    expect((a2 as any).reputationScore.get).toBe(0);
```
stephenh: If we move `reputationScore` into `Author.ts`, these these `as any`s should be able to go away

packages/core/src/batchloaders/populateBatchLoader.ts
```diff
@@ -151,16 +153,27 @@ export function populateBatchLoader(
               }
             }
           }
-          fallbackPromises.push(relation.load(opts) as Promise<any>);
+          const load = () => relation.load(opts) as Promise<any>;
```
stephenh: I'm a little nervous about creating a lambda for every single relation -- we've had some populates that end up having ~100-300k relations, and an extra 300k lambdas will had up in terms of perf/overhead.

Can we remove this `const load`, and instead have the two `const fallback...` arrays above be just `Array<Relation>`s, such that we still separate out "real SQL relations" from "derived relations".

But then below we can call just `relation.load()` directly instead of the lambda-ized wrapper.

stephenh (commented):
I had a chance to review the PR -- still looks great, but a few questions/asks -- thanks @Poyraxx !

packages/core/src/batchloaders/populateBatchLoader.ts
Poyraxx: i have made a some changes based on your comments thx for review btw

stephenh (approved):
This is great -- thank you @Poyraxx !

Good luck on landing your dream job! 💯

PR #1876, opened 2026-06-08
fix: load m2m join rows via base meta for mixed STI subtypes.

allangsilva:
## Problem
  
Given two STI subtypes (e.g. `TaskOld` and `TaskNew`) loading the same m2m field (`tags`) in the same tick, it batches both into a single `JoinRows` keyed by the join table (`task_to_tags`). Whichever subtype touches it first wins, and its *subtype* meta is used to load every row in the batch — so it becomes `em.loadAll(TaskOld, <both TaskOld and TaskNew ids>)`. That trips the STI guard in `loadAll` ("... were not of type TaskOld"), since the batch legitimately holds ids of the other subtype.

## Fix
 
 `JoinRows.loadRows` now loads via `getBaseMeta(meta)`, so joist hydrates each row as its correct subtype (from the discriminator) without the guard firing. Non-STI metas pass through `getBaseMeta` unchanged, so existing behavior is preserved.

### DIFF
--- a/packages/core/src/JoinRows.ts
+++ b/packages/core/src/JoinRows.ts
@@ -1,6 +1,6 @@
 import { Entity } from "./Entity";
 import { getEmInternalApi } from "./EntityManager";
-import { EntityMetadata, getBaseAndSelfMetas, getBaseMeta } from "./EntityMetadata";
+import { EntityMetadata, getBaseAndSelfMetas } from "./EntityMetadata";
 import { ReactionsManager } from "./ReactionsManager";
 import { JoinRowTodo } from "./Todo";
 import { keyToTaggedId } from "./keys";
@@ -153,12 +153,8 @@ export class JoinRows {
       twoIds.push(keyToTaggedId(meta2, dbRow[column2])!);
     }
 
-    // Make sure we have entities in memory for all the joined-to tables. For STI, the subtypes
-    // share the base table, so load via the base meta to let a batch hold mixed subtypes without
-    // tripping the STI guard in `em.loadAll`
-    const baseMeta1 = meta1.inheritanceType === "sti" ? getBaseMeta(meta1) : meta1;
-    const baseMeta2 = meta2.inheritanceType === "sti" ? getBaseMeta(meta2) : meta2;
-    await Promise.all([em.loadAll(baseMeta1.cstr, oneIds), em.loadAll(baseMeta2.cstr, twoIds)]);
+    // Make sure we have entities in memory for all the joined-to tables.
+    await Promise.all([em.loadAll(meta1.cstr, oneIds), em.loadAll(meta2.cstr, twoIds)]);
 
     // If we're doing a em.refresh/reload, we need to watch for rows that are no longer here
     const existingRows = new Set<JoinRow>();
--- a/packages/core/src/JoinRows.ts
+++ b/packages/core/src/JoinRows.ts
@@ -153,9 +153,12 @@ export class JoinRows {
       twoIds.push(keyToTaggedId(meta2, dbRow[column2])!);
     }
 
-    // Make sure we have entities in memory for all the joined-to tables. Load via the base
-    // meta so an STI base's join table can hold mixed subtypes without tripping the STI guard in `em.loadAll`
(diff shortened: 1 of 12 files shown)


packages/tests/integration/src/SingleTableInheritance.test.ts
```diff
@@ -87,6 +87,26 @@ describe("SingleTableInheritance", () => {
...
+  it("can load a m2m on a batch of mixed STI subtypes", async () => {
+    // Given two different subtypes that each have a tag, sharing the `task_to_tags` m2m
+    // that is declared on the STI base `Task`
+    await insertTag({ id: 1, name: "t1" });
+    await insertTask({ id: 1, type: "OLD" });
+    await insertTask({ id: 2, type: "NEW" });
+    await insertTaskToTag({ id: 1, task_id: 1, tag_id: 1 });
+    await insertTaskToTag({ id: 2, task_id: 2, tag_id: 1 });
+    const em = newEntityManager();
+    const unloaded = await em.loadAll(Task, ["task:1", "task:2"]);
```
stephenh, line 99: @allangaldinosilva are you sure this test reproduces the issue?

Afaiu after this `em.loadAll`, we'll already know that `task:1` is `TaskOld` and `task:2 is `TaskNew` so the assertions on line 104/105 would alredy pass.

I expected this test to come at the m2m from the other direction, i.e. `em.load` the tag `t1` and populate its `tasks` collection, and then make sure each task was the right type?

packages/core/src/JoinRows.ts
```diff
@@ -153,8 +153,9 @@ export class JoinRows {
       twoIds.push(keyToTaggedId(meta2, dbRow[column2])!);
     }
 
-    // Make sure we have entities in memory for all the joined-to tables
-    await Promise.all([em.loadAll(meta1.cstr, oneIds), em.loadAll(meta2.cstr, twoIds)]);
+    // Make sure we have entities in memory for all the joined-to tables. Load via the base
+    // meta so an STI base's join table can hold mixed subtypes without tripping the STI guard in `em.loadAll`
+    await Promise.all([em.loadAll(getBaseMeta(meta1).cstr, oneIds), em.loadAll(getBaseMeta(meta2).cstr, twoIds)]);
```
stephenh: @allangaldinosilva I'm fine shipping this as-is to fix a bug -- I'm a little curious if this could be conditional b/c:

* If `meta1` is a CTI subtype like `LargePublisher`
* And the schema/FKs say "these really should `LP`s"
* Then it will create more expensive queries to say "load these ids as Publishers" b/c it mean we'll have three tables in the SQL, `publishers`, `large_publishers` and `small_publishers`

So we could maybe check if the inheritanceType==sti?

allangsilva: Good call, just updated that to use `getBaseMeta` for `sti` m2m only.

packages/tests/integration/src/SingleTableInheritance.test.ts
allangsilva: I'm still trying to make the test changes but yes, the current test does reproduces the issue

Screenshot 2026-06-08 at 6 47 03 PM

allangaldinosilva: Yeah... seems like I cannot reproduce this from the other direction since the buggy `loadAll(<subtype>, <mixed ids>)` only happens when accessing from a subtype instance. The `Tag` side loads `Tasks` via the base meta and never gets to it.

Image

packages/core/src/relations/ManyToManyCollection.ts
```diff
@@ -210,11 +211,11 @@ export class ManyToManyCollection<T extends Entity, U extends Entity>
   }
 
   get meta(): EntityMetadata {
-    return getMetadata(this.entity);
+    return getMetadataForField(getMetadata(this.entity), this.fieldName);
```
stephenh, line 214: @allangaldinosilva I moved the fixed from inside of `JoinRows`, where it had read "an invalid collection.meta and then had to fix it up" to directly inside of `ManyToManyCollection.meta`, so that all callers get the benefit of the fix.

PR #1895, opened 2026-06-22
m2mBatchInsert throws "has no id yet" for an entity deleted mid-flush

allangsilva:
If an entity is deleted during a reaction after it's been hooked and after it was linked into a m2m, the pending join row isn't pruned. It survives into `m2mBatchInsert`, which then reads the deleted entity's idTagged, but the entity was never inserted (created-then-deleted), so it has no id:
 
 ```
  Tag has no id yet
    at failNoIdYet (core/src/index.ts:296)
    at m2mBatchInsert (orm/src/drivers/PostgresDriver.ts:262)
 ```

### DIFF
--- a/packages/tests/bun-sql-pg/src/entities/codegen/BookCodegen.ts
+++ b/packages/tests/bun-sql-pg/src/entities/codegen/BookCodegen.ts
@@ -73,7 +73,6 @@ export interface BookGraphQLFilter {
   id?: ValueGraphQLFilter<BookId>;
   title?: ValueGraphQLFilter<string>;
   author?: EntityGraphQLFilter<Author, AuthorId, GraphQLFilterOf<Author>, never>;
-  authorId?: ValueGraphQLFilter<AuthorId>;
 }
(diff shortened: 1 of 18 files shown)


packages/tests/bun-sql-pg/src/entities/codegen/BookCodegen.ts
```diff
@@ -73,7 +73,6 @@ export interface BookGraphQLFilter {
   id?: ValueGraphQLFilter<BookId>;
   title?: ValueGraphQLFilter<string>;
   author?: EntityGraphQLFilter<Author, AuthorId, GraphQLFilterOf<Author>, never>;
-  authorId?: ValueGraphQLFilter<AuthorId>;
```
allangsilva, line 76: I'm unsure why I got so many codegen changes, just did `yarn db` 🤔

allangsilva:
Solved by https://github.com/joist-orm/joist-orm/pull/1896

PR #1897, opened 2026-06-23
fix: make Temporal usable as a type on Node < 26

allangsilva:
Codegen emits `import { Temporal } from "joist-orm"` and uses `Temporal.ZonedDateTime`, `Temporal.PlainDate`, etc. in type positions. But `joist-orm` only exports `Temporal` as a `const` value (the lazy native/polyfill Proxy), so there's no namespace to resolve those types against.

This works on Node 26, where `Temporal` exists as a global namespace, but on older Node it fails to compile:

  > 'Temporal' only refers to a type, but is being used as a namespace here. ts(2702)

### DIFF
--- a/packages/core/src/temporal.ts
+++ b/packages/core/src/temporal.ts
@@ -30,6 +30,24 @@ export const Temporal = new Proxy(
 ) as typeof TemporalType;
 
 /**
+ * A type-only `Temporal` namespace that merges with the `const Temporal` above.
+ *
+ * The `const` is only a value, so using `Temporal.PlainDate` / `Temporal.ZonedDateTime` as a
+ * type needs a namespace to resolve against. Node 26 has that as a global, but older Node
+ * versions don't, so we add it here by pointing the names at `temporal-polyfill`'s types.
+ */
+export declare namespace Temporal {
+  export type Instant = TemporalType.Instant;
+  export type ZonedDateTime = TemporalType.ZonedDateTime;
+  export type PlainDate = TemporalType.PlainDate;
+  export type PlainTime = TemporalType.PlainTime;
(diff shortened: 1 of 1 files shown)


stephenh (approved):

packages/core/src/temporal.ts
```diff
@@ -29,6 +29,24 @@ export const Temporal = new Proxy(
...
 ) as typeof TemporalType;
 
+/**
+ * A type-only `Temporal` namespace that merges with the `const Temporal` above.
+ *
+ * The `const` is only a value, so using `Temporal.PlainDate` / `Temporal.ZonedDateTime` as a
+ * type needs a namespace to resolve against. Node 26 has that as a global, but older Node
+ * versions don't, so we add it here by pointing the names at `temporal-polyfill`'s types.
+ */
+export declare namespace Temporal {
```
stephenh, line 39: Native question @allangsilva but does:

```
export type Temporal = TemporalType;
```

Work? If it does not, np shipping this as-is. Ty!

allangsilva: yeah, I've tried some different ways but none of the worked out :( 

```js
export type { Temporal } from "temporal-polyfill"
// TS2323: Cannot redeclare exported variable 'Temporal'.
// TS2300: Duplicate identifier 'Temporal'.

export type Temporal = typeof TemporalType
// TS2709: Cannot use namespace 'TemporalType' as a type.
// TS2702: 'Temporal' only refers to a type, but is being used as a namespace here.

export type { Temporal } from "temporal-polyfill" 
// TS2713: Cannot access 'Temporal.ZonedDateTime' because 'Temporal' is a type

import { Temporal as TT } from "temporal-polyfill"
export type Temporal = typeof TT
// TS2713: (same as above)
```

PR #1900, opened 2026-06-25
fix: Align m2o & poly behavior on empty arrays.

stephenh:
Filtering a m2o by `{ publisher: [] }` turned into an `in: []` and would not match any rows -- we did not have a test for this, so added one.

Filtering a poly by `{ parent: [] }` would get pruned and so match _all_ rows -- very different behavior, which we've changed to match the "no rows" behavior.


roberthdg (approved):
Nice!

PR #1927, opened 2026-07-10
feat: Add `addCommitRule` for mid-txn validation rules.

stephenh:


zgavin (approved):
kinda worried about how many rules we're gonna have to move to commit rules, but lgtm 😅

PR #2022, opened 2026-09-14
feat: Insert STI subtypes separately so subtype columns keep their defaults

zgavin:
An STI flush used a single `INSERT` whose column list spanned every subtype present, so a subtype that didn't know a column bound `NULL` for it. That defeated the column's database default, and outright failed for `notNull` columns — but only when two subtypes landed in the same `em.flush()`, so it stayed hidden until it didn't.

```
newItemTemplateItem(em, {});
newTakeoffLineItem(em, {});
await em.flush();
// error: null value in column "is_tax_exempt" violates not-null constraint
```

### Changes

- **`addInserts` emits one `INSERT` per subtype**, built from `meta.allFields` (the base's fields plus that subtype's own). Each statement now carries exactly the columns that subtype can write, and anything it can't write is left out entirely so the database applies its default — including volatile ones like `now()`, which materializing a value client-side couldn't have done correctly. This costs one extra statement per additional subtype in a flush, and is unchanged for the common single-subtype case.
- **Codegen rejects pushing a `notNull` column with no default down to a subtype**, since that case has nothing to fall back on. `stiType` plus a nullable column is still the way to model "required on just this subtype", and a `notNull` column *with* a default is now fine.
- Fixed `newStiInsertOp`'s comment, which claimed subtypes carry the root's fields — a subtype's `fields` is `id` plus its own `stiType`-scoped fields only.

### Testing

`tasks` gains a `notNull default 0` column pushed down to `TaskOld`, and a test asserting a `TaskNew` created in the same flush lands on the default rather than failing. Reverting the `addInserts` change fails that test plus five existing ones that create both subtypes. All four integration variants pass, plus the codegen unit tests.


docs/src/content/docs/advanced/single-table-inheritance.md
```diff
@@ -141,9 +142,9 @@ Between Single Table Inheritance (STI) and [Class Table Inheritance](./class-tab
 
 2. With CTI, the schema is safer, because the subtype-only columns can have not-null constraints.
 
-   With STI, if we want `can_bark` to be required for all `Dog`s, we cannot use a `can_bark boolean NOT NULL` in the schema, because the `animals` table will also have `Cat` rows that fundamentally don't have `can_bark` values.
+   With STI, a `can_bark boolean NOT NULL` on the `animals` table would apply to `Cat` rows too, which fundamentally don't have `can_bark` values. It can be done with a default, i.e. `NOT NULL DEFAULT false`, because `Cat`s omit the column from their `INSERT`s and get the default -- but then the constraint is satisfied by the default, not by the `Dog` actually having a value.
```
stephenh: Last sentence should be `not by the Cat`, instead of Dog?

packages/codegen/src/inheritance.ts
```diff
@@ -163,6 +163,25 @@ function expandSingleTableInheritance(
         }
       });
 
+      // A `notNull` column with no database default cannot be pushed down to a subtype: the other
+      // subtypes omit it from their INSERT (see `addInserts`), and the database has nothing to fall back
+      // on. A `notNull` column *with* a default is fine -- the other subtypes simply get the default.
+      const unsettableFieldNames = new Set(
```
stephenh: The name `unsettable` feels wrong, b/c aren't these the fields that we actually must set?

...I guess "unsettable _by the database_" is what the name is trying to say.

Maybe `requiredFieldNames`?

stephenh (approved):

PR #2023, opened 2026-09-14
feat: Allow stiType to name several subtypes

zgavin:
> Stacked on #2022 — review that one first; this PR's diff is against it.

A column meaningful to some subtypes but not all could only be pushed down to one of them, so the alternative was leaving it on the base where every subtype sees it. `stiType` now also takes an array:

```json
"packageDesignNotes": { "stiType": ["TakeoffLineItemVersion", "ProjectCustomItem"] }
```

### Changes

- **`stiType` accepts a string or an array** of subtype names/codes, validated per entry.
- **The strip moved out of the partitioning loop.** It previously removed each subtype's fields from the base as it went, so a field claimed by a second subtype was already gone by the time that iteration looked for it — it would have silently got nothing. The strip now happens once, after the loop, which also makes the loop order-independent.
- **An FK pointing at the base is still only retyped when `stiType` names exactly one subtype.** `m2o.otherEntity` is a single entity feeding the generated `ManyToOneReference<X, Y>` types, so there's no union to put two subtypes in; naming several means the FK can point at any of them, which is what the base type already says.
- The `notNull`-needs-a-default rule from #2022 generalizes rather than special-casing: it fails only when such a column is claimed by a *proper subset* of subtypes. Claiming all of them is equivalent to staying on the base.

### Testing

`tasks` gains a column with `stiType: ["TaskNew", "TaskOld"]`, verified to land on both subtypes and off the base — the case the old code couldn't produce — plus an integration test that both subtypes can write it. Unit tests cover a field on several subtypes, the all-subtypes `notNull` case, and a bad name inside an array. All four integration variants pass.

### Note

Regenerating `config-json-schema.json` also picked up a `"lazy": { "type": "boolean" }` line that was missing — correct output, but unrelated to this change. Happy to drop it if you'd rather keep the diff tight.


docs/src/content/docs/advanced/single-table-inheritance.md
```diff
@@ -90,7 +90,8 @@ These hints in `joist-config.json` generally look like:
 1. Adding an `stiDiscriminator` mapping to the `type` field that Joist will use to know "which subtype is this?"
 2. Adding `stiType: "Dog"` or `stiType: "Cat"` to any column/field (like `canBark` or `canMeow`) in the `animals` table that should be limited to a specific subtype
    - The value of `"Dog"` or `"Cat"` should match a name in the `stiDiscriminator` mapping
-   - Currently, we only support a field being in a single subtype
+   - Use an array, i.e. `stiType: ["Dog", "Cat"]`, for a field that belongs to several subtypes but not all of them
+   - For an FK pointing _to_ your base type (see below), only a lone `stiType` retypes it; naming several leaves it as the base
```
stephenh, line 94: Maybe include an example like i.e. given a `pet_id` FK to the `animals` table that wants to `stiType: [Dog, Cat]`, but there is also a `Horse` subtype, `pet` will still be typed as `Animal` because ...include the reason we can't do this atm...

packages/codegen/src/inheritance.test.ts
```diff
@@ -30,6 +30,45 @@ describe("inheritance", () => {
       ]);
     });
 
+    it("can push a column down to several subtypes", () => {
```
stephenh: These tests look pretty tame, so are likely fine; particularly in the previous PR where we were catching an invalid schema setup, that can't be covered by an integration test.

But otherwise, a few times I've told the LLM "please stop adding mock-data tests in codegen/ and core/ b/c our convention is to test as many things as possible through black box integration tests".

(Because historically these "test from the inside" tests are brittle/make impl assumptions anyway, but maybe that's fine now since LLMs can insta-refactor them anyway?)

I haven't yet codified that in an AGENTS.md rule, but, dunno, wdyt?

packages/tests/integration/joist-config.json
```diff
@@ -122,6 +122,7 @@
     "Task": {
       "fields": {
         "asyncDerived": { "derived": "async" },
+        "sharedSubtypeField": { "stiType": ["TaskNew", "TaskOld"] },
```
stephenh, line 125: Seems like ideally we'd have a `TaskThird` that doesn't get the shared subtype field but 🤷 pretty okay with avoiding that complexity just for this one scenario atm.

stephenh (approved):

PR #2027, opened 2026-09-16
feat: Support recursive collection filters in em.find

stephenh:
## Summary
- Generate recursive collection filters and runtime metadata from shared relation discovery.
- Generate recursive CTEs that start with related entities matching the filter, and use EXISTS to select the entities to return. Handle cycles and exclude entities from their own recursive collections.
- Batch concurrent recursive filters by carrying _find.tag through the matching related entities and each step of the recursive query. Share recursive queries across counts, ID queries, and paginated finds.
- Support nested filters, scopes, inheritance, population, and Knex rendering. Preserve every find tag when a matching related entity joins to multiple rows.
- Document soft-delete behavior, batching, and supported alias comparisons.

## Design choices
- A recursive filter checks related entities at any depth. For example, mentorsRecursive: { firstName: "Alice" } can match Alice through Bob; Bob does not also have to be named Alice.
- By default, soft-deleted related entities do not match, and soft-deleted entities are not returned. Relationships can still be followed through deleted entities: Alice can remain a matching recursive mentor even if Bob, the mentor between her and the author, is deleted.
- Recursive UNION removes duplicate matches and stops cycles from repeating indefinitely; entities never match themselves.
- Comparing a recursive mentor with the author being returned remains unsupported. Scopes can use alias conditions that only refer to the related entity and its relations.

## Validation
- Full stock PostgreSQL integration suite: 2,575 tests and 215 snapshots pass.
- All 30 recursive-filter tests pass in stock, preloading, lazy-row, and lazy-row/preloading variants.
- Core query-parser tests: 28 pass.
- Package and consumer typechecks, builds, codegen, and formatting checks pass.

## Commits
1. Recursive filter API, metadata, CTE generation, and initial integration coverage.
2. Tagged batching and additional coverage for nested recursion, overlapping tags, joins that produce multiple rows, ranges, arrays, pagination, and inheritance.
3. Clarify terminology in the documentation, code comments, and test descriptions.


docs/src/content/docs/advanced/recursive-relations.md
```diff
@@ -47,6 +47,33 @@ Modeling Trees in relational database has historically been a challenge, requiri
 
 :::
 
+### Filtering with `em.find`
+
+Generated recursive collections also appear in entity filter types:
```
stephenh: Just say `Recursive collections ...`, drop the `Generated`

docs/src/content/docs/advanced/recursive-relations.md
```diff
@@ -47,6 +47,33 @@ Modeling Trees in relational database has historically been a challenge, requiri
...
+// Employees with Bob anywhere among their direct or indirect reports.
+await em.find(Employee, { reportsRecursive: { name: "Bob" } });
+
+// Employees without any direct or indirect reports.
+await em.find(Employee, { reportsRecursive: false });
+```
+
+The nested filter checks related entities at any depth. For example, `managersRecursive: { name: "Alice" }` can match Alice through Bob; Bob does not also have to be named Alice. An entity never counts as its own manager or report. Filters can use entities, IDs, arrays, scopes, nested relations, and `and`/`or`. Multiple conditions within one nested filter must match the same related entity.
+
+`true` requires a nonempty recursive collection; `false` or `null` requires an empty one. `undefined` and empty filter objects impose no constraint. `{ ne: employee }` requires some reachable employee other than that employee, following ordinary collection inequality semantics.
```
stephenh: `impose no constraint` -> are ignored

stephenh: Start this paragraph with: "Passing `reportsRecursive: true`

This should probably be a bulleted list.

docs/src/content/docs/advanced/recursive-relations.md
```diff
@@ -47,6 +47,33 @@ Modeling Trees in relational database has historically been a challenge, requiri
...
+
+Joist starts a recursive CTE with the related entities that match the filter, follows relationships back to their collection owners, and filters the outer query with `EXISTS`. This avoids duplicate results and allows normal counting, pagination, and population. The recursive filter itself does not load the collection.
```
stephenh: Remove this paragraph as impl details.

docs/src/content/docs/advanced/recursive-relations.md
```diff
@@ -47,6 +47,33 @@ Modeling Trees in relational database has historically been a challenge, requiri
...
+
+Concurrent finds with the same filter structure are automatically batched, even when their recursive filter values differ. Each find's tag stays with its matching related entities as Joist follows relationships, so nested recursive filters and overlapping paths cannot mix results between finds. Counts, ID queries, and paginated finds also share their recursive traversals within a batch.
```
stephenh: REmoved

docs/src/content/docs/advanced/recursive-relations.md
```diff
@@ -47,6 +47,33 @@ Modeling Trees in relational database has historically been a challenge, requiri
...
+
+Recursive filters use persisted database relationships. By default, soft-deleted related entities do not match, but Joist can follow relationships through them. For example, an employee can still match Alice as a recursive manager when Bob, the manager between them, is soft-deleted. Use `softDeletes: "include"` to include soft-deleted related entities and entities returned by the find. Cycles terminate through deduplication; filtering does not throw `RecursiveCycleError`.
```
stephenh: Remove `Recursive filters use persisted database relationships. `

docs/src/content/docs/advanced/recursive-relations.md
```diff
@@ -47,6 +47,33 @@ Modeling Trees in relational database has historically been a challenge, requiri
...
+
+Aliases cannot be exported from a recursive collection filter with `as`. Use a scope for alias conditions that only refer to the related entity and its relations. Comparisons between that related entity and an alias in the enclosing find query are not supported.
```
stephenh: Remove

packages/core/src/EntityFilter.ts
```diff
@@ -57,6 +57,23 @@ export type EntityFilterObject<T extends Entity, I = IdOf<T>, F = FilterOf<T>, N
 
 export type BooleanFilter<N> = true | false | N;
 
+/** Matches a reachable entity; aliases cannot escape the recursive filter's SQL scope. */
```
stephenh: Say this is a copy/paste of our existing CollectionFilter but with `as` prevented b/c aliases cannot be bound within a recursive filter's scope.

docs/src/content/docs/advanced/recursive-relations.md
stephenh: Dropped Generated in 09eb7078.

docs/src/content/docs/advanced/recursive-relations.md
stephenh: Changed to are ignored in 09eb7078.

stephenh: Converted to bullets starting with Passing `reportsRecursive: true`, using explicit filter examples throughout. Pushed in 09eb7078.

docs/src/content/docs/advanced/recursive-relations.md
stephenh: Removed the CTE implementation paragraph in 09eb7078.

docs/src/content/docs/advanced/recursive-relations.md
stephenh: Removed the batching paragraph in 09eb7078.

docs/src/content/docs/advanced/recursive-relations.md
stephenh: Removed that sentence in 09eb7078.

docs/src/content/docs/advanced/recursive-relations.md
stephenh: Removed the alias paragraph in 09eb7078.

packages/core/src/EntityFilter.ts
stephenh: Updated the JSDoc to explain that this copies the existing EntityFilter collection shape and prevents `as` because the enclosing query cannot use aliases bound inside the recursive filter. Pushed in 09eb7078.

docs/src/content/docs/advanced/recursive-relations.md
```diff
@@ -47,6 +47,30 @@ Modeling Trees in relational database has historically been a challenge, requiri
...
+await em.find(Employee, { managersRecursive: { name: "Alice" } });
+
+// Employees with Bob anywhere among their direct or indirect reports.
+await em.find(Employee, { reportsRecursive: { name: "Bob" } });
+
+// Employees without any direct or indirect reports.
+await em.find(Employee, { reportsRecursive: false });
+```
+
+The nested filter checks related entities at any depth. For example, `managersRecursive: { name: "Alice" }` can match Alice through Bob; Bob does not also have to be named Alice. An entity never counts as its own manager or report. Filters can use entities, IDs, arrays, scopes, nested relations, and `and`/`or`. Multiple conditions within one nested filter must match the same related entity.
```
stephenh: "checks related" -> "will match a related entity at any depth"

packages/core/src/dataloaders/batchRecursiveCtes.ts
```diff
@@ -0,0 +1,76 @@
+import { kqDot } from "../keywords.ts";
+import type { ExistsCondition, ParsedFindQuery } from "../QueryParser.ts";
+import { visitConditions, visitQueries } from "../QueryVisitor.ts";
+import { fail } from "../utils.ts";
+
+/**
+ * Carries the find tag through each generated recursive filter after argument replacement.
```
stephenh: Instead of carries, can we say pushes or threads?

Also given an example of i.e. `mentorsRecursive: { firstName: ? } }` that is `arg4` column of `_find` means we need to rewrite it to `{ firstName: _find.arg4 }`.

packages/core/src/dataloaders/batchRecursiveCtes.ts
```diff
@@ -0,0 +1,76 @@
...
+}
+
+/**
+ * Adds a local _find binding to matching related entities and carries its tag through reachability.
+ * The recursive UNION deduplicates the resulting (match_id, owner_id, tag) rows.
+ */
+function tagRecursiveCtes(query: ParsedFindQuery): boolean {
+  let tagged = false;
+  for (const cte of query.ctes ?? []) {
+    if (!cte.recursiveFilter || cte.query.kind !== "recursive") continue;
```
stephenh, line 50: Why do only recursive CTEs need this rewriting? Couldn't a non-recursive CTE need the same rewrite?

packages/core/src/dataloaders/batchRecursiveCtes.ts
stephenh: ...actually it sounds like we expose `firstName` as a new column? What if we need more than one column?

docs/src/content/docs/advanced/recursive-relations.md
stephenh: Changed to ‘will match a related entity at any depth’ in 8a23ca3f.

packages/core/src/dataloaders/batchRecursiveCtes.ts
stephenh: Changed to ‘threads’ and added the arg4 example in 8a23ca3f.

`_find.arg4` stores the firstName filter value for each find. `collectAndReplaceArgs` rewrites the matching query to `a1.first_name = _find.arg4`. If lastName also varies, that condition reads another argument, e.g. `a1.last_name = _find.arg5`. Shared conditions remain inline.

The matching CTE returns `(id, tag)` and the recursive CTE returns `(match_id, owner_id, tag)`. We add `tag` to those results, not firstName or lastName. The filter can use multiple fields without adding their entity columns to the recursive CTE. The updated JSDoc separates argument replacement from threading the tag.

packages/core/src/dataloaders/batchRecursiveCtes.ts
stephenh: Yes. A non-recursive CTE with varying batched arguments also needs `_find` in its own FROM, and its results need a tag if they combine matches for multiple finds. Argument replacement already visits both recursive and non-recursive AST-backed CTEs.

The guard here identifies the pair generated for a recursive collection filter: the ordinary matching CTE and the recursive CTE. The helper adds `CROSS JOIN _find` and selects `tag` in the ordinary matching CTE, then threads that tag through both recursive terms and compares it in EXISTS. So it already updates a non-recursive CTE; the guard selects the known pair rather than saying only recursive SQL needs this.

This helper does not provide general tag propagation for arbitrary CTEs with different returned columns and consumers. Added that distinction beside the guard in 8a23ca3f.

docs/src/content/docs/advanced/recursive-relations.md
```diff
@@ -47,6 +47,30 @@ Modeling Trees in relational database has historically been a challenge, requiri
...
+await em.find(Employee, { managersRecursive: { name: "Alice" } });
+
+// Employees with Bob anywhere among their direct or indirect reports.
+await em.find(Employee, { reportsRecursive: { name: "Bob" } });
+
+// Employees without any direct or indirect reports.
+await em.find(Employee, { reportsRecursive: false });
+```
+
+The nested filter will match a related entity at any depth. For example, `managersRecursive: { name: "Alice" }` can match Alice through Bob; Bob does not also have to be named Alice. An entity never counts as its own manager or report. Filters can use entities, IDs, arrays, scopes, nested relations, and `and`/`or`. Multiple conditions within one nested filter must match the same related entity.
```
stephenh: Make `Filters can use` a bullet point in the below list. 

Also put `An entity never counts as its own manager or report` in the same list, but as `never its own parent or child`.

packages/core/src/dataloaders/findIdsDataLoader.ts
```diff
@@ -72,14 +72,17 @@ export function findIdsDataLoader<T extends Entity>(
...
+        delete query.ctes;
         const query2: ParsedFindQuery = {
           selects: ["_find.tag as tag", "_data.id as id"],
           tables: [
             { join: "primary", table: "_find", alias: "_find" },
             { join: "lateral", query, table: meta.tableName, alias: "_data", fromAlias: "_f" },
           ],
           // For each unique query, capture its filter values in `bindings` to populate the CTE _find table
-          ctes: [buildUnnestCte("_find", argsColumns, createColumnValuesFromPrepared(args, entries))],
+          ctes: [buildUnnestCte("_find", argsColumns, createColumnValuesFromPrepared(args, entries)), ...ctes],
```
stephenh, line 86: Why do these ctes not need passes through `batchRecursiveCtes` like `findDataLoader`?

packages/core/src/QueryParser.recursive.ts
```diff
@@ -0,0 +1,186 @@
...
+export function findRecursiveRelation(meta: EntityMetadata, name: string): RecursiveRelationMetadata | undefined {
+  if (meta.recursiveRelations?.[name]) return meta.recursiveRelations[name];
+  for (const base of meta.baseTypes) {
+    if (base.recursiveRelations?.[name]) return base.recursiveRelations[name];
+  }
+  return undefined;
+}
+
+/**
+ * Adds an exact reachability CTE, seeded by entities matching the collection filter.
```
stephenh: What is an "exact reachability CTE", too jargon

packages/core/src/QueryParser.ts
```diff
@@ -67,6 +68,8 @@ export interface ExistsCondition {
   subquery: ParsedFindQuery;
   /** Outer aliases referenced by the correlation predicate, for join pruning. */
   outerAliases: string[];
+  /** The generated reachability CTE whose membership must stay within one batched find. */
+  recursiveCte?: string;
```
stephenh, line 76: Why do we need this new field?

docs/src/content/docs/advanced/recursive-relations.md
stephenh: Moved both sentences into the bullet list and changed the second to ‘An entity never counts as its own parent or child.’ Pushed in 93fca7de.

packages/core/src/dataloaders/findIdsDataLoader.ts
stephenh: They do pass through it: `findIdsDataLoader` calls `collectAndReplaceArgs(query, entries)`, which calls `batchRecursiveCtes(query)` before returning. The entity, count, ID, and paginated loaders all use that shared helper. By the time these CTEs move above the lateral query, their tags are already added.

Added that explanation to the shared helper and the three lateral-query loaders in 93fca7de.

packages/core/src/QueryParser.recursive.ts
stephenh: Replaced it with ‘Adds a recursive CTE that finds entities whose recursive collection contains a matching related entity.’ Also removed the remaining reachability wording from the source comments. Pushed in 93fca7de.

packages/core/src/QueryParser.ts
stephenh: It identifies which EXISTS was generated for a recursive collection and which CTE it reads, e.g. `rr`. `correlateRecursiveTag` uses that marker to add `rr.tag = _find.tag` when batching. Without the comparison, one find could accept a match belonging to another find. Ordinary EXISTS conditions should not get that comparison.

It is parser metadata, not another SQL column. We could infer the association by looking up each EXISTS table in the enclosing CTE definitions; this field records it directly when generating the filter. `outerAliases` cannot replace it: that field records the outer author alias (`a`), not the recursive CTE (`rr`). Expanded the JSDoc with the actual tag comparison in 93fca7de.

PR #2032, opened 2026-09-17
feat: Add SQL expression literals

stephenh:
## Summary
- Support expression literals for CASE, COALESCE, NULLIF, GREATEST, and LEAST as named values in select.
- A plain select object always names result columns. Scalar expression literals require select: expr(...); columns and other existing Expr values can be selected directly.
- Use expr(...) for reusable expressions and method calls such as .eq(...).
- Keep CASE self-contained with WHEN entries and an optional final ELSE.
- Support nesting, dynamic operand arrays, bound literals, scalar subqueries, derived tables, CTEs, and named set operands.
- Remove operand-shape guessing and simplify select result types, validation, scope checks, and ordering.
- Infer LEFT-join nullability and preserve compatible codecs, including tagged ids and custom values.
- Preserve join references and prune CASE entries whose optional conditions disappear.
- Validate operand counts, CASE ordering, compatible values, and table scope, with detailed errors.
- Organize operation-specific types, parsing, SQL rendering, and nullability in expressions/case.ts, coalesce.ts, nullIf.ts, greatest.ts, and least.ts. Shared inference and codec rules stay centralized.
- Document all five operations under Expression Literals, with Author/Book examples and inline SQL snapshots.

## Verification
- Package build and package/consumer TypeScript checks passed.
- Query integration suites: 423 tests and 163 snapshots passed in each of stock and lazy-row modes.
- Coverage includes keyword-named projection columns, rejection of bare operand arrays, explicit scalar expressions, correlation, LEFT joins, codecs, and set queries.
- Module-loading, lint, and formatting checks passed.


docs/src/content/docs/features/queries-raw.md
```diff
@@ -148,6 +148,57 @@ the same conditions as `where`, including `sql.condition` and `and`/`or` groups.
...
+
+Use `expr` to choose among values with object descriptions. Values can be columns, other expressions,
+bound literals, or nested descriptions:
+
+```ts
+import { expr } from "joist-orm";
+
+const marketName = expr({
+  coalesce: [
+    { case: { when: selfMarket.id.isNotNull(), then: p.custom_name } },
```
stephenh: Rework this example to use Authors/Books and also be simpler; keeping a spread of some sort would be nice to show how easily "building a dynamic clause" is.

docs/src/content/docs/features/queries-raw.md
```diff
@@ -148,6 +148,57 @@ the same conditions as `where`, including `sql.condition` and `and`/`or` groups.
...
+    standaloneMarket.custom_name,
+    ...markets.map((market, i) => ({
+      case: { when: market.id.isNotNull(), then: parents[i].custom_name },
+    })),
+  ],
+});
+```
+
+`coalesce` returns the first non-null value, in list order. A CASE without `else` returns null when its
```
stephenh: Make these bullet points, and just preface with "they all match their SQL behavior" or what not.

docs/src/content/docs/features/queries-raw.md
```diff
@@ -148,6 +148,57 @@ the same conditions as `where`, including `sql.condition` and `and`/`or` groups.
...
+
+For several CASE arms, use an array. The first true condition wins, even if its value is null:
+
+```ts
+const ageGroup = expr({
+  case: [
+    { when: a.age.gte(18), then: "Adult" },
+    { when: a.age.gte(0), then: "Child" },
+  ],
+  else: "Unknown",
```
stephenh: Shouldn't the `else` be within the `case` array, like `case: [{ when: ... }, { else: ... }]`, to make the `case` keyword self-contained? That seems a lot simpler

packages/tests/integration/src/query/query.expressions.test.ts
```diff
@@ -0,0 +1,664 @@
+import { expectTypeOf } from "expect-type";
+import { expr, query, sql, table, tables } from "joist-orm";
+import { Author, Book, User } from "src/entities";
+import { insertAuthor, insertBook, insertUser, update } from "src/entities/inserts";
+import { PasswordValue } from "src/entities/types";
+import { newEntityManager, queries, resetQueryCount } from "src/testEm";
+
+describe("em.query / expression descriptions", () => {
+  it("treats empty Author last names as missing before choosing a fallback", async () => {
```
stephenh: Let's make 2nd level `describes` for the `coalesce` / `case` / etc behavior we're testing.

packages/core/src/expression.ts
```diff
@@ -0,0 +1,419 @@
...
+
+type CheckListLiterals<A, R> = { readonly [K in keyof A]: CheckLiterals<A[K], R> };
+
+/** Valid inputs need no extra constraint; this keeps spread arrays from losing their fixed final operand. */
+type CheckDescription<D> = [D] extends [CheckExpression<D> & CheckLiterals<D, Exclude<BranchResult<D>, null>>]
+  ? unknown
+  : CheckExpression<D> & CheckLiterals<D, Exclude<BranchResult<D>, null>>;
+
+/** Builds a reusable SQL value expression, binding literal values as parameters. */
+export function expr<const D extends ExprDescription>(description: D & CheckDescription<NoInfer<D>>): DescribedExpr<D> {
```
stephenh: Don't make this change yet, but could we build this without a top-level function like `expr`?

I.e. we have `{ and: ... }` and `{ or: ... }` that work without a top-level `condition(...)`

Could we do something similar here?

packages/core/src/expression.ts
```diff
@@ -0,0 +1,419 @@
...
+  if (value === null || !type) return;
+  const domain = type.domain;
+  if (
+    (domain === String && typeof value !== "string") ||
+    (domain === Number && typeof value !== "number") ||
+    (domain === Boolean && typeof value !== "boolean") ||
+    (domain === BigInt && typeof value !== "bigint") ||
+    (domain === Date && !(value instanceof Date))
+  ) {
+    throw new Error("Expression values must have compatible types");
```
stephenh: Maybe interpolate the domain/value into the error message.

packages/core/src/expression.ts
```diff
@@ -0,0 +1,419 @@
...
+  const leaves = valueLeaves(value);
+  const expressions = leaves.filter((v): v is BaseExpr => v instanceof BaseExpr);
+  const first = expressions[0];
+  if (first) {
+    for (const other of expressions.slice(1)) {
+      if (other === first) continue;
+      const a = first.outputType;
+      const b = other.outputType;
+      if (!a || !b || a.dbType !== b.dbType || a.domain !== b.domain || a.idMeta !== b.idMeta) {
+        throw new Error("Expression operands need matching SQL types and codecs");
```
stephenh: Interpolate what the mismatch is into the error message.

packages/core/src/expression.ts
```diff
@@ -0,0 +1,419 @@
...
+    return value;
+  }
+
+  decode(value: unknown): unknown {
+    return this.outputType.domain === BigInt ? BigInt(value as string) : value;
+  }
+}
+
+/**
+ * Validates descriptions and copies their value lists before rendering SQL.
```
stephenh: descriptions, what are descriptions?

packages/tests/integration/src/query/query.expressions.test.ts
```diff
@@ -0,0 +1,664 @@
...
+    const em = newEntityManager();
+    const [a, b] = tables(Author, Book);
+
+    // When comparing Author names to their Book titles
+    const rows = await em.query({
+      from: a,
+      join: [a.books.as(b)],
+      select: { name: expr({ nullIf: [a.first_name, b.title] }) },
+      orderBy: [{ asc: a.id }],
+    });
```
stephenh, line 338: It feels like canonical Joist testing to have each of these tests assert against the generated SQL as inline snapshots.

packages/core/src/expression.ts
stephenh: Yes. We can normalize these objects wherever an expression is accepted, like the condition parser already does for `{ and: ... }`. For example, `select: { name: { coalesce: [a.last_name, a.first_name] } }` could work directly. That would require extending the accepted types and normalization for select values, ordering, grouping, and predicate operands, while preserving codec checks and per-candidate LEFT-join nullability. We would also need to distinguish expression objects from projections or JSON literal values whose keys happen to be `case` or `coalesce`. `expr(...)` could remain optional for building reusable values with methods such as `.eq(...)`. I have kept the required wrapper for now, as requested.

packages/tests/integration/src/query/query.expressions.test.ts
stephenh: Done in f51a2f30. Every expression test now snapshots its query log inline, with fixture inserts excluded. Rejection tests snapshot an empty log to show that they fail before executing SQL. The full query suite passes in both stock and lazy-row modes: 411 tests and 152 snapshots in each.

packages/core/src/expression.ts
stephenh: Done in f51a2f30. Literal errors now include the expected SQL type/domain and the actual value/JavaScript type, e.g. `expected varchar (domain String), got 42 (number)`. Applied this to both column-backed expressions and incompatible literal-only inputs, with exact error snapshots.

packages/core/src/expression.ts
stephenh: Done in f51a2f30. The error prints both operands and names the mismatched parts: SQL type, conversion domain, ID target, or unknown codec. For example, Author.age vs its sum reports `int4 (domain Number) vs int8 (domain Number); mismatched SQL type`. Added error snapshots for storage-type, custom-codec, entity-ID, and unknown-codec mismatches.

packages/core/src/expression.ts
stephenh: I meant the plain objects passed to `expr`, such as `{ coalesce: [b.id, "b:9"] }`. Replaced that vague wording with “expression objects” and concrete examples in f51a2f30. Renamed the public types to `ExprInput` and `ExprFromInput`, and the parameter to `input`.

docs/src/content/docs/features/queries-raw.md
stephenh: Updated in f51a2f30. The example now chooses between Author.last_name, an optional Book.title, and Author.first_name. It keeps `...bookTitles` to show how a conditional array adds or removes a candidate.

packages/tests/integration/src/query/query.expressions.test.ts
stephenh: Done in f51a2f30. Added second-level groups for `nullIf`, `greatest and least`, `coalesce`, and `case`, plus groups for codec/SQL behavior and validation. The tests remain individual `it` cases.

docs/src/content/docs/features/queries-raw.md
stephenh: Updated in f51a2f30: the section starts with “These expressions match their SQL behavior” and lists the COALESCE, CASE, and null rules as bullet points.

docs/src/content/docs/features/queries-raw.md
stephenh: Agreed—changed in f51a2f30 to `{ case: [{ when, then }, { else: fallback }] }`. ELSE must be the final entry and can appear only once. Updated inference, rendering, docs, and tests; fixed tuples get type errors for invalid ordering, while dynamic arrays get runtime validation. The sibling `else` syntax is rejected.

packages/core/src/expression.ts
stephenh: Implemented the requested select support in 550b3e07. Both `select: { coalesce: [...] }` (scalar rows) and `select: { name: { coalesce: [...] } }` (named rows) now work without a wrapper. `expr(...)` remains available for reuse and methods. This also works in query(...) selects, including scalar subqueries, derived tables, CTEs, and named set operands. Projection keys such as `coalesce` and `case` still work because parsing checks the operand shape. All 424 query tests and 165 snapshots pass in stock and lazy-row modes.

docs/src/content/docs/features/queries-raw.md
```diff
@@ -148,6 +148,111 @@ the same conditions as `where`, including `sql.condition` and `and`/`or` groups.
 and order entries are pruned. No matching values produces `null`; use `.coalesce([])` for an empty array.
 With `distinct: true`, PostgreSQL requires ordering expressions to match the aggregate argument.
 
+### CASE and COALESCE
```
stephenh: Rename this section to `Expression Literals` and cover everything.

docs/src/content/docs/features/queries-raw.md
```diff
@@ -148,6 +148,111 @@ the same conditions as `where`, including `sql.condition` and `and`/`or` groups.
...
+const bookTitles = includeBookTitle ? [b.title] : [];
+const rows = await em.query({
+  from: a,
+  join: [a.books.as(b)],
+  select: { name: { coalesce: [a.last_name, ...bookTitles, a.first_name] } },
+});
+// rows: { name: string }[]
+```
+
+`expr(...)` remains available for building a reusable expression or calling methods such as `.eq(...)`.
```
stephenh: `expr` remains available is a worded based on our impl history, users don't care that we refactored to later allow regular litera.s

packages/core/src/expression.ts
```diff
@@ -0,0 +1,506 @@
...
+  type ExprLike,
+  type SqlFragment,
+  exprBrand,
+  joinFragments,
+} from "./Expr.ts";
+import type { MaybeNull, QueryCondition, QueryJoins } from "./query.ts";
+import type { TypeInfo } from "./TypeInfo.ts";
+
+/** One condition and the value to return when it is true. */
+export interface CaseArm {
```
stephenh: Coudl we call this `CaseWhen`?

packages/core/src/expression.ts
```diff
@@ -0,0 +1,506 @@
...
+}
+
+/** The fallback value, allowed only as the last entry in a CASE array. */
+export interface CaseElse {
+  readonly else: unknown;
+  readonly when?: never;
+  readonly then?: never;
+}
+
+const listNames = ["coalesce", "nullIf", "greatest", "least"] as const;
```
stephenh: `listNames` is weird, `exprNames`? And then rename the rest of the `List...` as well.

packages/core/src/expression.ts
```diff
@@ -0,0 +1,506 @@
...
+    });
+
+declare const inputBrand: unique symbol;
+
+/** Retains the input object so query results can account for LEFT joins in each candidate separately. */
+export type ExprFromInput<I> = Expr<InputResult<I, []>, ExpressionSources<I>> & {
+  readonly [inputBrand]: I;
+};
+
+/** Resolves values with the selecting query's joins; an unknown input stays unknown instead of recursing into it. */
```
stephenh: What do you mean "resolves values"?

packages/core/src/expression.ts
```diff
@@ -0,0 +1,506 @@
...
+          : V extends { readonly coalesce: readonly unknown[] }
+            ? CoalesceValue<ListValues<V>, J>
+            :
+                | Exclude<ExpressionValue<ListValues<V>[number], J>, null>
+                | (null extends CoalesceValue<ListValues<V>, J> ? null : never)
+        : V extends { readonly case: infer A }
+          ?
+              | ExpressionValue<ArmValue<A>, J>
+              | (A extends readonly [...unknown[], { readonly else: unknown }] ? never : null)
+          : V extends string
```
stephenh: THis seems very verbse

packages/core/src/expression.ts
```diff
@@ -0,0 +1,506 @@
...
+              ? number
+              : V extends boolean
+                ? boolean
+                : V extends bigint
+                  ? bigint
+                  : V extends readonly (infer E)[]
+                    ? ExpressionValue<E, J>[]
+                    : V;
+
+/** Literal fallbacks use the expression's domain, i.e. "b:9" stays a BookId when paired with Book.id. */
```
stephenh: I don't understand this comment

packages/core/src/expression.ts
```diff
@@ -0,0 +1,506 @@
...
+/** A fixed non-null candidate guarantees a result; a possibly empty candidate array does not. */
+type CoalesceValue<A extends readonly unknown[], J extends QueryJoins> = A extends readonly [infer H, ...infer T]
+  ? Exclude<ExpressionValue<H, J>, null> | (null extends ExpressionValue<H, J> ? CoalesceValue<T, J> : never)
+  : A extends readonly [...infer Before, infer Last]
+    ?
+        | Exclude<ExpressionValue<A[number], J>, null>
+        | (null extends ExpressionValue<Last, J> ? (null extends CoalesceValue<Before, J> ? null : never) : never)
+    : ExpressionValue<A[number], J> | null;
+
+type ArmValue<A> = A extends readonly unknown[]
```
stephenh: Is this for `case`? Should we prefix with `Case...` like `CaseArmValue`?

packages/core/src/expression.ts
```diff
@@ -0,0 +1,506 @@
...
+  : V extends ExprLike<unknown>
+    ? V[typeof exprBrand]["__source"]
+    : V extends ListInput
+      ? ExpressionSources<ListValues<V>[number]>
+      : V extends { readonly case: infer A }
+        ? ExpressionSources<ArmValue<A>>
+        : never;
+
+/** Literal strings and numbers may differ in value while still sharing a SQL result type. */
+type ValueFamily<V> = V extends string
```
stephenh: I'm surprised we didn't already have something like this in `expr.ts`? or `query.ts`?

packages/core/src/expression.ts
```diff
@@ -0,0 +1,506 @@
...
+          ? readonly ValueFamily<Exclude<E, null>>[]
+          : V;
+type IsUnion<V, All = V> = V extends All ? ([All] extends [V] ? false : true) : never;
+type CompatibleValues<V> =
+  true extends IsUnion<ValueFamily<Exclude<ExpressionValue<V, []>, null>>>
+    ? "Expression values must have compatible types"
+    : unknown;
+
+/** Checks nested expression objects without replacing the caller's inferred tuples or mapped arrays. */
+type CheckExpression<V> =
```
stephenh: All of these `Check...` types are extremely complicated; can you go type-by-type and a) simplify it as much as possible, or b) validate it's even worth having/that it's not more complexity than its worth.

packages/core/src/expression.ts
```diff
@@ -0,0 +1,506 @@
...
+  | BaseExpr
+  | { kind: "literal"; value: unknown }
+  | { kind: ListName; values: Value[] }
+  | { kind: "case"; arms: { when: QueryCondition | undefined; then: Value }[]; otherwise: Value; hasElse: boolean };
+
+/** The conversions shared by all result values, separate from their SQL rendering. */
+type ResultCodec = Pick<BaseExpr, "outputType" | "encode" | "decode">;
+
+/** All operands share one result decoder, regardless of which value PostgreSQL chooses. */
+class ChoiceExpr extends BaseExpr {
```
stephenh: When is this `ChoiceExpr` used? Only for `CASE`? If so include that in the nam.e

packages/core/src/expression.ts
stephenh: Renamed these to exprNames, ExprName, ExprArgsInput, and ExprArgs in dd295282. Added a short comment explaining that these operations take operand arrays, while CASE has separate WHEN/THEN entries. The CheckListLiterals helper was removed during the validation simplification.

packages/core/src/expression.ts
stephenh: Yes, it extracts only CASE THEN/ELSE values. Renamed it CaseArmValue and added that explanation in dd295282.

packages/core/src/expression.ts
stephenh: Audited each type and pushed the simplification in dd295282:

| Previous type | Action / reason |
| --- | --- |
| CheckExpression | Merged shape checking and literal fallback checking into one recursive pass. |
| CheckValues | Renamed CheckOperands; retains argument counts, operand recursion, and compatible-type checks. Dynamic lengths are checked at runtime. |
| CheckArms | Renamed CheckCase; retains the distinction between a single WHEN entry and an array that can end with ELSE. |
| CheckCaseEntry | Retained the small local check for WHEN/THEN vs ELSE keys and their nested values. |
| CheckCaseOrder | Retained: catches ELSE-before-WHEN, repeated ELSE, and non-final ELSE in fixed tuples. Dynamic arrays use runtime validation. |
| CheckLiterals | Removed; its leaf assignability check is now in CheckExpression. |
| CheckArmLiterals | Removed; CASE values use the main traversal. |
| CheckListLiterals | Removed; operands use the main traversal. |
| CheckInput | Simplified to wrap one check instead of intersecting two separate traversals. The valid-input-to-unknown step is still needed to preserve tuple inference for spreads followed by a fixed fallback. |

Also removed ValueFamily/IsUnion and reused query.ts CompatibleValue. SQL storage-type and codec compatibility remain runtime checks, since TypeScript alone cannot establish them.

All existing rejection/type assertions still pass. Added focused assertions against generated Book/Author fields showing that readonly valid array fallbacks retain BookId/Color types, while null Book-id elements and invalid Color values are rejected. The full query suite passes in stock and lazy-row modes: 425 tests, 165 snapshots each.

packages/core/src/expression.ts
stephenh: It is used for all five expression literals, not only CASE. Renamed it ObjectExpr in dd295282 and made the comment list CASE, COALESCE, NULLIF, GREATEST, and LEAST explicitly.

packages/core/src/expression.ts
stephenh: Found CompatibleValue in query.ts, used by set operations. Expression validation now shares that check instead of maintaining ValueFamily plus IsUnion; both of those helpers are gone. This also catches mixed AuthorId/BookId operands at compile time, while retaining the runtime codec check. Pushed in dd295282.

packages/core/src/expression.ts
stephenh: Renamed to CaseWhen in dd295282, including the public export and test imports/annotations.

packages/core/src/expression.ts
stephenh: This preserves the result type supplied by an existing expression when a literal fallback would otherwise widen it. For example, COALESCE(Book.id, "b:9") returns BookId, not string, because the Book id codec also handles that fallback. Nullability is computed separately. Rewrote the comment around that explanation and documented BranchResult as collecting column/existing-expression types while excluding plain literal fallbacks. Pushed in dd295282.

docs/src/content/docs/features/queries-raw.md
stephenh: Agreed. Changed to “Use expr(...) to build a reusable expression or call methods such as .eq(...).” Also removed “still” from the projection-key explanation. Both now describe the API directly, without the implementation history. Pushed in dd295282.

packages/core/src/expression.ts
stephenh: It computes the TypeScript result type, rather than evaluating values. Rewrote the comment in dd295282 with a concrete example: LEFT-joined Book.title has type string | null, while COALESCE(Book.title, "Unknown") has type string.

packages/core/src/expression.ts
stephenh: Simplified this in dd295282 by extracting primitive and array literal conversion into WidenLiteral. ExpressionValue now ends with one fallback, and focuses on SQL expressions and join nullability. The helper comment explains the operation directly: literal types such as "Alice" and 18 become string and number.

docs/src/content/docs/features/queries-raw.md
stephenh: Done in dd295282. The section is now “Expression Literals” and covers all five operations, with CASE, NULLIF, and GREATEST/LEAST examples nested under it. The introduction covers both scalar and named selects, dynamic operands, and reusable expr(...) values.

PR #2041, opened 2026-09-18
fix: Propagate left joins through default reference joins

stephenh:
## Summary
- Make default relationship `.as()` joins inherit LEFT joins from their source, including multi-hop required references, physical foreign keys, and polymorphic references.
- Propagate nullability into inferred result types and subquery output metadata while preserving explicit `.inner()` filtering and reusable join entries.
- Add PostgreSQL integration regressions with type assertions and SQL snapshots, and document the behavior.

## Validation
- Query integration suite: 442 tests and 175 snapshots passed.
- `yarn typecheck:packages`
- `yarn typecheck:consumers`
- Formatting and lint checks passed for changed TypeScript files.


packages/tests/integration/src/query/query.test.ts
```diff
@@ -768,6 +768,144 @@ describe("em.query", () => {
...
+      // And two Books that can be selected independently of their reviews
+      await insertBook({ title: "reviewed", author_id: 1 });
+      await insertBook({ title: "unreviewed", author_id: 1 });
+      // And only the first Book has a BookReview
+      await insertBookReview({ book_id: 1, rating: 5 });
+      // And a named root Book distinct from the Book reached through its review
+      const em = newEntityManager();
+      const root = table(Book, "root");
+      const [br, b, a] = tables(BookReview, Book, Author);
+      const authorJoin = Object.freeze(b.authorId.as(a));
```
stephenh, line 784: Why is there an object.freeze here?

packages/tests/integration/src/query/query.test.ts
```diff
@@ -768,6 +768,144 @@ describe("em.query", () => {
...
+      resetQueryCount();
+
+      // When following an optional review through its required Book and Author
+      const rows = await em.query({
+        from: root,
+        join: [{ left: br, on: br.bookId.eq(root.id) }, undefined, br.book.as(b), authorJoin],
```
stephenh, line 791: why is undefined the 2nd item?

packages/tests/integration/src/query/query.test.ts
stephenh: It guards against resolving `.as()` by mutating the original join entry. This test first uses the entry below a LEFT-joined Book, then reuses that same entry with Book as the root, where it must still resolve to INNER. Freezing makes accidental in-place rewriting fail immediately. Added a setup comment explaining that purpose in efcf47c1; the focused regression and its SQL snapshot pass.

packages/tests/integration/src/query/query.test.ts
stephenh: Join arrays allow `undefined` for omitted conditional joins. Placing one between the initial LEFT join and the required-reference chain checks that skipping an entry does not interrupt propagation, at either runtime or the type level. Added an inline comment explaining the deliberately omitted entry in efcf47c1; the focused regression passes.

packages/core/src/query.ts
```diff
@@ -544,12 +545,29 @@ export type NameOf<A> = A extends { readonly [tableMgmt]: { readonly __name: inf
...
 
+/** Adds default reference targets whose source is already nullable; explicit INNER joins do not inherit. */
+type InheritedLeft<X, Left> = X extends { readonly [referenceJoinSource]: infer Src; readonly inner: infer A }
+  ? [Extract<Src, Left>] extends [never]
+    ? never
+    : NameOf<A>
+  : never;
+
+/**
+ * Computes nullable aliases to a fixed point, including join lists inferred as arrays rather than tuples.
```
stephenh: What do you mean "including join lists inferred as arrays"? This seems like a better `// ...` comment within our implementation than public-facing jsdoc?

packages/core/src/query.ts
```diff
@@ -544,12 +545,29 @@ export type NameOf<A> = A extends { readonly [tableMgmt]: { readonly __name: inf
 /** The names of every alias that was LEFT JOINed; `X` is a naked type parameter so this distributes. */
 type LeftJoined<X> = X extends LeftJoin<infer A> ? NameOf<A> : never;
 
+/** Adds default reference targets whose source is already nullable; explicit INNER joins do not inherit. */
+type InheritedLeft<X, Left> = X extends { readonly [referenceJoinSource]: infer Src; readonly inner: infer A }
+  ? [Extract<Src, Left>] extends [never]
```
stephenh: Can you add a inline comment explaining htis?

packages/core/src/query.ts
```diff
@@ -1344,6 +1362,26 @@ function joinedOutput(output: QueryOutput, joins: QueryJoins | undefined): Query
   };
 }
 
+/**
+ * Resolves default reference joins in declaration order without changing reusable join entries.
```
stephenh: Do we need "default" in this sentence?

packages/core/src/query.ts
```diff
@@ -1344,6 +1362,26 @@ function joinedOutput(output: QueryOutput, joins: QueryJoins | undefined): Query
...
+ * Only .as() carries a source handle; explicit INNER joins retain their filtering semantics.
+ * I.e. LEFT Book followed by Book.author.as(a) adds Author to the exact set of nullable handles.
```
stephenh: Do we need "exact set" vs. just "set"?

packages/core/src/Tables.ts
```diff
@@ -174,13 +174,21 @@ type TableShape<T extends Entity, Name extends string> = {
 /** Keeps explicit base and subtype sources independent, with generic entities untracked. */
 type TableNameOf<T> = T extends { __type: { 0: string } } ? TypeNameOf<T> & string : string;
 
-/** A relationship join factory, without a selectable FK expression. */
-export interface ReferenceJoin<U extends Entity, N extends null | never> {
-  as<A extends TableFor<U>>(other: A): [N] extends [never] ? InnerJoin<A> : LeftJoin<A>;
+/** A join-only relationship factory; default required references inherit their source's LEFT join. */
```
stephenh: "A join-only relationship factory" is awkward, say something like "Provides fluent methods for creating joins through this m2o/o2m relationship."

Also the "default required references..." should be a 2nd sentence, and say "Reference joins like b.author.as(a) created via `as` are inner unless the `b` relation was itself already an optional/left joined alias.". Maybe put it on the `as` method itself.

packages/core/src/query.ts
stephenh: Agreed. This refers to `const joins = [j1, j2]` normally inferring an array rather than a tuple, so the implementation works from the union of entries instead of walking tuple positions. Moved that explanation into an implementation comment and simplified the JSDoc in 613a507c.

packages/core/src/query.ts
stephenh: Added an inline explanation and a concrete Book/Author example in 613a507c: Extract checks whether the reference source is among the nullable aliases; no overlap leaves its target INNER, while an overlap adds the target name.

packages/core/src/query.ts
stephenh: Removed “default” in 613a507c. The following sentence already explains that only `.as()` carries the source handle and explicit INNER joins keep their filtering semantics.

packages/core/src/query.ts
stephenh: Agreed—changed it to “set” in 613a507c.

packages/core/src/Tables.ts
stephenh: Reworded the interface description to “Provides fluent methods for creating joins through an owning reference relationship.” Moved the behavior onto `.as()` itself, using `b.author.as(a)` as the example and explaining both required and nullable references. Used “owning reference” because o2m joins have a separate CollectionJoin interface. Pushed in 613a507c.

packages/core/src/query.ts
```diff
@@ -544,12 +545,32 @@ export type NameOf<A> = A extends { readonly [tableMgmt]: { readonly __name: inf
...
+    : NameOf<A>
+  : never;
+
+/**
+ * Finds all aliases that can be NULL through LEFT joins and subsequent reference joins.
+ * I.e. LEFT Book adds Author via Book.author.as(a), then further required references inherit Author's NULL.
+ * Alias-name collisions are conservative, as they are for explicit LEFT joins; runtime uses exact handles.
+ */
+type NullableSources<X, Left = LeftJoined<X>> =
+  // X is J[number], a union of join entries. A variable like `const joins = [j1, j2]` normally infers
```
stephenh: `a union of join entries`, can you suffix i.e. the `join: ...` array in `em.query`.

packages/core/src/query.ts
```diff
@@ -544,12 +545,32 @@ export type NameOf<A> = A extends { readonly [tableMgmt]: { readonly __name: inf
...
+  // an array, not a tuple, so we cannot walk its positions. Add nullable targets until no new names remain.
```
stephenh: "not a tuple, so we cannot walk its positions", can you say "...which means we structure this mapped typed in some specific way..."

Should we say `LeftJoined<X>` seeds this with the actual `left: ...` joins, and then we use `InheritedLeft` to ...walk?...discover... the effectively left joins?

packages/core/src/query.ts
stephenh: Updated to explicitly say “the union of entries in em.query’s `join: [...]` array” in 440f6c76.

packages/core/src/query.ts
stephenh: Expanded the implementation comment in 440f6c76 to explain the structure: this recursive conditional type works over the entry union rather than tuple positions; LeftJoined<X> seeds Left from actual `left: ...` entries; InheritedLeft discovers reference targets whose sources are already nullable; Exclude checks for new aliases, which are added before repeating for further hops.

packages/core/src/query.ts
```diff
@@ -544,12 +545,35 @@ export type NameOf<A> = A extends { readonly [tableMgmt]: { readonly __name: inf
...
+ * Finds all aliases that can be NULL through LEFT joins and subsequent reference joins.
+ * I.e. LEFT Book adds Author via Book.author.as(a), then further required references inherit Author's NULL.
+ * Alias-name collisions are conservative, as they are for explicit LEFT joins; runtime uses exact handles.
+ */
+type NullableSources<X, Left = LeftJoined<X>> =
+  // X is J[number], the union of entries in em.query's `join: [...]` array. A variable like
+  // `const joins = [j1, j2]` normally infers an array, not a tuple, so this recursive conditional type
+  // works over that union instead of walking tuple positions. LeftJoined<X> seeds Left with the aliases
+  // from `left: ...` entries. InheritedLeft<X, Left> discovers reference targets whose sources are in Left.
+  // Exclude removes names already in Left; if any remain, add them and repeat to discover further hops.
```
stephenh: ...oh, so is the `Left` in `Left = LeftJoined<X>` actually a `Todo`-of-joins-to-still-examine, and not "actually left joins"?

stephenh: No—`Left` is the accumulated union of aliases known to be effectively LEFT joined, not a todo list. It only grows; nothing is removed after examination.

For example, with an explicit LEFT BookReview followed by `br.book.as(b)` and `b.author.as(a)`:

1. `LeftJoined<X>` seeds `Left` with `"BookReview"`.
2. Examining all entries in `X` discovers `"Book"`, so the next `Left` is `"BookReview" | "Book"`.
3. Examining all entries again discovers `"Author"`, so `Left` becomes `"BookReview" | "Book" | "Author"`.
4. Another pass adds no new names, so that accumulated union is returned.

`X` remains the full union of join entries on every pass. `Exclude<InheritedLeft<X, Left>, Left>` is just the newly discovered aliases used to decide whether another pass is needed.

stephenh: Oh that's cute. can we reword "if any remain" to "if they were all already known, we stop recursing & use `Left` as the value, otherwise we add the newly-discovered left joins to our set and recurse

stephenh: Updated in 483af647: “If all were already known, stop recursing and use Left as the value; otherwise, add the newly discovered LEFT-joined aliases to our set and recurse.” Formatting check passed.

PR #2050, opened 2026-09-19
feat: Support implicit AND arrays in query and execute filters

stephenh:
## Summary
- Accept `where: [...]` as shorthand for `where: { and: [...] }` in `em.query` and `em.execute`.
- Support readonly arrays, nested condition groups, and undefined-condition pruning, including `queryMaybe` and relationship join trees.
- Preserve the `allowAll: true` requirement for fully pruned mutation filters.

## Validation
- Query and execute integration suites: 298 tests passed, 82 snapshots passed.
- Integration TypeScript check passed.
- Formatting checks passed.


packages/core/src/queries/sql/query.ts
```diff
@@ -246,8 +246,12 @@ export interface Clauses<S extends QuerySelect = QuerySelect, J extends QueryJoi
   with?: WithInput;
   /** A flat join list, or a domain relationship tree rooted at the entity table in from. */
   join?: J;
-  /** A boolean group, an exists/notExists query, or a bare condition such as `a.age.gte(18)`. */
-  where?: QueryCondition;
+  /**
+   * A boolean group, an exists/notExists query, or a bare condition such as `a.age.gte(18)`.
```
stephenh: Do we need to mention `{ and: ... }` and `{ or: ... }`?

packages/core/src/queries/sql/execute.ts
```diff
@@ -364,7 +364,8 @@ export function decodeStatementResult(
 type MutationTarget<T extends Entity> = TableFor<T> &
   (TypeMapEntry<T, "supportsEmExecute"> extends true ? unknown : never);
 type MutationFilter = {
-  readonly where?: SqlCondition | ExprLike<boolean>;
+  /** Arrays are shorthand for an AND group; undefined conditions are pruned. */
```
stephenh: The `query.ts` comment is better, we should mirror them here.

packages/core/src/queries/sql/query.ts
stephenh: Updated in 3b6cd070: explicitly names `{ and: [...] }` and `{ or: [...] }`, and describes arrays as shorthand for `{ and: [...] }`.

packages/core/src/queries/sql/execute.ts
stephenh: Updated in 3b6cd070 with matching documentation structure and array/pruning wording. The predicate list reflects mutation support: and/or groups, boolean expressions, and bare conditions.

PR #2055, opened 2026-09-21
feat: Skip provably-false em.find queries.

stephenh:
I.e. those that are looking for `author: newAuthor` where b/c `newAuthor` isn't even in the db yet, ofc there won't be any rows.


packages/core/src/queries/valueFilters.ts
```diff
@@ -260,39 +264,20 @@ export function makeLike(search: any | undefined): any {
...
- *
- * For int IDs we use -1, and for uuid IDs, we use the nil UUID value:
- *
- * https://en.wikipedia.org/wiki/Universally_unique_identifier#Nil_UUID
+ * We check `idTaggedMaybe` instead of `isNewEntity` because commit rules run after INSERTs while those
+ * entities still report as new until the flush completes (and yet, because they are in the database, we
+ * must keep them in the query instead of ignoring them). An ID can also be assigned before INSERT, so an
+ * ID-bearing entity conservatively falls through to a harmless database query.
  */
-function nilIdValue(meta: EntityMetadata): any {
```
stephenh, line 280: No more `nilIdValue` hack, we use `in: []` instead.

